PR #2, opened 2025-07-08
Fix/ffi2

jmlmir369:
Since the example appears to work correctly on other operating systems, it is suspected that the issue only occurs on Windows. The cause seems to be as follows:

1. endmode error
  In bicube.f or spline.f, endmode is declared as a character* type. This appears to cause a memory error when the .f function terminates. To resolve this, endmode was aligned with the numbering system used on the Julia side.

2. ix_op, iy_op
  When present(ix_op) or present(iy_op) is called with a null value, it is sometimes still interpreted as present(ix_op) == true and executed. To avoid this, the conditional was changed to ix_op > 0.


matt-pharr:
Hi! I appreciate the work, I'm not able to review this at the moment but I want to comment that we cannot include the .dll files in the repository, those are architecture specific and thus are not able to be used for all users. I want to keep the deps/ folder empty of binaries other than those from code we compile here, perhaps you can install the binary libraries on your machine in a separate folder? Are you running on windows or in WSL?

jmlmir369:
> Hi! I appreciate the work, I'm not able to review this at the moment but I want to comment that we cannot include the .dll files in the repository, those are architecture specific and thus are not able to be used for all users. I want to keep the deps/ folder empty of binaries other than those from code we compile here, perhaps you can install the binary libraries on your machine in a separate folder? Are you running on windows or in WSL?


Thank you for your response.

Yes, I'm currently working in a Windows environment.

Among the DLLs, only `libspline.dll` is specifically built for this code and is actually required.  
The other DLLs were only used to make `gfortran` work properly on Windows and are not needed by other users.

Also, `libspline.dll` was generated by modifying the Makefile - specifically by changing `OUTLIB = libspline.dylib` to `OUTLIB = libspline.dll`.  
So as long as the Makefile is properly set up, the DLL can be built automatically and shouldn't cause any issues.

I'll remove all unnecessary DLLs from the repository and keep the `deps/` folder clean,  
leaving only the `build.jl` script.  
I'll also rename my platform-specific Makefile to `Makefile_win64` and upload it separately.

Please let me know if that works for you, or if you have any suggestions.

PR #5, opened 2025-07-28
Import Vacuum fortran code

matt-pharr:
Adds module for calculating vacuum energy matrix wv. So far in this PR we have:

- [x] Fortran source
- [x] Unit testing format set up
- [x] Automatic Github actions testing
- [ ] Julia wrapper for the fortran source
- [ ] Jupyter notebook example

My opinion is that we should do the wrapper and jupyter notebook in a second PR so that we can roll out unit testing ASAP.


matt-pharr (commented):
After a few changes, this work is ready for merge.

PR #7, opened 2025-07-28
Equil - New - Initial works done by Jaymyoung Lee

satelite2517:
These are the works from Jaemyoung with bunch of different new features.


Manifest.toml
bclyons12, line 1: You should not push a Manifest.toml to a repo associate with the main Project.toml

notebooks/equil_test.ipynb
bclyons12, line 1: Please clear all output from a Jupyter notebook before pushing it to the repo. This goes for all Jupyter notebooks.

src/Equilibrium/io.jl
```diff
@@ -0,0 +1,180 @@
+# src/Equilibrium/io.jl
+
+"""
+The `IO` submodule is responsible for reading and parsing raw plasma equilibrium
+data from files. It translates the raw data into the initial data structures and
+fitted splines required by the coordinate solvers.
+"""
+module IO
```
bclyons12, line 8: Don't make this a module. See main comment in review.

src/Equilibrium/types.jl
```diff
@@ -0,0 +1,140 @@
+# src/Equilibrium/types.jl
+
+"""
+- `EquilInput`:          User-facing input parameters.
+- `DirectRunInput`:      Internal data structure for the direct solver.
+- `InverseRunInput`:     Internal data structure for the inverse solver.
+- `PlasmaEquilibrium`:   The final, user-facing output object.
+"""
+module Types
```
bclyons12, line 9: Don't make this a module. See main comment in review.

src/Splines/Helper.jl
```diff
@@ -0,0 +1,120 @@
...
+- `bctype`: The boundary condition as a `String` or `Int`. Valid options are:
+    - `"natural"` or `1`
+    - `"periodic"` or `2`
+    - `"extrap"` or `3`
+    - `"not-a-knot"` or `4`
+
+## Returns:
+- A validated `Int` code (1-4).
+"""
+function parse_bctype(bctype::Union{String, Int})
```
bclyons12, line 29: Use Julia multiple dispatch:
```
function parse_bctype(bctype::String)
    normalized_bctype = replace(lowercase(bctype), r"[-_\s]" => "")
    if haskey(BCTYPE_MAP, normalized_bctype)
        return BCTYPE_MAP[normalized_bctype]
    else
        error("Invalid string for `bctype`: '$bctype'. Valid options are: $(keys(BCTYPE_MAP))")
    end
end

function parse_bctype(bctype::Int)
    # If it's already an integer, validate it.
    if bctype in values(BCTYPE_MAP)
        return bctype
    else
        error("Invalid integer for `bctype`: $bctype. Valid options are: $(collect(values(BCTYPE_MAP)))")
    end
end
```

src/Equilibrium/types.jl
```diff
@@ -0,0 +1,140 @@
...
+- `rmin`: Minimum R-coordinate of the computational grid [m].
+- `rmax`: Maximum R-coordinate of the computational grid [m].
+- `zmin`: Minimum Z-coordinate of the computational grid [m].
+- `zmax`: Maximum Z-coordinate of the computational grid [m].
+- `psio`: The total flux difference `abs(ψ_axis - ψ_boundary)` [Weber / radian].
+"""
+mutable struct DirectRunInput
+    equil_input::EquilInput
+    sq_in::Any       # 1D profile spline (CubicSplineType)
+    psi_in::Any      # 2D flux spline (BicubicSplineType)
```
bclyons12, line 83: Putting `Any` or any abstract type in a struct will lead to bad performance. Not critical to fix now, but if you know the concrete, use it. If you're not sure, you want to use [parametric types](https://docs.julialang.org/en/v1/manual/types/#Parametric-Types)

src/Equilibrium/types.jl
```diff
@@ -0,0 +1,140 @@
...
+        # Quantity 4: The scaled physics Jacobian
+- `ro`: R-coordinate of the magnetic axis [m].
+- `zo`: Z-coordinate of the magnetic axis [m].
+- `psio`: Total flux difference `|Ψ_axis - Ψ_boundary|` [Weber / radian].
+"""
+mutable struct PlasmaEquilibrium
+    equil_input::EquilInput
+    sq::Any             # Final 1D profile spline
+    rzphi::Any          # Final 2D coordinate mapping spline
+    eq_quantities::Any
```
bclyons12, line 134: See comment about abstract types in structs above

src/Splines/BicubicSpline.jl
```diff
@@ -38,40 +67,57 @@ function _bicube_setup(xs::Vector{Float64}, ys::Vector{Float64}, fs::Array{Float
...
+	- `bctypex`: An integer specifying the boundary condition type for x (Default is 4, not a knot)
+	- `bctypey`: An integer specifying the boundary condition type for y  (Default is 4, not a knot) 
 	## Returns:
 	- A `BicubicSpline` object containing the spline handle, x-coordinates, y-coordinates,
 	function values, number of x-coordinates, number of y-coordinates, number of quantities,
 	and boundary condition types.
 	"""
+	
+	local bctype_code_x::Int = parse_bctype(bctypex)
+	local bctype_code_y::Int = parse_bctype(bctypey)
```
bclyons12, line 112: I'm not sure why you'd need `local` here.

src/Splines/FourierSpline.jl
```diff
@@ -0,0 +1,272 @@
...
+    
+    results = _fspline_eval(fourier, x, y, derivs)
+    
+    
+    if fourier.nqty == 1 && isa(x, Real) && isa(y, Real)
+        if isa(results, Tuple)
+            return map(vec -> vec[1], results)
+        else
+            return results[1]
+        end
```
bclyons12, line 265: Julia functions should not be return unstable types. Here you could have it return a scalar or a vector, depending on the inputs. You should either:
1. Use multiple dispatch so that the behavior is separated into different functions depending on the argument type.
2. If there's not specific operation for vector input and you just want to do the scalar operation on each input, then use Broadcasting. Define it only for scalars and call it with `fspline_eval.(x, y)`, for example.

bclyons12 (changes requested):
There's no need to wrap each `.jl` file in its own module. Just fill out the file as a bunch of functions and structs and do `include()` at the top level Equilibrium.jl. Then you don't need to worry about importing/exporting between the various modules.

src/Splines/FourierSpline.jl
jmlmir369: Since the vector logic is simply composed of repeated scalar calls, I'll comment it out.

Manifest.toml
jmlmir369: If there's no Manifest.toml, how is version management handled? Wouldn't there be cases where the code breaks as libraries get updated?

PR #14, opened 2025-07-28
Julia interface for fortran VACUUM

matt-pharr:


matt-pharr (commented):
This should be ready for merge, subject to final comments. @bclyons12 I requested your review to see if you have any comments on julia-ification, though this is pretty bare-bones.

src/Vacuum/Vacuum.jl
```diff
@@ -0,0 +1,74 @@
+module VacuumMod
+
+export mscvac, set_dcon_params
+
+
+const libdir = joinpath(@__DIR__, "..", "..", "deps")
+const libvac = joinpath(libdir, "libvac")
+
+function set_dcon_params(mthin::Int32, lmin::Int32, lmax::Int32, nnin::Int32,
```
bclyons12, line 9: Users are gonna want to pass native system integers(i.e., `Int`, which is `Int64` on most systems), meaning just calling it with `mthin, lmin, lmax, nnin = 4, 1, 4, 2` instead of forcing the user to wrap in `Int32` like you do in runtests.jl. I would either expand the type to `::Integer` here and then wrap each in `Int32(mthin)`, etc. in the function. Alternatively you could have 
```
function set_dcon_params(mthin::Integer, lmin::Integer, lmax::Integer, nnin::Integer, qa1in::Float64, xin::Vector{Float64}, zin::Vector{Float64}, deltain::Vector{Float64})
    return set_dcon_params(Int32(mthin), Int32(lmin), Int32(lmax), Int32(nnin), qa1in, xin, zin, deltain)
end
```

notebooks/vacuum_fortran_example.ipynb
bclyons12, line 1: Be sure the clear all output on a Jupyter notebook before pushing it to GitHub

bclyons12 (approved):
Looks good, but a few minor suggestions

PR #20, opened 2025-07-29
Equilibrium interface updates

logan-nc:
These changes establish a nested TOML file interface similar to the fortran namelists. @jmlmir369 can check the new structs and top level workflow now in Equilibrium.jl

Also de-modularizes the code to simplify things a bit in julia (@bclyons12 can make sure this is ok)


src/Equilibrium/EquilibriumTypes.jl
```diff
@@ -1,55 +1,243 @@
...
+    outstruct = EquilOutput(;output...)
+    return EquilConfig(control=construct, output=outstruct)
+end
+
+"""
+Outer constructor for EquilConfig that enables a toml file 
+    interface for specifying the configuration settings
+"""
+# if this also have default, then conflicts with @kwdef mutable struct EquilConfig.
+function EquilConfig(path::String)
```
jmlmir369, line 122: if this also have default, then conflicts with @kwdef mutable struct EquilConfig.
@kwargs makes EquilConfig(;....) function, and this also make EquilConfig() because EquilConfig(path = "equi.toml") will oberlap.
for this reason, every similar structure made same error

src/Equilibrium/EquilibriumTypes.jl
```diff
@@ -1,55 +1,243 @@
...
+    etol::Float64 = 1e-7
+    use_classic_splines::Bool = false
+
+    input_only::Bool = false
+    use_galgrid::Bool = true
+
+    """
+    Modified internal constructor that enforces self consistency within the inputs
+    """
+    function EquilControl(eq_type, eq_filename, jac_type, power_bp, power_b, power_r,
```
jmlmir369, line 44: (I fixed all error for now)
As I remember, they were all keyword arguments, and sturct was defined by @kwdef mutable struct EquilControl.
this make two same function, EquilControl().

jmlmir369 (commented):
All errors have been fixed, and it works properly in my directory (please check equil_test.ipynb). If it runs successfully on your side and everything looks good, I believe it's ready to be merged.

One thing that concerns me is that the q profile of beta1.00 bends near the edge, unlike the result from equil_dev. (You can see the result for beta1.00 by expanding equil_control = ..., equil_config = ..., and plasma_eq = JPEC.Equilibrium.setup_equilibrium(equil_config).) I'm sure I didn't modify the internal formula in direct.f, so it might be worth double-checking why this result occurs.

PR #32, opened 2025-08-02
Julia-fy Splines

bclyons12:
So far I've just removed the abstract `CubicSplineType`, as well as the concrete `RealSplineType` and `ComplexSplineType` in favor of the parametric `CubicSpline{T}`. I then julia-fied the constructors removing a lot of code duplication. I still need to do the same for the evaluation and integration, but it should be straightforward.


bclyons12:
@logan-nc @matt-pharr Here's an updated version of CubicSpline.jl done in a much more "julia" manner. It uses parametric types and multiple dispatching on the `ccall` definitions to avoid a lot of duplicate code. The whole file is less than half the size of the old one with the same behavior.

matt-pharr:
This is fantastic! Thank you so much. Is it ready to merge now/should I submit my review?

bclyons12:
`BicubicSpline` is now updated as well. I see some changes were made to FourierSpline on `equil_dev`, so I'll wait for that to get merged to `main` and for everything to get synced before I touch that.

Also note that the build is broken right now. The `build.jl` file got wiped and the makefile for Splines got messed up.

logan-nc:
Very helpful! Thanks @bclyons12!

These still have the issue that they return 1, 3, or 5 things right? Should we use this opportunity to julia-fy that into something type stable? What do you recommend?

logan-nc:
I like the explicitly named functions best. Right now, if you need the 2nd derivative and do `nqty=2` you allocate and calculate for the value, first derivative, and second. But maybe you just wanted the second! So, I would vote we make 
```
spline_eval
spline_derivative
spline_second_derivative
```
then, 
```
f, f1, f2 = spline_eval(..., 2)
```
becomes
```
f = spline_eval(...)
f1 = spline_derivative(...)
f2 = spline_second_derivative(...)
```
where, the `f` and/or`f1` line(s) might not be needed and can be cut if not.

matt-pharr:
Personally, I would favor moving to only length-1 splines, and having a struct:

```julia

mutable struct surface_quantities
    psifac: CubicSplineType
    twopif: CubicSplineType
    .
    .
    .
end
```

I think it would be wise to wait to do this until we have a working version of ode merged into main, however, to avoid too much confusion.

matt-pharr:
@bclyons12 The fortran computational advantage, I believe, is limited to not having to repeat the binary search for the spline point for quantities using the same x-axis -- however, as I intend for most such evaluations to be done in batches (e.g. get q for a whole array of psis at once) this in particular should be minimal.

bclyons12:
I updated FourierSpline as well. I think this is ready to merge

Issue #34, opened 2025-08-05
JPEC Name & License


matt-pharr:
Opening this discussion here -- we need to come up with a name for this code suite as well as a license. Without a license, we implicitly withhold all rights to use of the code. Additionally, JPEC is only a tentative name (and wasn't based on an actual acronym, just GPEC -> JPEC). Thoughts on code names or on which license to use are welcome here.

logan-nc:
*Re License:*  We will use the same MIT open source license as the [existing fortran GPEC code](https://github.com/PrincetonUniversity/GPEC/blob/develop/install/LICENSE) but with the copyright nod to CU instead of PU. @matt-pharr can you add a `LICENSE` file next to the top level `Project.toml` and add `license = 'MIT'` to the `Project.toml` file?

logan-nc:
I'll  reproduce the name discussion from the Hackathon here for reference.

Historical context of code names:
- DCON (killer of PEST), RDCON [Glasser PoP 2016], STRIDE (State Transition Rapid Integration with DCON) [Glasser PoP 2018]
- IPEC (Ideal Perturbed Equilibrium Code) [Park PoP 2007]
- GPEC (Generalized Perturbed Equilibrium Code) [Park & Logan PoP 2017]
- EPEC (Extended Perturbed Equilibrium Code) [Fitzpatrick PoP 2020, 2021, 2021]
New name suggestions
- GPEC-J (x3, similar to MARS-* convention)
- JPEC (x3, julia perturbed equilibrium code)
- PEC (“muscular, short”)
- APECS (Analysis of Perturbed Equilibria Code Suite)
- JuPEC

I would default to just keeping it as the Generalized Perturbed Equilibrium Code (GPEC) suite. Thoughts on this include,
- It maximizes name recognition
- It minimizes confusion by providing continuity across the literature
- It sets an important president for name consistency
  * We plan to add a lot more physics capability. If we change the name every time we add something, it will get too confusing.
  * There is already precedent for not changing the name (we didn't when adding SLAYER or STRIDE)
  *  M3DC1 offers a good example of a modeling suite that has kept its name through many additions and despite many capabilities. I find all the MARS-* short hand more confusing than the M3DC1 papers, which explicitly lay out what capabilities they are using the code (ideal, resistive, linear, nonlinear, etc.).
[...]

PR #37, opened 2025-08-05
MIT License

matt-pharr:
Add MIT license.


matt-pharr:
@logan-nc do you want it to be copyrighted in your name, or the trustees of columbia university? I temporarily added "copyright the gpec team" which is too ambiguous.

logan-nc:
Note, this addresses part of #34

PR #41, opened 2025-08-18
Documentation on Setting up JPEC

mfairborn23:
Added a Markdown file explaining how to set up JPEC on a Windows Machine using WSL Ubuntu. This file also has room to add how to set up on a Mac and can be further expanded.


mfairborn23:
I'm not sure exactly where to add this to our docs page.

docs/src/set_up.md
```diff
@@ -0,0 +1,204 @@
...
+    **OR**
+
+    Alternatively you can clone it from GitHub directly to your virtual machine.
+
+    ``` shell
+    git clone https://github.com/OpenFUSIONToolkit/JPEC.git
+    cd JPEC
+    ```
+
+    Technically, you can use the don't have to copy it from your local installation and you can just `cd mnt/c/Users/<Username>/<Path>/JPEC`, but this in not recommended.
```
matt-pharr, line 103:
```suggestion
5. Clone or download JPEC.
    Clone JPEC from github:
    ```shell
    git clone https://github.com/OpenFUSIONToolkit/JPEC.git
    cd JPEC
    ```

    You may use your local installation and just `cd mnt/c/Users/<Username>/<Path>/JPEC`, but this is not recommended.
```

Why is this not recommended?

docs/src/set_up.md
```diff
@@ -0,0 +1,204 @@
...
+        LIBSUFFIX ?= .so
+        ```
+
+        to the makefile in place of 
+        ``` Makefile
+        LIBSUFFIX ?= .dylib
+        ```
+        the Mac version.
+
+        ⚠ Important: Be careful **not to commit** changes to the makefile as most developers use macOS.
```
matt-pharr, line 132: Do you think it would be better to recommend just setting the environment variables? Additionally, take note of this code in JPEC/deps/build.jl:

```julia
if Sys.isapple()
    ENV["LIBS"] = "-framework Accelerate"
    ENV["LIBSUFFIX"] = ".dylib"
elseif Sys.islinux()
    ENV["LIBS"] = "-lopenblas"
    ENV["LIBSUFFIX"] = ".so"
elseif Sys.iswindows()
    ENV["LIBS"] = "-lopenblas"
    ENV["LIBSUFFIX"] = ".dll"
else
    error("Unsupported OS for Fortran build")
end
```

Is this not working? This is supposed to solve the problems you are trying to instruct how to solve here.

matt-pharr: I should note: When we do `LIBSUFFIX ?= .dylib`, we are telling the makefile, "set $LIBSUFFIX to '.dylib' if there is nothing set there yet." The variable LIBSUFFIX should be set already by the code mentioned above in build.jl.

docs/src/set_up.md
```diff
@@ -0,0 +1,204 @@
...
+        ``` shell
+        ls ../../../deps/libspline.so
+        ```
+
+   6. Export library path so Julia can find it
+        ``` shell
+        export LD_LIBRARY_PATH=~/JPEC/deps:$LD_LIBRARY_PATH
+        ```
+
+        Optional: add to `~/.bashrc` for persistence.
```
matt-pharr, line 155: I would favor just adding this to build.jl.

matt-pharr (changes requested):
Thanks and good work! I will add it to the main page. However, in the meantime, let's discuss a few changes to build.jl.

docs/src/set_up.md
mfairborn23: So it looks like that is sort of working. I think it is the ENV["LIBS"] part that is not working. When I restore the makefile to its original form and try to run the program, I get the error `f77: error: unrecognized command-line option ‘-framework’`. If I only remove the `-framework Accelerate` line from the makefile, it seems to build properly, so the change to the LIBSUFFIX seems to be unnecessary. Thanks for catching that!

docs/src/set_up.md
mfairborn23: Sounds good to me!

docs/src/set_up.md
mfairborn23: I think using the mnt/c/ option is also supposed to be slower and can cause some permissions issues. Apparently Linux heavy operations (even including Git commands) may take longer if you are using a version on your Windows machine and not cloned directly to the VM. There is also a greater likelihood of permissions issues. Personally, I have not run into these challenges yet, but that is what I have read.

matt-pharr:
@mfairborn23 you assigned this to me -- are you asking that I tweak it and merge it?

mfairborn23:
@matt-pharr If it isn't too much trouble, that would be fantastic. Thanks! I think you have a better idea of how to keep the documentation consistent with our current documentation than I do, but I would also be happy to make the changes if that works better for you and then if you could review them that would be great.

logan-nc:
@mfairborn23 @matt-pharr can we get this merged before the Hackathon follow-up meeting next week?

docs/src/set_up.md
```diff
@@ -82,72 +82,36 @@
...
         ```
 
-    5. Verify the library exists using
+    4. Verify the library exists using
         ```shell
         ls ../../../deps/libspline.so
         ```
 
-   6. Export library path so Julia can find it
+    5. Export library path so Julia can find it
```
mfairborn23, line 115: Did we already implement something in build.jl so we can get rid of this step? I think I already added this path to my bashrc so I cannot tell if someone starting from scratch would need to do this or not

mfairborn23:
We should also eventually add a short blurb on how to set up for the first time on MacOS to set_up.md.

logan-nc:
> We should also eventually add a short blurb on how to set up for the first time on MacOS to set_up.md.

I assume @matt-pharr or @jhalpern30 could do that now in less than 5 min

logan-nc:
Is this (other than the possible macOS addition) ready to merge @mfairborn23?

docs/src/set_up.md
```diff
@@ -0,0 +1,168 @@
...
+        julia --version 
+        ```
+
+4. Install Python/Jupyter in WSL 
+   
+   This step is only really required if you want to run the `.ipynb` test notebooks.
+   1. To install Python3 and Jupyter notebooks, use these commands
+        ```shell
+        sudo apt install python3-pip python3-venv -y
+        python3 -m pip install --user jupyter jupyterlab notebook ipykernel
```
matt-pharr, line 72: Is there a reason that we need to install python to install jupyter? I have used julia notebooks in environments without a jupyter-enabled python kernel installed on mac

mfairborn23: I am not entirely sure. This is just the way I set it up (and ChatGPT told me to set it up). I think is is probably unnecessary?

mfairborn23:
> Is this (other than the possible macOS addition) ready to merge @mfairborn23?

I think it is mostly ready to merge. Probably just the macOS addition and any clean up like what Matt is suggesting. I will look into his comment and make that change. Otherwise, I think it should be pretty much good to go.

docs/src/set_up.md
matt-pharr: OK -- I will leave a minimal review for now and we can test it some point with a fresh VM to cut it down to being the least amount of work/new stuff to install possible.

matt-pharr:
> > We should also eventually add a short blurb on how to set up for the first time on MacOS to set_up.md.
> 
> I assume @matt-pharr or @jhalpern30 could do that now in less than 5 min

Will do this monday

docs/src/set_up.md
mfairborn23: According to ChatGPT, even if your kernel is Julia, the Jupyter infrastructure (server) typically runs in Python. So it recommends either installing Python or you need to run 
```
using Pkg
Pkg.add("IJulia")
using IJulia
notebook()
```
in the terminal. I think I will leave teh recommendation to install Python3, but I can mention this. Then I will make a merge request.

mfairborn23:
@logan-nc and @matt-pharr I think this pull request is ready to be merged after a quick review. It passed all the checks and I believe that the documentation is accurate on how to get set up on a Windows machine. We can probably go ahead and merge it unless you have other commentary.

logan-nc:
@matt-pharr the merge is blocked until your review comments are resolved. I'll let you resolve and hit merge when you are happy

docs/src/set_up.md
matt-pharr: OK. We can revise these in the future, but my julia installation does not have python installed, so I am sure it is at least possible on macs.

docs/src/set_up.md
matt-pharr: It does not hurt to export it so let's leave it

matt-pharr (approved):

logan-nc:
@matt-pharr I see you approved but didn't resolve comments and merge...is there a reason it's still open? 

Should we ditch the "all comments must be resolved" requirement? I like seeing the comment chains in the PR and resolving them in github hides them

matt-pharr (approved):

PR #59, opened 2025-10-16
Edge truncation logic

jhalpern30:
This PR does a few things and adds some nice functionality to the code addressing edge truncation logic.
1. It implements `ode_record_edge_dW` (previously `ode_record_edge` in the Fortran), which allows the user to set the `psiedge` parameter in the input file; if `psiedge` < the `psilim` determined by `sing_lim!`, we will store the dW at all integration points in this region and take the maximum. We then calculate the final dW in `free_run at this location`. The logic in this function is much cleaner than Fortran because we can remove all "_edge" variables since we store the entire integration in memory.
2. To get (1) working, I wrote a `free_compute_wv_spline` (previously `free_wvmats`) function, which forms a rough spline of the `wv` matrix in the `psiedge -> psilim` region. I also wrote a `free_compute_total` function (previously `free_test`), which computes the first eigenvalue of the wt matrix rapidly within the integration loop based on the wv spline.
3. I also updated `sing_lim!`, making the logic more clear, replacing the name `sas_flag` (which doesn't describe what the flag does, and helps address this [issue](https://github.com/OpenFUSIONToolkit/JPEC/issues/31)) to `set_psilim_via_dmlim`, and removing the "_edge" variables. I also updated the docstring to describe in much greater detail what this function does
4. In Fortran, once the peak dW in the edge is determined, the entire integration is performed again to this new psilim. In Julia, I have no rewritten this functionality to first identify the peak dW in the edge (`ode_record_edge_dW`, same as Fortran); once this peak is found, all we need to do is identify the integration step # of this peak, performs the same `trim_storage!` operation that we do anyway (but this time also cutting off the data after the peak), and updating psilim, qlim, and u for use in the `free_run` calculation. This is all performed at the end of `ode_run` after the integration is finished and before we dump to `euler.h5`, so the file dump also contains the truncated values. This removes a large chunk of semi-confusing code from `Main.jl`, and I am kindof obsessed with how simple this logic becomes when integrated with in-memory integration.
5. Lastly, I added a warning if both `set_psilim_via_dmlim` and `psiedge < psilim` are true. Based on Nik and I's discussion, we want the initial control flow of the code to look like this:
- If the user sets `set_psilim_via_dmlim`  (previously `sas_flag`) = true, then `qlim` is determined from this and we evaluate dW at this location
- If the user sets psi_edge and it is less than `psilim = psi(qlim = min(qmax, qhigh))` , then we determine dW from the maximum within the region `psiedge -> psilim`
- If the user sets both `set_psilim_via_dmlim`  and `psiedge < psilim` , then we send a warning and set `psiedge = psilim ` (i.e. enforce the dmlim behavior only)
- If the user sets neither, then evaluate dW at the usual `psilim = psi(qlim = min(qmax, qhigh))`


src/DCON/Ode.jl
```diff
@@ -873,37 +897,34 @@ function ode_test(ctrl::DconControl, intr::DconInternal, odet::OdeState)::Bool
...
-            @error "Vacuum code not yet integrated. This run has size_edge = $(intr.size_edge) > 0 and integrator passed psiedge/q_edge"
-            #free_test(plasma1, vacuum1, total1, odet.psifac)
-            intr.dw_edge[intr.i_edge] = total1
-            intr.q_edge[intr.i_edge] = q_psifac
-            intr.psi_edge[intr.i_edge] = odet.psifac
-            intr.i_edge = min(intr.i_edge + 1, intr.size_edge)  # just to be extra safe
+function ode_record_edge_dW!(odet::OdeState, ctrl::DconControl, equil::Equilibrium.PlasmaEquilibrium, ffit::FourFitVars, intr::DconInternal)
+
+    # Why is this not just odet.psifac >= ctrl.psiedge?
+    # TODO: ask Nik about this. Is removing the q_edge condition ok?
```
jhalpern30: @logan-nc Fortran had this line:
`IF(sq%f(4) >= q_edge(i_edge) .AND. psifac >= psiedge)THEN  ! second codition handles initialization plasmas with q0~qa`

Is `psifac >= psiedge` sufficient? Or is there a special case that I'm missing

src/DCON/Ode.jl
```diff
@@ -137,8 +144,25 @@ function ode_run(ctrl::DconControl, equil::Equilibrium.PlasmaEquilibrium, ffit::
...
+        if ctrl.verbose
+            println("Truncating integration at peak edge dW: ψ = $(odet.psi_store[odet.step]), q = $(Spl.spline_eval(equil.sq, odet.psi_store[odet.step], 0)[4])")
+        end
+
+        # Update u, psilim, and qlim for usage in determining wp and wt
+        intr.psilim = odet.psi_store[end]
+        intr.qlim = Spl.spline_eval(equil.sq, intr.psilim, 0)[4]
+        odet.u .= odet.u_store[:, :, :, end]
+
+        # TODO: ask Nik if there would be any issues with no overwriting psilim and qlim in euler.h5
```
jhalpern30: @logan-nc Currently, I'm outputting the psilim and qlim determined before truncation to euler.h5, which seems to match the structure of `euler.bin`, since all of these terms are written in `ode_output_step` before integration. Do you forsee any issues appearing down the line if I retain this behavior? It would also be easy for me to just overwrite the `euler.h5` file, but since ultimately we only have the intermediate `euler.h5` file to compare to the Fortran, it seems like matching the Fortran behavior would be correct?

src/DCON/Free.jl
```diff
@@ -326,4 +326,133 @@ function free_write_msc(psifac::Float64, ctrl::DconControl, equil::Equilibrium.P
...
+        for ipert in 1:intr.mpert
+            wv[ipert, :] .*= singfac[ipert]
+            wv[:, ipert] .*= singfac[ipert]
+        end
+
+        # Store flattened matrix in spline field
+        wv_array[i, :] .= reshape(wv, intr.mpert^2)
+    end
+
+    # TODO: do we need an unset_dcon_params function to free VACUUM memory?
```
matt-pharr: Yes, the memory does not get freed in the fortran code, but once we are using solely julia vacuum, we can possibly improve on this process.

logan-nc:
> I am kind of obsessed with how simple this logic becomes when integrated with in-memory integration

😍😍 😆

src/DCON/Ode.jl
logan-nc: So I believe the fortran code does this: integrate once, find the right q to truncate at, hard set psilim and qlim to there, then integrate again. See lines ~270 of dcon.F:
```
            qhigh = q_edge(peak_index)
            sas_flag = .FALSE.
            psiedge = psihigh
            CALL sing_lim
...
            ! full re-run because outputs were written to disk each step
            ! making it hard to backtrack
            CALL ode_run
``` 

So I believe the (second) euler.bin has the new qlim that accurately matches the end of the integration. 

If you agree, then `euler.h5` should also store `qlim` that matches the end of the eigenmodes.

src/DCON/OdeOutput.jl
```diff
@@ -170,15 +170,15 @@ function ode_output_monitor(ctrl::DconControl, equil::Equilibrium.PlasmaEquilibr
...
+    q, singfac, logpsi1, logsingfac, crit = ode_output_get_crit(odet.psifac, odet.u, intr.mpert, odet.m1, ctrl.nn, equil.sq)
 
     # Check for zero crossing
     if crit * odet.crit_prev < 0 # if crit changes sign, zero crossing has occurred
         fac = crit / (crit - odet.crit_prev)
         psi_med = odet.psifac - fac * (odet.psifac - odet.psi_prev)
         dpsi = psi_med - odet.psi_prev
         u_med = odet.u .- fac .* (odet.u .- odet.u_prev)
-        q_med, singfac_med, logpsi1_med, logpsi2_med, crit_med = ode_output_get_crit(psi_med, u_med, intr.mpert, odet.m1, ctrl.nn, equil.sq)
+        q_med, singfac_med, logpsi1_med, logsingfac_med, crit_med = ode_output_get_crit(psi_med, u_med, intr.mpert, odet.m1, ctrl.nn, equil.sq)
```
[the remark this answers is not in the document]
jhalpern30: In the original Fortran, it was output to `crit.bin`. Even though it is unused for now, I'm going to keep it so we can do an analogous `crit.h5`, which I think I'll do eventually for easy plotting of crit vs. psi

src/DCON/Free.jl
jhalpern30: Just added it - can you double check everything looks ok? I ran a test and it seems like everything is getting called correctly

matt-pharr: Yeah looks good to me!

PR #63, opened 2025-10-17
DCON - PERFORMANCE - Various allocation reductions

bclyons12:
Lots of different small changes that reduce allocations significantly (and give a ~15% speed up). Most of the remaining allocations are due to the allocation of small vectors when evaluating splines. There is likely some type instability as we return different number of vectors depending on the number of derivatives we ask for at runtime. At least half of the runtime is in spline evaluations right now, so I suspect cleaning up the allocations will get rid of much of that too. 

## Proposed Solution
Preallocate storage arrays for the spline evaluation inside the spline `struct` itself, then have evaluation operate in-place. We'd need to be careful that we don't reuse the work arrays before we're done with an evaluation. One could explicitly copy the results if needed, which would lead to the same performance we have now. Most places that's not necessary though. We also would need to be careful if multi-threading is used to avoid race conditions

### Before
```
Energies: plasma = -1.6295106497364298, vacuum = 2.5161657505734376, real = 0.886655100837106, imaginary = 8.131792954459206e-5
  2.556694 seconds (7.11 M allocations: 5.300 GiB, 17.08% gc time)
```
image

### After
```
Energies: plasma = -1.629510650318024, vacuum = 2.516165750825614, real = 0.8866551005075556, imaginary = 8.131792146603146e-5
  2.137391 seconds (4.30 M allocations: 2.441 GiB, 14.90% gc time)
```
image

### DIFF
--- a/src/DCON/Ode.jl
+++ b/src/DCON/Ode.jl
@@ -170,7 +170,7 @@ function ode_run(ctrl::DconControl, equil::Equilibrium.PlasmaEquilibrium, ffit::
             euler_h5["normalizations/mfix"] = odet.ifix
             euler_h5["normalizations/fixstep"] = odet.fixstep[1:odet.ifix]
             euler_h5["normalizations/fixfac"] = odet.fixfac[:, :, 1:odet.ifix]
-            euler_h5["normalizations/sing_flag"] = odet.sing_flags[1:odet.ifix]
+            euler_h5["normalizations/sing_flag"] = odet.sing_flag[1:odet.ifix]
             euler_h5["normalizations/index"] = odet.index[:, 1:odet.ifix]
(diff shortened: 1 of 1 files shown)


src/DCON/Sing.jl
```diff
@@ -646,6 +649,9 @@ function sing_der!(du::Array{ComplexF64, 3}, u::Array{ComplexF64, 3},
         # TODO: banded matrix calculations would go here
     end
 
+    tmp = zeros(ComplexF64, size(du,1), size(du,2))
```
jhalpern30: This will create a new allocation every time this is called in the integration loop, right? Maybe adding tmp to OdeState and making this `odet.tmp .= 0`?

Issue #50, opened 2025-10-13
Either making a Julia form of Banded Matrices work or ditching it in the code entirely


jhalpern30:
In the Fortran, banded firms are used extensively for the F, G, and K matrices. I could not get initial attempts of this to work in the Julia code, and based on conversations with @logan-nc, it appears that the banded matrices in the Fortran are not often used or well benchmarked, so they might not necessarily be a required feature in the Julia code. 

All matrix operations in the initial Julia version in this [PR](https://github.com/OpenFUSIONToolkit/JPEC/pull/21) are performed using dense matrices; however, there are still some vestigial banded components that would have taken a larger effort to remove. Specifically, the Fourier splines accept a banded term, which are currently function calls to the base Fortran. This then propagates to `FourFit.jl` where things like the metric tensor terms accept mband as arguments, even though we are only constructing dense matrices. The Fortran also sometimes only stores the lower half of the Hermitian banded matrices, which makes for some ugly lines of Julia code that can definitely be made cleaner (e.g. see the usage of `fmat` in `sing_mmat`.

I see two approximate paths to choose from:
1. Embrace Banded matrices: using `BandedMatrices.jl` make the Julia banded matrix operations work as in the Fortran code and store F/G/K in similar forms. This might make it easier for future users to convert additional sections of the code without an in-depth understanding of my initial implementation. This could be accompanied by a benchmarking of the base DCON code when using banded matrices. 
2. Get rid of all mentions of Banded matrices: this would involve getting rid of all mentions of `mband` in the code, probably starting from the `FourierSpline` implementation, then going into `FourFit.jl`, then the following matrix operations throughout the rest of the code. Depending on what we have to do with the splines, this could be relatively simple and have straightforward code. The main downside is if the banded approach does provide significant speedup and we won't be able to access it

bclyons12:
Julia should use multiple dispatch to specialize on BandedMatrix types and do the factorizations in an optimized way. If `B` is a BandedMatrix, you can always do `A = Matrix(B)` to get the full matrix with all the off-diagonal zeros, then test the code with the dense linear algebra.

For future reference, I use BandedMatrices.jl here: https://github.com/ProjectTorreyPines/QED.jl

jhalpern30:
> I would not prioritize this in the short term though! Reasons:
> 
> 1. If the full matrix Julia is fast enough for fuse then there is no motivation
[...]

PR #64, opened 2025-10-17
Cleaning up dense matrix implementation

jhalpern30:
The main contributions to this PR are providing clarity to the existing implementation of the dense matrices in Julia DCON, addressing the discussion in issue #50. Note this does not "fix the issue" of either fully converting to banded or dense, but puts the code in an intermediate state where I think the "little bit of both" implementation is sufficiently clear to leave as is for now. There are no functional changes to the code. The main cleanups are:

1. Renaming all variables that only store the lower triangle of a Hermitian matrix to "_lower" (this was mostly fmat, but also g in `sing_mmat`), to make it extremely clear to the user that these variables store the prefactored, lower triangle of the Hermitian matrix F = L L^*. This took me probably two weeks of staring at the code to figure out, so I'd like to make the name more clear in Julia.
2. Removing the sections where I manually fill in the upper half of the F and G matrices in `sing_mmat`, creating some rather dirty looking code. Instead, I wrap these functions with `Hermitian(...)`, which fills in the upper triangle under the hood during the LAPACK function call, and mirrors exactly what Fortran is doing.
3. Adding a warning to `Main.jl` that forces `delta_mband = 0` if the user tries to use banded matrices by making it nonzero
4. Removing `dbats, ebats, fbats` - I think these are kinetic matrices that made it into my first go and I didn't realize they were never used until I was more thoroughly going through it now
5. General cleanups to docstrings and comments for increased clarity

Sanity check:
DIIID ideal example on this branch - 
image

DIIID ideal example on develop - 
image

### DIFF
--- a/src/DCON/DconStructs.jl
+++ b/src/DCON/DconStructs.jl
@@ -216,18 +216,17 @@ end
     dmats::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
     emats::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
     hmats::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
-    dbats::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
-    ebats::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
-    fbats::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
     fmats_lower::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
     kmats::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
     gmats::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
+
+    # Used in Free.jl
+    jmat::Vector{ComplexF64} = Vector{ComplexF64}(undef, 2 * mband + 1)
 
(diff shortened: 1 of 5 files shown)


src/DCON/DconStructs.jl
```diff
@@ -216,18 +216,17 @@ end
...
-    fbats::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
-    fmats::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
+    fmats_lower::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
     kmats::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
     gmats::Union{Missing,Spl.CubicSpline{ComplexF64}} = missing
 
+    # Used in Free.jl
+    jmat::Vector{ComplexF64} = Vector{ComplexF64}(undef, 2 * mband + 1)
+
     # TODO: these might be deprecated? They're used with a ahb_flag in free
```
logan-nc: Yes, deprecate these. This just converts the least stable eigenmode displacements to b fields on the surface and forces into boozer coordinates (ahb = Allen Boozer). Utilizing the old GPEC code as a guide, our new julia code will have way more flexible options to output in any jacobian and lots of components (normal, covariant, contravarient fields or flux or half-area-weighted flux). So no need for this outdates functionality. This can be a separate PR.

PR #68, opened 2025-10-21
Reduce spline allocations

bclyons12:
This allocates work arrays for each CubicSpline and BicubicSpline so that evaluations can be done in-place avoiding allocations. It also separates out functions for evaluation of the value and it's derivatives, so the function calls are type stable. This cuts the number of allocations by 75%. Time performance does not improve much, just a few percent, kind of within normal variation. Here are the BenchmarkTools `@benchmark JPEC.DCON.Main("./") seconds=30` Benchmark results for `JPEC/examples/DIIID_ideal_example/`

**develop** ([518966f](https://github.com/OpenFUSIONToolkit/JPEC/commit/518966fe5337c94b828d54fdce08454402764159)): 4.02 M allocations
```
BenchmarkTools.Trial: 15 samples with 1 evaluation per sample.
 Range (min … max):  1.975 s …   2.102 s  ┊ GC (min … max): 7.20% … 11.99%
 Time  (median):     2.042 s              ┊ GC (median):    8.40%
 Time  (mean ± σ):   2.037 s ± 39.959 ms  ┊ GC (mean ± σ):  9.28% ±  1.75%

  █     █ ██  █ █             █ █ █      █ ██  █   █      █  
  █▁▁▁▁▁█▁██▁▁█▁█▁▁▁▁▁▁▁▁▁▁▁▁▁█▁█▁█▁▁▁▁▁▁█▁██▁▁█▁▁▁█▁▁▁▁▁▁█ ▁
  1.97 s         Histogram: frequency by time         2.1 s <

 Memory estimate: 2.12 GiB, allocs estimate: 4020260.
```

**This PR**: 1.10 M allocations
```
BenchmarkTools.Trial: 16 samples with 1 evaluation per sample.
 Range (min … max):  1.891 s …   2.043 s  ┊ GC (min … max):  5.21% … 10.92%
 Time  (median):     2.004 s              ┊ GC (median):    10.68%
 Time  (mean ± σ):   1.984 s ± 42.135 ms  ┊ GC (mean ± σ):   9.46% ±  2.00%

  ▁                ▁▁▁    ▁  ▁ ▁           ▁ ██ ▁ ▁    ▁  ▁  
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁███▁▁▁▁█▁▁█▁█▁▁▁▁▁▁▁▁▁▁▁█▁██▁█▁█▁▁▁▁█▁▁█ ▁
  1.89 s         Histogram: frequency by time        2.04 s <

 Memory estimate: 2.01 GiB, allocs estimate: 1098889.
```

## To-do

- [x] Fix docstrings
- [ ] Clean up vestigial code (if any)


jhalpern30:
Wow you work fast...

Pinging both @logan-nc and @matt-pharr since this is a pretty substantial decision for moving forward and they have more experience with the Fortran splines.

## Clarification

Quick question on this: so the `spline_eval!` I had implemented to speed up the `sing_der` hot loop is slightly different than the `spline_eval!` you have added here. There won't be any issues because of multiple dispatch, but just wanted to think it through since there's a chance for confusion.

Your `spline_eval!`,
```
spline_eval!(spline::CubicSpline{T}, x::Float64) where {T<:Union{Float64, ComplexF64}}
```
is in place in the sense that it modifies the internal work arrays, but it still returns a function, i.e.
```
q = spline_eval!(sq, psi)[4]
```
(which still results in an allocation, yeah?). So this modifies `sq`, but not in a manner that is very important to the user, since the main thing they care about is still the returned value instead of the modified internal work arrays.

My `spline_eval!` (at the bottom of `CubicSpline.jl`),
```
spline_eval!(f, spline, x; derivs=0, f1=nothing, f2=nothing, f3=nothing)
```
fills in a passed-in, preallocated array (ignore the derivatives, I don't think they're implemented right). This way, the function can be called like I do in `sing_der`
```
Spl.spline_eval!(odet.amat, ffit.amats, psieval; derivs=0)
```
to modify `odet.amat` in-place without creating an allocation there. @bclyons12 is my interpretation of the allocations here correct?

## Proposed Solution
In the Fortran, there is a very similar structure in terms of the work arrays except they are exposed to the user. So a line like:
```
 CALL spline_eval(sq,psifac,1)
```
modifies the fields of `sq` in place, which can then be accessed via
```            
q=sq%f(4)
q1=sq%f1(4)
```
I think we avoided this in the initial implementation of Julia, because the Fortran does this on globals and doesn't always do a `spline_eval!` within the function it uses the data in, so it gets confusing as to where something was set. However, I think this provides good performance improvements, as suggested by this PR. I propose doing something similar to the Fortran and adding
```
@expose_fields CubicSpline xs fs fsi fs1 f f1 f2 f3
```
and the equivalent in BicubicSpline, allowing a Fortran-like allocation efficiency and spline interface.

To maintain readability and avoid the downsides of the Fortran implementation, I would also propose that we 1) make it standard practice to always evaluate the spline right before we use it, so it is clear where the internal worker structs are being used and 2) rename the "s" variables which represent the actual nodes the spline is based on to something more clear, like `xnodes, fnodes`, open to suggestions here. I have done 1) already in the initial implementation, so it shouldn't be too hard to enforce. I am unsure why the splines were initially converted to Julia as out-of-place operations, so there may be other reasons I am not aware of to avoid using the Fortran style.

logan-nc:
My hot takes:
- This PR seems like a fine step for now. I fully support distinguishing the derivative calls for type stability. The internal arrays are consistent with the underlying fortran behavior, so reasonable at this point while we are wrapping.
- Jakes approach of filling a passed-in, preallocated array is preferable to the internal work arrays. The primary reason is that this makes parallelization easy/possible. We had to add this functionality to fortran (`spline_eval_external`) when moving to parallel runs for STRIDE, PENTRC, etc. We know will want parallelization in Julia for those as well as new things like stellarator runs.
- As a consequence, I do *not* support exposing the work arrays. This is a step too far towards fortran behavior that we know we want to move away from. Furthermore, I find the idea that we could enforce a rule to "always evaluate the spline right before we use it" to be pie in the sky. If you create this fortran ability you will create the same fortran opacity. It is inevitable.

logan-nc:
Based on the above, the PR would be better if it changed all the spline evals to the passed-in, preallocated array approach..... but....

I am personally still not convinced we shouldn't be moving to an existing Julia spline package. Haven't others spent long hours optimizing all these things? Why not use that community knowledge? If a package doesn't have periodic boundary conditions, who cares - we just need to tack on 2 nodes below 0 and 2 above 2π. None have an `fspline` equivalent (fits functions to cubic spline in x and Fourier series in y)? These are just a wrapper that does an fft and then puts the coefficients into a bunch of `cspline`s - probably just a few lines of Julia to wrap whatever pre-packaged splines we are using. They are also never really used - just their nodes are, which means that could have just been a `fft` line directly. And so on and so forth.

So I guess I am ok with either converting to converting to the passed-in, preallocated array approach or not - but mostly just think we shouldn't spend too much time on this until someone gives a solid attempt at plugging in an existing spline package. @bclyons12 knows more and didn't have the same instinct. Maybe he can tell me why I am wrong.

bclyons12:
@jhalpern30 Haha I realized I could do this pretty quickly so I gave it a try. There is no issue with allocation here on the return. We just return the existing work array, so it's essentially returning the pointer to that array. It's just a convenience thing to make refactoring the rest of the code easier. It would be the same thing as doing
```
spline_eval!(sq, psi)
q = sq.f[4]
```

@logan-nc I understand your concerns and support moving to a Julia package. We just need to find one that fits what we need and/or could be easily wrapped to provide that. Here are the most well-known ones to consider:
- [Interpolations.jl](https://juliamath.github.io/Interpolations.jl/)
- [Dierckx.jl](https://github.com/JuliaMath/Dierckx.jl)
- [DataInterpolations.jl](https://docs.sciml.ai/DataInterpolations/stable/)

And then Interpolations.jl gives a [long list of other packages](https://juliamath.github.io/Interpolations.jl/stable/other_packages/)

matt-pharr:
I agree with your takes @logan-nc. It is acceptable to sacrifice a small amount of performance for dramatically increased clarity. I wanted to fully move away from the fortran style spline outputs toward something like fortran spline_eval_external, very strongly, as it is extremely important to do so to avoid race conditions easily when parallelizing our heavily parallelizable code. Of the bugs I have dealt with in the fortran parallelization, every one of them at least tangentially involved problems with handling constituent variables like spline.f. These splines are mostly global quantities so I would honestly prefer that we keep them _immutable._

I support moving to an external julia package for these eventually provided we can find a package that matches the functionality of the fourier splines; however, it is imperative that we ensure everything is strongly thread safe because this is intended to be both a living and a heavily parallel code.

logan-nc:
Gemini and ChatGPT both recommend Interpolations.jl.
Gemini says
> Yes, you can use Interpolations.jl in parallelized code, such as with Julia's built-in multithreading and distributed computing capabilities. An Interpolations.jl object is read-only after construction, making it safe to use across multiple threads for interpolation lookups. 

ChatGPT warns agains Direckx.jl 
>Avoid Dierckx.jl for threaded evaluation of derivatives. There’s an open issue showing wildly incorrect results when calling Dierckx.derivative from multiple threads; single-thread is fine, but multithreaded derivatives are not thread-safe in practice. 

while recommending Interpolations.jl as
> Pure Julia; very fast construction & throughput on both B-spline and gridded methods; handles multi-D regular and irregularly spaced axes; well-documented boundary conditions/extrapolation; works with modern Julia versions. The docs include benchmark “shootouts” vs Dierckx and others where Interpolations wins across cases. Maintenance stance is “maintained, semver, fixes prioritized,” which signals low surprise from version to version.

bclyons12:
I see an issue about Dierckx.jl not being thread safe, though there is a outdated work-around: https://github.com/JuliaMath/Dierckx.jl/issues/77

If Interpolations.jl can do what we want (perhaps even with some minimal development, that's probably the direction we want to go.

logan-nc:
I think so. The main complaint seems to be the assumption of regular grids, but there is an option to scale the grid that I think would work for us.... except that I don't think we even need it! Our irregular equilibrium psi grids are all just known functional packings of the points so we can spline in regular variable `x` where `x=psin^2` and then easily just eval like`qspl(psin^2)` or even add explicit `x=psin^2; qspl(x)` type stuff in all the functions that happen at a psin value

Note, GPEC splines the solutions on the dynamically determined force-free solution grid (`u` grid) but only ever does calculations at nodes - so that is just wasted of overhead really.

Theta grids are uniform. 

So I think we are g2g. Do you agree @jhalpern30 and @matt-pharr ?

logan-nc (approved):
I approve but defer to @jhalpern30 to merge after deciding if he wants to use this branch to convert to all `eval_external` style in-place evaluations.

matt-pharr:
I agree @logan-nc. I am, however, surprised that Interpolations.jl does not support irregular grids! I am interested to see how/whether using new splines improves the numerics like the integrator did. Honestly I think the eigenmodes probably even shouldn't be splined since they are so non-smooth -- it would be best, when adding GPEC functionality, to only evaluate at points that were output by the integrator, so that we ensure accuracy, though this would be easy for someone to investigate with a scan over fineness of representation in the future.

jhalpern30:
> I approve but defer to @jhalpern30 to merge after deciding if he wants to use this branch to convert to all `eval_external` style in-place evaluations.

Since it sounds like we're going to be overhauling the splines anyway, I'm going to push this off to avoid doing something that will just get overwritten again and doesn't seem to be giving much time improvement. I can just merge this in since the conflicts are minimal.

#70 would be a good option to work on for anyone at the Hackathon meeting tomorrow who wants to start contributing - I am getting backed up fixing the equilibrium code so I'm not sure when I'd be able to get to it.

PR #69, opened 2025-10-21
Solovev_vac_bugfix

jhalpern30:
This pull request addresses a bug I discovered when running the Solovev ideal example. The changes are relatively minor, and the notebook itself is pretty much a copy of the DIIID ideal example just renamed.

## Bug Fixes
1. `wv_farwall_flag` in DCON_CONTROL now defaults to true instead of false, matching the Fortran behavior
2. the `mtheta` being passed into VACUUM was 1 higher than what it should have been, causing the "retrograde angles present" warning. I have fixed this and tried to make it more clear how we are using `mtheta` in Julia, i.e. arrays go from `1:mtheta + 1`
3. Relatively unrelated to this bug but a bug nonetheless, but I noticed that I had a missing `unset_dcon_params` in `free_run` that I now added

## Documentation
With this fixed, I have now added the Solovev ideal example to the `examples folder`. Note that as of this commit, it will not match the Fortran DCON outputs exactly - this is due to another known bug in the equilibrium splines affecting `dV/dPsi` that will be fixed in a future PR. However, this can be bypassed by hardcoding `normalize = false` in `free_run`, which is the only part of this section that uses these values. When I do this, I achieve great agreement between Fortran and Solovev, indicating that all other bugs are fixed.
Fortran normalize = false energies:
image
Julia normalize = false energies:
image


logan-nc (approved):

src/DCON/Free.jl
```diff
@@ -82,7 +80,7 @@ function free_run(ctrl::DconControl, equil::Equilibrium.PlasmaEquilibrium, ffit:
     # end
 
     kernelsignin = 1.0
-    VacuumMod.mscvac(wv, intr.mpert, mtheta, ctrl.mthvac, complex_flag, kernelsignin,
+    VacuumMod.mscvac(wv, intr.mpert, equil.config.control.mtheta, ctrl.mthvac, complex_flag, kernelsignin,
```
logan-nc, line 83: Just to be extra: Can @jhalpern30 just explicitly confirm that this is indeed the right `mtheta` to pass to the fortran despite the fortran indexing being different? It doesn't need a `+1` here?

jhalpern30: I think this section from the Vacuum code demonstrates this - you pass in the `mthin` parameter (`equil.config.control.mtheta` in our case, the global `mtheta` in Fortran), then the arrays that you pass into `set_dcon_params` go from `1:mthin+1`, just like in Julia. I can't make any kind of connection to the `mscvac` function call because it gets assigned to something called `ntsin0`, which then gets reassigned to `ntsin`, which then gets used to assigned `nd12`... so I am completely lost already

image

PR #71, opened 2025-10-22
Bug fix in write logic for dcon.out and crit.out

jhalpern30:
I had forgotten to add if statements to check if `write_dcon_out` and `write_crit_out` are true before trying to write to the files, so the code would try to write to files that didn't exist if either of these two were false and the code would error out. This (very minor) PR fixes that by wrapping these write lines in if statements so the code can be run without dumping to the .out files. All I did (and that was needed to test this) was running the DIIID ideal example with `write_dcon_out` and `write_crit_out` set to false and confirming the files weren't created and the code ran through without erroring out


matt-pharr (approved):
Looks good

Issue #45, opened 2025-09-05
Equilibrium knots bug


jhalpern30:
While testing the fourfit functions, I've noticed discrepancies in the equilibrium knot values between the Fortran and Julia codes, particularly around the magnetic axis. I have attempted to document the errors I've seen in this [notebook](https://github.com/OpenFUSIONToolkit/JPEC/blob/ode_team/notebooks/spline_bug_solovev_ideal.ipynb), but to avoid me getting majorly sidetracked from developing on the DCON side of things, I'd appreciate if someone with existing familiarity with the equilibrium code could help me track it down/tell me if I am incorrectly interpreting things.

I've assigned the equil team members on this, and will send a message in the main GPEC channel to solicit help.

jhalpern30:
In addition, I have found that the third derivative of the q spline evaluation away from the axis, specifically when used to assemble mmat and vmat in the asymptotics for ua. I believe this might be a related issue so I am including it here

i.e.
```      
CALL spline_eval(sq,psifac,3)
 q(3)=sq%f3(4)
```
does not produce the same f3 value as 
```
q[4] = SplinesMod.spline_eval(equil.sq, intr.sing[ising].psifac, 3)[4][4]
```
at the rational surface psi. They are off by nearly an order of magnitude and sign. Note that the q and q1 values are almost identical, and the q2 value is still relatively close. Most likely compounded errors occurring somewhere, but the third derivative is directly used in some terms so I will need this relatively accurately unless the output is insensitive to it

jhalpern30:
> Also, the dV/dpsi calculation appears off. For the Solovev equilibrium I am testing with, the Julia and Fortran values agreed at the axis, but diverged away from it. The Julia appears to still be a smooth function and the discrepancy monotonically increases with psi, so it is likely a constant somewhere in this calculation. i.e.
> 
> ```
> CALL spline_eval(sq,psi,0)
> sq%f(3)
> ```
> 
> does not give the same results as
> 
> ```
> Spl.spline_eval(sq, psi, 0)[3]
> ```

I believe I have tracked down this issue - within `equilibrium_solver`, the dVdpsi term is set like this
```
 sq_fs_nodes[ipsi, 3] = y_out[end, 5] * (2pi) * psio # dV/d(psi)
```
However, if we look at the original Fortran, this term is set via
```
 sq%fs(ipsi,3)=y_out(istep,1)*twopi*psio
```
So the Julia line should actually be
```
[...]

PR #73, opened 2025-10-23
Equilibrium Bugfixes

jhalpern30:
## Bottom Line
This PR addresses some bugs that I've seen popping up in the equilibrium spline knots and documented in #45
With these fixes, I believe the Solovev ideal example is pretty much fully debugged. We achieve great agreement with the Fortran
Julia (develop)
```
Energies: plasma = -24.07138095505862, vacuum = 13.894868338457542, real = -10.176512616601226, imaginary = 1.956812717198786e-8
```
Julia (this branch)
```
Energies: plasma = -15.557902930443888, vacuum = 8.98058256781093, real = -6.577320362633082, imaginary = 7.043038782294978e-9
```
Fortran
```
 Energies: plasma = -1.556E+01, vacuum =  8.981E+00, real = -6.578E+00, imaginary = -2.657E-07
```
### `sol_run`
Arrays were only going to `mr/ma/mz` instead of `mr/ma/mz +1`, essentially leaving off the last grid point in these splines. For reference, we can clearly see in the Fortran that both the max and min value should be in these arrays
image
I fixed this issue and cleaned up the code and docstring in this section since a lot of lines were unused. Here is a before and after for the fix of printing out some relevant values:
Before
image
After (we can see that psi, r, and z now contain their max value)
image

I also added a semi-comprehensive suite of AI generated unit tests for `sol_run` to try to get this trend going (i.e. find a bug, fix it, add unit tests)

### `direct_run`
As mentioned in #45, the `dV/dPsi` values in the spline were extremely off in the splines. To demonstrate this, I wrote a short script that processes the dumped spline nodes for both Fortran and Julia. For the Solovev ideal case (with fixed `sol_run`), I get
image
(note that column 3 of the sq nodes is dV/dpsi). To fix this bug, I found an indexing issue in this line
`sq_fs_nodes[ipsi, 3] = y_out[end, 5] * (2pi) * psio # dV/d(psi)`, which should actually be
`sq_fs_nodes[ipsi, 3] = y_out[end, 2] * (2pi) * psio # dV/d(psi)` based on the Fortran. With this fix, the differences in the `sq` spline nodes become negligible
image

While the `sq.fs` nodes now appear to be all correct, if we look at the `sq.fs1` nodes, we can see there are still discrepancies - small but nonnegligible, and largest at the endpoints. Note this was documented in #45 by seeing larger errors in the `q1` profile at the axis.
image
This is fixed by noticing that the boundary condition on the sq spline is not set correctly:
`sq = Spl.CubicSpline(sq_x_nodes, sq_fs_nodes; bctype=4)`
which corresponds to "not-a-knot", while this spline is supposed to use the "extrap" boundary condition. If I set this line to
`sq = Spl.CubicSpline(sq_x_nodes, sq_fs_nodes; bctype="extrap")` (this is bctype = 3)
I get good agreement in the splines
image
I noticed that in many places elsewhere in Equilibrium, `not-a-knot` was used where the Fortran used `extrap` (such as in `rzphi`. To avoid issues like this in the future, I also switched all (or at least most) of the spline definitions to their string representations instead of integer for clarity, since we use a dictionary mapping anyway when turning them into splines. 

With these fixes, I also checked the `rzphi` and `eqfun` outputs, and they all appeared sufficiently close for me to believe the code is bug-free. There are still some small differences in the derivatives, but I believe this will always exist since these terms are computed from a derivative of a spline of an integration, so numeric differences between Fortran and Julia can compound. I also ran this test on the DIIID ideal example, and achieved similar results, but not identical enough for me to confidently say that #45 is fully resolved, so I am going to leave the issue open. A future PR can do this process for an equilibrium file input, I want to avoid this PR getting too long.

### DIFF
--- a/src/Equilibrium/Equilibrium.jl
+++ b/src/Equilibrium/Equilibrium.jl
@@ -435,7 +435,7 @@ function equilibrium_gse!(equil::PlasmaEquilibrium)
 
 
 
-    flux = Spl.BicubicSpline(collect(rzphi.xs), collect(rzphi.ys), flux_fs; bctypex="periodic", bctypey="periodic")
+    flux = Spl.BicubicSpline(collect(rzphi.xs), collect(rzphi.ys), flux_fs; bctypex="extrap", bctypey="periodic")
 
     source = zeros(Float64, mpsi + 1, mtheta + 1)
     for ipsi in 0:mpsi
--- a/src/Equilibrium/Equilibrium.jl
+++ b/src/Equilibrium/Equilibrium.jl
@@ -63,7 +63,7 @@ function setup_equilibrium(eq_config::EquilibriumConfig, additional_input=nothin
     elseif eq_type == "sol"
 
         if additional_input === nothing
-            additional_input = SolevevConfig(eq_config.control.eq_filename)
+            additional_input = SolovevConfig(eq_config.control.eq_filename)
         end
 
         eq_input = sol_run(eq_config, additional_input)
--- a/src/Equilibrium/Equilibrium.jl
+++ b/src/Equilibrium/Equilibrium.jl
@@ -435,7 +435,7 @@ function equilibrium_gse!(equil::PlasmaEquilibrium)
 
 
 
-    flux = Spl.BicubicSpline(collect(rzphi.xs), collect(rzphi.ys), flux_fs; bctypex=2, bctypey=2)
+    flux = Spl.BicubicSpline(collect(rzphi.xs), collect(rzphi.ys), flux_fs; bctypex="periodic", bctypey="periodic")
 
     source = zeros(Float64, mpsi + 1, mtheta + 1)
     for ipsi in 0:mpsi
@@ -473,7 +473,7 @@ function equilibrium_gse!(equil::PlasmaEquilibrium)
         fs_matrix[:, 1] = flux.fsx[ipsi+1, :, 1]
         fs_matrix[:, 2] = source[ipsi+1, :]
 
-        spline = Spl.CubicSpline(Vector(flux.ys), fs_matrix; bctype=2)
+        spline = Spl.CubicSpline(Vector(flux.ys), fs_matrix; bctype="periodic")
         Spl.spline_integrate!(spline)
 
         term[ipsi+1, :] .= spline.fsi[end, :]
--- a/src/Equilibrium/ReadEquilibrium.jl
+++ b/src/Equilibrium/ReadEquilibrium.jl
@@ -104,8 +104,7 @@ function read_efit(config::EquilibriumConfig)
         qprof_data,
         sqrt.(psi_norm_grid)
     )
-    # According to Spline_document.txt, bctype=4 is Not-a-Knot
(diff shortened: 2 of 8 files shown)


src/Equilibrium/Equilibrium.jl
```diff
@@ -435,7 +435,7 @@ function equilibrium_gse!(equil::PlasmaEquilibrium)
 
 
 
-    flux = Spl.BicubicSpline(collect(rzphi.xs), collect(rzphi.ys), flux_fs; bctypex=2, bctypey=2)
+    flux = Spl.BicubicSpline(collect(rzphi.xs), collect(rzphi.ys), flux_fs; bctypex="periodic", bctypey="periodic")
```
logan-nc: Shouldn't this be extrap and periodic? Like the fortran `      CALL bicube_fit(flux,"extrap","periodic")`

src/Equilibrium/ReadEquilibrium.jl
```diff
@@ -266,8 +265,8 @@ function read_chease2(config::EquilibriumConfig)
     fs[:, :, 2] .= transpose(zzcp[1:ntnova+1, :])
 
 
-    # Setup bicubic spline with periodic boundary conditions (bctype=2)
-    rz_in = Spl.BicubicSpline(xs, ys, fs; bctypex=2, bctypey=2)
+    # Setup bicubic spline with periodic boundary conditions
+    rz_in = Spl.BicubicSpline(xs, ys, fs; bctypex="periodic", bctypey="periodic")
```
logan-nc: Again, I expect this should be extrap, periodic. The fortran is always `CALL bicube_fit(flux,"extrap","periodic")
`

Do a quick repo search for periodic `bctypex="periodic"` and bctypex=2`. I doubt either should ever be used, as I don't know of any (theta, phi) grids in DCON. 2D things are almost always (psin, theta).

src/Equilibrium/ReadEquilibrium.jl
```diff
@@ -402,7 +401,7 @@ function read_chease(config::EquilibriumConfig)
         ys = range(0, 2π; length=mtau) |> collect
 
         # Setup bicubic spline with periodic boundary conditions
-        rz_in = Spl.bicube_setup(xs, ys, fs; bctypex=2, bctypey=2)
+        rz_in = Spl.bicube_setup(xs, ys, fs; bctypex="periodic", bctypey="periodic")
```
logan-nc:
```suggestion
        rz_in = Spl.bicube_setup(xs, ys, fs; bctypex="extrap", bctypey="periodic")
```

logan-nc (changes requested):
Great investigative work! Thanks a ton for digging into this.
Thanks also for setting a good example re unit tests. 

I found a suspicious use of double periodic splines in multiple places that I think needs fixing. Other than that, it looks good

src/Equilibrium/ReadEquilibrium.jl
jhalpern30: Agreed - for the inverse runs I just blindly copied the existing code since I wasn't debugging that, probably should have double checked them all. Just pushed a fix that changes all three of these that I found in the repo search

logan-nc (approved):

PR #74, opened 2025-10-23
Unit Tests Added For Sing.jl

mfairborn23:
This needs clean up and some more development before it is merged with develop, but here is an early version of the sing_der unit test. We still need to add unit tests for the other sing.jl functions and clean up the sing_der one because there is currently a lot of clutter with it


notebooks/sing_find_example.ipynb
jhalpern30, line 1: Should this be here?

jhalpern30:
I think a lot of this can be combining the.dat files into a single HDF5 - this will mean only one data file needs to be added and we can also remove some of the read in functions within the actual testing script, since it can just be like
`amat = h5file["data/amat"]` or something

notebooks/sing_find_example.ipynb
mfairborn23: I don't think so- I think this one got uploaded by accident. Oops

matt-pharr:
We also have to make sure that the compile test passes! It looks like the workflow halts when we add a package to the repo, I will need to figure that out. I will do so over the weekend.

test/runtests.jl
```diff
@@ -1,5 +1,4 @@
-
-
+using LinearAlgebra, BandedMatrices, Random, Test, DelimitedFiles, Printf #TODO: clean up imports - do we need all of these here?
```
jhalpern30: I believe this is causing the issue that @matt-pharr raised with the build failing - the JPEC package isn't using BandedMatrices (it's not in the Project.toml), so I would expect the build to rightfully fail here. So that part of this call should be removed. Also, test is duplicated, so probably that too. And for the rest of them, if they're not in the Project.toml, we can't import them here I don't think, so they either need to be added or alternatives found. I think for the ones that are in there (LinearAlgebra, Printf), doing a `using JPEC` gives them to you for free without needed a using statement here. I think delimited files won't be needed if we switch to HDF5.

mfairborn23: That sounds good! I will look at cleaning up the imports and making sure the Project.toml matches the develop branch later this week

logan-nc:
@mfairborn23 what is the status here?

mfairborn23:
@logan-nc I have to double check this branch again but at least with the previous version of the code I believe that there were unit tests for most functions. My new branch has some additional unit tests for the kinetic versions of the code as well.

logan-nc:
please ping me and @jhalpern30 when this is ready for review. if not soon, then we should just close the pr. it's going to become too stale and out of date

mfairborn23:
@jhalpern30 and @logan-nc I think we need to figure out what this "Documentation" issue is, then this can be reviewed. I attempted to merge develop into it and have confirmed the unit tests are still working. There will be some more unit tests we can add when we merge the kinetic branch in because I wrote a few more there, but I believe this is a good start. Once I am finished with the kinetic part, I will review the code and ensure there aren't any other functions we should make unit tests for.

logan-nc:
@matt-pharr any chance you can make progress on the documentation issue?

logan-nc:
@jhalpern30 it is ok to merge this with .the documentation issue still at large since it is an issue on develop not unique to this branch. 
Hopefully, @matt-pharr  can resolve it quickly in a separate PR.

notebooks/sing_find_example.ipynb
jhalpern30: Reminder to delete this file

test/runtests.jl
```diff
@@ -12,7 +14,7 @@ if !isempty(ARGS)
         include(testfile)
     end
 else
-    include("./runtests_build.jl")
+    #include("./runtests_build.jl") #TODO: I think this can be deleted because it looks like it's all Fortran stuff that has been deleted
```
jhalpern30: This is getting removed in #159 - either remove this entire line or just the comment so you don't get hit with merge conflicts

test/runtests_sing.jl
```diff
@@ -1,22 +1,238 @@
...
-        # odet.u .= ones(ComplexF64, size(odet.u))
-        # Initialize structs with relevant data, can do something like
-        # ffit = FourfitVars()
-        # amat = load in from test_data folder
-        # ffit.amats = SplinesMod.CubicSpline("some psi", reshape(amats, mpsi+1, :); bctype="extrap")
-        # if using one psi values causes issues, can make a psi array and just fill it all with amat
-        # repeat for other matrices, equil data, other relevant constants, etc.
+using Test
+using TOML
+using FastInterpolations: cubic_interp, CubicFit, LinearBinary
```
jhalpern30: I don't think we need using statements at the beginning of every file, try to consolidate them in runtests.jl. i.e. just copy paste them there, since these files are all just added via include statements anyway

test/runtests_sing.jl
```diff
@@ -1,22 +1,238 @@
...
+
+#TODO: these take forever to run- can we optimize them up?
```
jhalpern30: Is this still true? I'd avoid adding tests that are going to jack up the time of our unit tests runs on github. 

If no, just remove the comment and ignore. 

If yes, I think we can remove the comprehensive sing_der test with the external data if that's taking up the time. We already validated against the Fortran, and every PR includes a comparison to develop, so there's no need to check against the Fortran data every time. @logan-nc thoughts on this?

test/runtests_sing.jl
```diff
@@ -1,22 +1,238 @@
...
+            parts = split(line, '=')
+            if length(parts) >= 2 && strip(parts[1]) == item
+                val = strip(parts[2])
+                return parse(Float64, replace(val, 'D' => 'E'))
+            end
+        end
+        return NaN
+    end
+
+    function read_solutions_3d(fname::String)
```
jhalpern30, line 42: Don't worry about it for this PR, but if we leave all the external data in, please make an issue of moving .dat data to HDF5s since it can just be made into one file

test/runtests_sing.jl
```diff
@@ -1,22 +1,238 @@
...
+        for i in 1:length(num_range)
+            mmat[i, :] .= vec(prev_mat)
+        end
+        return mmat
+    end
+
+    # --------------------------
+    # Write output in Fortran-compatible format
+    # --------------------------
+    function write_sing_output(filename, psifac, q, mlow, nn, ud)
```
jhalpern30: Is this needed? It looks like it just prints stuff out, which we don't need in unit tests

test/runtests_sing.jl
```diff
@@ -1,22 +1,238 @@
...
+
+        du_fortran = read_solutions_3d(joinpath(@__DIR__, "test_data/sing_der_testing/mat_dat/sing_der_output_du.dat"))
+        write_sing_output(joinpath(@__DIR__, "test_data/sing_der_testing/mat_dat/sing_der_output_julia.dat"), odet.psifac, odet.q, intr.mlow, ctrl.nn_low, du) #TODO: should it be ctrl.nn_high or ctrl.nn_low?
+
+        @show size(du)
+        @show size(du_fortran)
+        @show maximum(abs.(du .- du_fortran))
+        @test isapprox(du[:, 1, 1], du_fortran[:, 1, 1]; rtol=1e-3) || true
+
+        for sol_idx in 1:1#odet.msol
```
jhalpern30: should this be hardcoded? Why not odet.mpert

jhalpern30 (commented):
Here are some requests before this gets merged - don't spend too much time on this, or lmk if you want help getting this merged. 

This is a rather large diff, I didn't look through the top section since it looks like it was just changes caught by the Julia formatter. I think the biggest things are to remove the extra notebook you added and make sure not to add a unit test that will greatly increase the runtime of the test suite. 

@logan-nc can provide input on this, but the sing_der unit test is rather complicated in creating its own functions + using ~10 external files and might be tricky to maintain. It might be worth replacing it with a minimal test that makes sure sing_der runs through and produces realistic values since we have been checking the outputs against develop in each PR. If sing_der is broken, those outputs would break in pretty much every case

matt-pharr:
@jhalpern30 @mfairborn23 I went and fixed the tests and merged in develop, so this should be good to go now I think pending y'alls review!

matt-pharr:
@jhalpern30 I am just going to merge this now, it has sat for long enough. I did some signficant cleanup on it and it is working now. I talked to Meg and she thought it was ready once I cleaned it up

PR #75, opened 2025-10-24
2D Multiple Toroidal Mode Capabiltiies in DCON

jhalpern30:
NOTE: This PR should probably be merged after #92 and #91 are merged, it will reduce the number of total merge changes from this PR into develop.

This PR adds the capability to run DCON on a 2D equilibrium with multiple toroidal modes at once. It is a stepping stone to a full 3D DCON, since the matrix size is the same for multi-n 2D and 3D, being MN x MN, where M is the total number of poloidal modes and N the total number of toroidal modes. The main difference is that in 2D, these matrices are block diagonal (with N blocks of size M x M) since the toroidal modes are decoupled. In this PR, I have made most sections general for either 2D or 3D, with the exception of the singular asymptotics, which will need full generalization based on an understanding of Glasser PoP 2023.

This is a rather large PR. Here is a list of the main changes:

## `nn -> nlow/nhigh`
In DCON, `nn` is set as an input variable and stored in the `DconControl` data structure. For multi-n, I have added additional input parameters `nn_low` and `nn_high`. There is a section in `Main.jl` that checks to make sure that the user is trying to use only one of `nn` and `nn_low/nn_high`, and then sets the corresponding parameters in the `intr` data structure. I also define `numpert_total = mpert * npert`, which replaces `mpert` throughout most of the code

## Finding singular surfaces
The `sing_find` functions is built to find singular surfaces at a single n, but is fairly straightforward to adapt to multiple n modes by just adding an outer loop. The main differences are now storing `sing.m` as an array of the resonant modes to account for singular surfaces with multiplicity > 1 and sorting the singular surfaces by `psifac` after finding them (since we search in order of increasing `n`, but we'll encounter them during integration in the order of increasing `psifac`

## Constructing Fourier Matrices
There is now an additional loop over the toroidal mode numbers to assemble these matrices. I have kept this code general to fill in matrices of size MN x MN, where we use more complex indexing to figure out where we are in the larger matrix. This keeps it relatively general for 3D DCON. Note that in the 2D case these matrices are block diagonal, and so there can be further optimizations to treat these are block diagonal, as right now the code treats all matrices as dense

## Singular Surface Asymptotics
This is probably the section that is most unlike what its final state will be in 3D, and lies somewhere in between 2D and fully 3D. Essentially, I added a treatment for resonant surfaces with multiplicity > 1, which mostly involves storing the multiple indices of the resonant modes and performing the same analysis as in the 2D case multiple times for each resonance. It appears that the parameter alpha is the same for all n in the 2D equilibrium (but I don't have a proof for this). When subbing in the asymptotics in `ode_ideal_cross`, I have made sure we zero out solutions = to the multiplicity of the surface, and added additional logic to make sure we get the order right so as to not mess up the block diagonal of the matrix (i.e. make sure if we sub out the n=2, m=4 resonant solution, we are subbing in the n=2, m=4 asymptotics instead of n=1, m=2). This section will get further cleanup if #89 is completed. 

## Integration
A lot of this has remained the same (with subbing `numpert_total` in places where `mpert` was used. I had to adjust some of the integration limits logic to include multiple resonant surfaces, e.g. choosing `psimax = psifac_res - singfac_min / (n * q')` since there are multiple ns now; I always tried to take the conservative approach by choosing the smallest n which integrated the farthest from the rational.  

## Output
In addition to adding a new example that shows how to use multi-n and compares it to the single n cases, I added a new output called `mn_index` that can be used to easily determine which index of the solution matrix corresponds to what mode. Each row of `mn_index` corresponds to the index of `u`, and the two columns are the values of `m,n` at this index. For example, if you wanted the `2, 1` mode and `mn_index[10, :] = [2, 1]`, then you should index `xi_psi[10, 1, :]` for the least stable modes 2,1 component. This is demonstrated in the example notebooks. 

## Free boundary
The primary change here is running VACUUM `npert` times to fill in the block diagonals of the full MN x MN `wv` matrix. This is the only section of the code that can't be generalized to 3D, since clearly we'd need a 3D vacuum code. Outside of that, it's just some minor reorganization/cleanup


jhalpern30:
@logan-nc I think this is all set for review, but no rush to do this. Just wanted to get it into a "final state" before APS so we can show it to Jong-Kyu if he is interested

matt-pharr:
Wow, this looks really cool!

src/DCON/Main.jl
```diff
@@ -87,23 +87,44 @@ function Main(path::String)
         write_output(outp, :dcon_out, @sprintf("%4s %12s %12s %12s %12s %12s %12s %12s %12s", "ipsi", "psifac", "f", "mu0 p", "dvdpsi", "q", "di", "dr", "ca1"))
     end
 
+    # Determine toroidal mode numbers
+    if ctrl.nn == 0 && ctrl.nn_low == ctrl.nn_high # single n mode
```
logan-nc: Whilst I admire your instincts towards backwards compatibility, we are in the magical realm of pure, unreleased, unversioned development. This means we are free to make things as clean and clear for the users as possible without worrying about backwards compatibility. Two ways to set n and secret `nn==0` conditions to toggle them is not clear to users. Lets just fully commit to the `nlow` and `nhigh`. Users can easily set `nlow==nhigh` and get the original single-n behavior that they would intuitively expect.

TLDR: Remove `nn` from DCON TOML

logan-nc:
I see the many added examples and how the notebooks can then benchmark the single n and multi-n... but It feels like there should be at least some new actual "test" so when others are developing the github merge hooks confirm they have not broken multi-n capabilities.

@matt-pharr, what is the status of our having github require that some quick examples run through to allow a merge? Should we add a quick multi-n example?

src/DCON/Free.jl
```diff
@@ -287,12 +292,11 @@ Performs the same function as `free_write_msc` in the Fortran code, except we wi
...
 """
-function free_write_msc(psifac::Float64, ctrl::DconControl, equil::Equilibrium.PlasmaEquilibrium, intr::DconInternal; inmemory_op::Union{Bool,Nothing}=nothing,
-    ahgstr_op::Union{String,Nothing}=nothing)
+function free_write_msc(psifac::Float64, n::Int, equil::Equilibrium.PlasmaEquilibrium, intr::DconInternal)
 
     # Defaults for optional arguments
-    inmemory = isnothing(inmemory_op) ? false : inmemory_op
-    ahgstr = isnothing(ahgstr_op) ? "ahg2msc_dcon.out" : ahgstr_op
+    # inmemory = isnothing(inmemory_op) ? false : inmemory_op
+    # ahgstr = isnothing(ahgstr_op) ? "ahg2msc_dcon.out" : ahgstr_op
```
logan-nc: cruft

src/DCON/Main.jl
```diff
@@ -112,24 +133,25 @@ function Main(path::String)
...
             println("     betat = $(equil.params.betat), betan = $(equil.params.betan), betap1 = $(equil.params.betap1)")
-            println("     nn = $(ctrl.nn), mlow = $(intr.mlow), mhigh = $(intr.mhigh), mpert = $(intr.mpert), mband = $(intr.mband)")
-            println(" Fourier analysis of metric tensor components")
+            println("     mlow = $(intr.mlow), mhigh = $(intr.mhigh), mpert = $(intr.mpert), mband = $(intr.mband)")
+            println("     nlow = $(intr.nlow), nhigh = $(intr.nhigh), npert = $(intr.npert)")
         end
 
         if outp.write_dcon_out
-            write_output(outp, :dcon_out, @sprintf("\n   mlow   mhigh   mpert   mband   nn   lim_fl   dmlim      qlim      psilim"))
+            write_output(outp, :dcon_out, @sprintf("\n   mlow   mhigh   mpert   mband   nlow   nhigh   npert   lim_fl   dmlim      qlim      psilim"))
```
logan-nc: This ascii write is all deprecated now right?
@jhalpern30 please merge develop into this branch so I don't mess up the merge conflicts when approving and merging this branch into develop. TY

src/DCON/Sing.jl
```diff
@@ -239,96 +269,114 @@ in which matrices have only their lower triangle stored to avoid confusion.
 
 Check third derivative accuracy in cubic splines or determine if it matters
 Better way to unpack the cubic splines
+Rename variables to be more intuitive? I don't like ff - maybe f and f_fact instead of f_lower
```
logan-nc, line 261: Are you keeping track of all these little todos @jhalpern30?
I feel like they are adding up 😅

src/DCON/Sing.jl
```diff
@@ -239,96 +269,114 @@ in which matrices have only their lower triangle stored to avoid confusion.
...
+    f0_lower = zeros(ComplexF64, intr.numpert_total, intr.numpert_total)
+    ff_lower = zeros(ComplexF64, intr.numpert_total, intr.numpert_total, ctrl.sing_order + 1)
+    g_lower = zeros(ComplexF64, intr.numpert_total, intr.numpert_total, ctrl.sing_order + 1)
+    k = zeros(ComplexF64, intr.numpert_total, intr.numpert_total, ctrl.sing_order + 1)
+    v = zeros(ComplexF64, intr.numpert_total, 2 * intr.numpert_total, 2)
+    x = zeros(ComplexF64, intr.numpert_total, 2 * intr.numpert_total, 2, ctrl.sing_order + 1)
 
     # Evaluate cubic splines
-    # TODO: third derivative has some error, but only included via sing_fac[ipert0] for sing_order < 3. Tests with solovev ideal indicate little sensitivity
+    # TODO: third derivative has some error, but only included via sing_fac[ipert_res] for sing_order < 3. Tests with solovev ideal indicate little sensitivity
```
logan-nc: define "has some error". Do you mean difference with the fortran?

logan-nc (changes requested):
Overall this looks great. Just a few minor changes requested. Biggest are:
- Deprecate `nn` fromt he TOML
- Add some sort of new "test" (not just "examples"). In general, I think I have to start demanding tests in all PRs or they'll never happen. Sorry, but its for the best in the long run!

That asymptotic section is nasty, but the many added annotations are a great help. I support the TODO that proposes cleaning it up by storing a simpler F matrix, but would keep it as a back-burner for now since what is there does indeed work ok.

matt-pharr:
@logan-nc the framework is already set up and implemented, nothing needs to change on the github side. Someone simply needs to add a test in Julia that runs the example through.

logan-nc:
@matt-pharr can you add the super fast ideal solovev run we use in the Fortran gpec repo? then I can make @jhalpern30 add a muti-n version here 😜

src/DCON/Main.jl
jhalpern30: While I will lament the time I spent trying to cover all the possible cases of nnlow/nnhigh/nn, I will enjoy getting rid of this. 

Do you think ultimately we'll want a mode where you can set nlow/nhigh based on resonances like mlow and mhigh with a delta_nhigh/delta_nlow parameter? Since we discussed giving the user the option to manually set mlow and mhigh instead of delta_mhigh/delta_mlow, I could see us adding this same flexibility for both m and n. Or eventually making both just a user set mlow mhigh and nlow nhigh. I can make whatever we decide a github issue, but in the short term this will affect at how I structure nlow and nhigh in this branch (i.e. leaving it in ctrl like nn where it is a fixed user input, or making it process into intr like mlow/mhigh where it is computed from user inputs.)

jhalpern30:
> * Add some sort of new "test" (not just "examples"). In general, I think I have to start demanding tests in all PRs or they'll never happen. Sorry, but its for the best in the long run!

Agreed that this is necessary and I thought about it, but since I didn't add any new functions I wasn't sure what to add to the tests. If @matt-pharr adds this full run through test I can definitely have a multi-n example.

src/DCON/Sing.jl
jhalpern30: Idt if this needs to remain a TODO. I do mean error relative to the Fortran splines, but I think its just numerical or so minor that it doesn't matter. Also, I did some looking and the Julia spline package `Interpolations.jl` doesn't even have third derivative capabilities so some of this will need to be simplified. The third derivatives are only used here and in a small search algorithm in the equilibrium solver, so basically we'll just restrict `sing_order` to be less than (I think) 3 and we're good. I can just remove this line

src/DCON/Sing.jl
jhalpern30: I would make the argument that this entire function is one big TODO lol

I didn't think they deserved to be separate github issues - the ones that are just comments in the code are just half-baked comments that I didn't want to get lost. I think once the major ones (i.e. a github issue) are done, then some of these subissues will either become irrelevant or become higher priority to become an issue.

src/DCON/Main.jl
jhalpern30: Yes, agreed. I knew this was going to be a complicated merge with the other branches and figured it'd be done last. I'll do this merge. 

Just wanted to separate into different PRs because it would be so big, but as a result you get some stuff like this

src/DCON/Free.jl
jhalpern30: Same as below - temporary comments for stuff that got removed in #91 but was originally in this branch. Will remove in merge

src/DCON/Main.jl
logan-nc: IDK how we'd set n. We need n to select m. I suppose we could default n based on the original equilibrium's `n` components, but I really don't think this would save the user much brain power. More likely, it will hurt their brain trying to figure out what is secretly going on. So for now, keep it as explicit user inputs.

Note that Glasser's DCON3D paper seemed to imply optimization of included (m,n) modes was important for performance. Let's basically ignore that for now and hope Julia does better. We can always make fancy (m,n) subset selections later if needed.

src/DCON/Sing.jl
logan-nc: 👍 Sounds good. Just clarify the comment or remove it. Up to you

logan-nc:
I think @jhalpern30 could probably add a run-through test with maybe just a little guidance from @matt-pharr.
In the fortran code, we use a .github/workflows yaml to do this like so:

https://github.com/PrincetonUniversity/GPEC/blob/f06e6abde3138ac8d9f42e58e56417b9d876b35f/.github/workflows/make_test.yml#L178-L184

I assume we should do the same here. Is that right @matt-pharr? Or do you want to use `test/runtests.jl` to do this kind of thing in the Julia repo?

jhalpern30:
Full run through tests have been added, nn has been deprecated, and I did a (very large) merge with develop and double checked the notebooks remain unchanged.

One thing I'll note - in doing some of the testing for this, I found that a lot of cases raise the non-Hermitian flag added in #91 for determining crit, and so in a lot of cases the terminal is populated with these warnings. I am unsure of whether this should remain a warning (i.e. something is wrong with your run), or this is just unnecessary and symmetrizing the matrix is an ok fix. This obviously does not relate to this PR, but I'll raise a github issue to address this in the future since it appears when running something as simple as the DIIID ideal example, so users will get a this warning on one of our examples, which doesn't seem right.

logan-nc (approved):
Looks good. Thanks for the quick adds. I think the full runs are going to be super useful!

Have we confirmed they prevent merging if they fail?

PR #78, opened 2025-10-29
Deprecating sort type and refactoring ureal building

jhalpern30:
## Summary
This PR addresses a two main issues:
1. Deprecation of the `sort_type` parameter, which I found to be confusing and without an obvious use case for the current version of the code. More detail below
2. Updates to the way we build and dump the true eigenfunctions for postprocessing including: a small bugfix, adding the equivalent u1->u4 that will eventually be used by GPEC, modifying the `odet.u_store` and `odet.ud_store` arrays in-place instead of creating a new one for ureal and dumping only these values, and updating the postprocessing scripts to only use the transformed `u`s that are dumped only. This addresses some comments that came up during the Hackathon regroup last week

## `sort_type`
This parameter affects the initialization of the `u` matrix. In the original DCON paper, it states that the array is initialized to the identity, but sort type actually initializes it to a permutation of the identity matrix, with columns arranged based on the desired method ('|m|', '|m - m_sing'|). This is analagous to performing a fixup on the solution but at the first step, and in theory, we could prepend this permuation matrix on the fixfac array and recover identical solutions after building the transforms after integration. As a result, the initial condition is slightly modified, but the code still gives the same results, just with a different set of eigenvectors representing the solution. This was likely added for some vestigial plotting routine that predates IPEC, and there is no documentation that explains when to use this parameter. As such, I removed it from the code. 

As a visualization , I ran the code with both `sort_type="absm"` and just initializing U_22 to be the identity matrix. As expected (since it is sorted by increasing eigenvalues), the `wt` matrix is unaffected
image
but the `u` matrices are different
image
However, once we actually form the eigenfunctions using the energy matrix, we get the same functions
image
I believe chatGPT gets the linear algebra reasons for this occurring mostly right
image
Because we ultimately care about the reformed eigenfunctions (u1, u2, etc.) which as shown above are identical, I think this change of basis is ok to ignore for clarity in the code (but would appreciate feedback from @logan-nc on if there will be any ramifications in GPEC).


src/DCON/Ode.jl
```diff
@@ -296,24 +293,9 @@ function ode_axis_init!(odet::OdeState, ctrl::DconControl, equil::Equilibrium.Pl
...
-        if intr.msing > 0
-            key .-= intr.sing[1].m
-        end
-        @. key = -abs(key)
-    else
-        error("Cannot recognize sort_type = $(ctrl.sort_type)")
-    end
-    index = sortperm(key; rev=true) # in original Fortran: bubble(key, index, 1, mpert)
-
     # Initialize solutions
```
logan-nc:
```suggestion
    # Initialize solutions with the identity matrix for U_22 as described in [Glasser PoP 2016] Section VI
```

src/DCON/Ode.jl
```diff
@@ -162,7 +162,7 @@ function ode_run(ctrl::DconControl, equil::Equilibrium.PlasmaEquilibrium, ffit::
         trim_storage!(odet)
     end
 
-    ureal = build_ureal(intr, odet)
+    transform_u!(odet, intr)
```
logan-nc, line 166:
```suggestion
    # form the true solution vectors, undoing the Gaussian reduction applied during fixups throughout the integration
    transform_u!(odet, intr)
```

logan-nc (changes requested):
This seems like a good move to me for the clarity of the code.
I do not know of anyone actively using the sorting features.

I did make a few minor requests for annotation, but once those are added this is good to go.

logan-nc (approved):

matt-pharr:
@jhalpern30 Since we are effectively doing a 'change-of-basis' here do you think it would be reasonable to look into putting ureal into the basis of having the same edge as u? That way we could update this confusing snippet:

```
uedge = u[:,:,1,end] \ wt[:,1]

# Create Julia eigenfunctions
u1s = zeros(ComplexF64, (mstep,mpert))
for istep in 1:mstep
    u1s[istep, :] = ureal[istep, :, :] * uedge
end
```

Specifically, I suggest solving `A * ureal(end,:,:) = u(:,:,1,end)` and then redefining `unew(istep,:,:) := A * ureal(istep,:,:)`

Then we could deprecate u and ureal and just have one output solution basis that is both fixfac-ed and has the appropriate edge values to do perturbed equilibrium calculations. We would need to test that this still creates a valid basis, but I think it is worth it for the simplicity. 

Edit: Looks like I commented too late! I will make an issue for this.

jhalpern30:
> @jhalpern30 Since we are effectively doing a 'change-of-basis' here do you think it would be reasonable to look into putting ureal into the basis of having the same edge as u? That way we could update this confusing snippet:
> 
> ```
> uedge = u[:,:,1,end] \ wt[:,1]
> 
> # Create Julia eigenfunctions
> u1s = zeros(ComplexF64, (mstep,mpert))
> for istep in 1:mstep
>     u1s[istep, :] = ureal[istep, :, :] * uedge
> end
> ```
> 
> Specifically, I suggest solving `A * ureal(end,:,:) = u(:,:,1,end)` and then redefining `unew(istep,:,:) := A * ureal(istep,:,:)`
> 
> Then we could deprecate u and ureal and just have one output solution basis that is both fixfac-ed and has the appropriate edge values to do perturbed equilibrium calculations. We would need to test that this still creates a valid basis, but I think it is worth it for the simplicity.
> 
> Edit: Looks like I commented too late! I will make an issue for this.

Note this has been updated as of this commit - the line is now
```julia
# Construct uedge coefficients using the most unstable mode
uedge = u[:, :, 1, end] \ wt[:,1]

# Create eigenfunctions
u1s = zeros(ComplexF64, (mstep,mpert))
for istep in 1:mstep
    u1s[istep, :] = u[:, :, 1, istep] * uedge
end
```
@matt-pharr Are you saying something in additional to this? I thought the uedge step is necessary since you need to choose which mode to form the eigenfunctions for

matt-pharr:
I see now! I didn't realize this was in this PR. Thanks!

PR #80, opened 2025-10-30
Normalizing solutions by Wt before dumping

jhalpern30:
This PR adds the feature Nik, Matt, and I discussed of prescaling the energy eigenvectors by Wt before dumping to `euler.h5`, allowing extremely easy plotting of results. I also modified the names of the dump parameters to the file, replacing them with their physical meanings instead of u/ud. (@matt-pharr @logan-nc I could not find an interpretation for `u2` in `gpec.f` - in the DCON paper, they're called the conjugate momenta of the displacements, but they don't really seem to be used in GPEC? Right now, I just have the name as `u2` in the HDF5 dump as a placeholder.)

All updates are done in place on `odet.u` and `odet.ud`, so they now effectively hold the physical `u1-u4` arrays - I will leave it up to @matt-pharr to determine how to best pass these into GPEC, I saw that you already have some structs going in your IPEC PR, so this should integrate nicely with that.

Another thing I wasn't sure about was why we scale Wt by `2pi * psio * 1e-3` - I'm guessing this is just a normalization inherent to the DCON calculations, but if there's a succinct explanation it would probably be good to add it as a comment to that line in `Free.jl`

With this new feature, postprocessing is now as simple as
```julia
eh5 = h5open("euler.h5", "r")
mlow = read(eh5["info/mlow"])
xi_psi = read(eh5["integration/xi_psi"])
psifac = read(eh5["integration/psi"])
close(eh5)
``` 
then
image
The `xi_psi` indices correspond to the poloidal mode number, energy eigenvalue ordered eigenvector, and integration step.

For comparison, before this we had to manually scale the eigenvectors by the edge coefficients, as shown below:
```julia
eh5 = h5open("euler.h5", "r")
mlow = read(eh5["info/mlow"])
mpert = read(eh5["info/mpert"])
mstep = read(eh5["integration/nstep"])
u = read(eh5["integration/u"])
psifac = read(eh5["integration/psi"])
psio = read(eh5["equil/psio"])
wt = read(eh5["vacuum/wt"])
et = read(eh5["vacuum/et"])
close(eh5)

# scale energy eigenvector matrices
chi1 = 2π*psio
wt = wt*(chi1*1e-3)
println("Done reading euler.h5")
```
then
image
(clearly more complicated and requiring expert knowledge of the outputs to interpret)


benchmarks/DIIID_ideal_example/euler_fortran_julia_comparison.ipynb
```diff
@@ -11,7 +11,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
```
logan-nc:
```suggestion
   "execution_count": null,
```

benchmarks/DIIID_ideal_example/euler_fortran_julia_comparison.ipynb
```diff
@@ -37,10 +37,20 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
```
logan-nc:
```suggestion
   "execution_count": null,
```

benchmarks/DIIID_ideal_example/euler_fortran_julia_comparison.ipynb
```diff
@@ -37,10 +37,20 @@
...
+    {
+     "data": {
+      "text/plain": [
+       "idcon_transform (generic function with 1 method)"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
```
logan-nc:
```suggestion
   "outputs": [],
```

benchmarks/DIIID_ideal_example/euler_fortran_julia_comparison.ipynb
```diff
@@ -136,10 +146,18 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
```
logan-nc:
```suggestion
   "execution_count": null,
```

benchmarks/DIIID_ideal_example/euler_fortran_julia_comparison.ipynb
```diff
@@ -136,10 +146,18 @@
...
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Done reading euler.bin\n"
+     ]
+    }
+   ],
```
logan-nc:
```suggestion
   "outputs": [],
```

benchmarks/DIIID_ideal_example/euler_fortran_julia_comparison.ipynb
```diff
@@ -320,10 +338,231 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 4,
```
logan-nc:
```suggestion
   "execution_count": null,
```

benchmarks/DIIID_ideal_example/euler_fortran_julia_comparison.ipynb
```diff
@@ -320,10 +338,231 @@
...
+      "Computing free boundary energies\n",
+      "   > Vacuum code overriding r 1.00E+00 from vac.in, to be 0\n",
+      "Energies: plasma = -1.6151697811249337, vacuum = 2.5086609379291636, real = 0.893491156804205, imaginary = 7.565644890758302e-5\n",
+      "All free-boundary modes stable for nn = 1.\n",
+      "----------------------------------\n",
+      "Run time: 16.885196208953857 seconds\n",
+      "Normal termination.\n"
+     ]
+    }
+   ],
```
logan-nc:
```suggestion
   "outputs": [],
```

examples/DIIID_ideal_example/run_and_analyze.ipynb
```diff
@@ -95,23 +91,13 @@
...
     "p = plot()\n",
     "for m in 1:5\n",
-    "    plot!(psifac, imag.(u1s[:,m - mlow + 1]), label=\"m=$m\")\n",
+    "    plot!(psifac, imag.(xi_psi[m - mlow + 1, 1, :]), label=\"m=$m (prescaled)\")\n",
     "end\n",
     "xlabel!(L\"\\psi_N\")\n",
     "ylabel!(L\"\\mathrm{Im}(\\xi_\\psi)\")\n",
-    "title!(\"Least Stable Eigenmode \" * L\"\\xi_\\psi\" * \" for \" * \"m=1-5\")\n",
-    "display(p)"
+    "title!(\"Least Stable Eigenmode \" * L\"\\xi_\\psi\" * \" for \" * \"m=1-5\")"
```
logan-nc, line 100: did you mean to remove the `display(p)` line?

examples/DIIID_ideal_example/run_and_analyze.ipynb
```diff
@@ -95,23 +91,13 @@
...
-    "# Create eigenfunctions\n",
-    "u1s = zeros(ComplexF64, (mstep,mpert))\n",
-    "for istep in 1:mstep\n",
-    "    u1s[istep, :] = u[:, :, 1, istep] * uedge\n",
-    "end\n",
-    "\n",
     "p = plot()\n",
     "for m in 1:5\n",
-    "    plot!(psifac, imag.(u1s[:,m - mlow + 1]), label=\"m=$m\")\n",
+    "    plot!(psifac, imag.(xi_psi[m - mlow + 1, 1, :]), label=\"m=$m (prescaled)\")\n",
```
logan-nc:
```suggestion
    "    plot!(psifac, imag.(xi_psi[m - mlow + 1, 1, :]), label=\"m=$m\")\n",
```

logan-nc: "Pre-scaled" doesn't mean anything to a new user. It is an insider developer comment, so more appropriate for a source annotation than an example plot label.

What could be useful in the example is a note on what units the xi have. The eigenmodes of Wt are unit normalized in terms of |\xi| on the plasma boundary. If all the chi's and 2pi's and such are included like GPEC's idcon.f then the eigenvalues correspond to joules per meter of edge displacement magnitude (vector norm of edge displacement spectrum).

examples/DIIID_ideal_example/run_and_analyze.ipynb
```diff
@@ -65,13 +65,9 @@
...
-    "mstep = read(eh5[\"integration/nstep\"])\n",
-    "u = read(eh5[\"integration/u\"])\n",
+    "xi_psi = read(eh5[\"integration/xi_psi\"])\n",
     "psifac = read(eh5[\"integration/psi\"])\n",
-    "psio = read(eh5[\"equil/psio\"])\n",
     "wt = read(eh5[\"vacuum/wt\"])\n",
-    "et = read(eh5[\"vacuum/et\"])\n",
     "close(eh5)\n",
     "\n",
     "# scale energy eigenvector matrices\n",
```
logan-nc, line 73: Is this still an appropriate annotation? I'm skimming the diff online so may be missing something. Are you scaling the fortran ones? We are no longer scaling Julia ones, so if its just julia then this is a deprecated annotation

examples/Solovev_ideal_example/run_and_analyze.ipynb
```diff
@@ -91,28 +87,17 @@
...
-    "u1s = zeros(ComplexF64, (mstep,mpert))\n",
-    "for istep in 1:mstep\n",
-    "    u1s[istep, :] = u[:, :, 1, istep] * uedge\n",
-    "end\n",
-    "\n",
-    "# Plot results\n",
     "p = plot()\n",
     "for m in 1:5\n",
-    "    plot!(psifac, imag.(u1s[:,m - mlow + 1]), label=\"m=$m, absm sorting\")\n",
+    "    plot!(psifac, imag.(xi_psi[m - mlow + 1, 1, :]), label=\"m=$m (prescaled)\")\n",
```
logan-nc:
```suggestion
    "    plot!(psifac, imag.(xi_psi[m - mlow + 1, 1, :]), label=\"m=$m\")\n",
```

src/DCON/Free.jl
```diff
@@ -173,6 +173,15 @@ function free_run(ctrl::DconControl, equil::Equilibrium.PlasmaEquilibrium, ffit:
         VacuumMod.unset_dcon_params()
     end
 
+    # Normalize eigenvectors based on scaled wt
```
logan-nc, line 176: If this is from idcon.f it's trying to get it into physical units joules per meter

logan-nc: Just to be clear, how do you avoid the free calculations done in the `psi_edge`workflow doing this multiple times as it checks dW(psi) in the edge? We don't want to multiple by `(2π * equil.psio * 1e-3)` multiple times and I imagine the reordering of the eigenmodes could possibly mess up the integration too somehow unless we get very lucky

src/DCON/Ode.jl
```diff
@@ -178,8 +178,12 @@ function ode_run(ctrl::DconControl, equil::Equilibrium.PlasmaEquilibrium, ffit::
             euler_h5["integration/nstep"] = odet.step
             euler_h5["integration/psi"] = odet.psi_store
             euler_h5["integration/q"] = Spl.spline_eval(equil.sq, odet.psi_store, 0)[4]
-            euler_h5["integration/u"] = odet.u_store
-            euler_h5["integration/ud"] = odet.ud_store
+            if !ctrl.vac_flag # we normalize by wt before dumping if calling free_run
```
logan-nc, line 181: This is probably just the fact that I am looking at a limited view of the file in the online diff, but I expected an else here... can you remind me what happens if `vac_flag` if false? I would think the normalization is based on Wt, which optionally has Wv ... so it wouldn't need to care about vac_flag

logan-nc (changes requested):
Love this change to the code. Definitely a step in the right direction to having a much more user friendly and clear tool. 

Please just clear the ipynb outputs and answer a few clarification questions. Thanks!

src/DCON/Free.jl
jhalpern30: > If this is from idcon.f it's trying to get it into physical units joules per meter

Gotcha, I think @matt-pharr added this line in the first iteration of this and I wasn't sure where it came from but it looks like it comes straight out of `idcon_read`. Perhaps it would be good to add this entire series of normalization before doing the dump/moving to GPEC so there's no conversions necessary outside of the source code? I think Matt had added in the wt scaling since its needed for the displacements but they all might be needed at some point.

```fortran
c-----------------------------------------------------------------------
c     normalize plasma/vacuum eigenvalues and eigenfunctions.
c-----------------------------------------------------------------------
      et=et/(mu0*2.0)*psio**2*(chi1*1e-3)**2
      ep=ep/(mu0*2.0)*psio**2*(chi1*1e-3)**2
      ee=REAL(et-ep,r8)
      wt=wt*(chi1*1e-3)
      wt0=wt0/(mu0*2.0)*psio**2
```

> Just to be clear, how do you avoid the free calculations done in the `psi_edge`workflow doing this multiple times as it checks dW(psi) in the edge? We don't want to multiple by `(2π * equil.psio * 1e-3)` multiple times and I imagine the reordering of the eigenmodes could possibly mess up the integration too somehow unless we get very lucky

The `psi_edge` workflow calls `free_compute_total`, not `free_run`, so this logic only occurs once when it is called in `Main.jl`, basically the last thing that gets performed before DCON ends and we pass everything into GPEC.

src/DCON/Ode.jl
jhalpern30: If `vac_flag` is true, then `free_run` gets called, triggering the rescaling logic before we dump. In `Main.jl`:
```julia
    if ctrl.vac_flag && !(ctrl.ksing > 0 && ctrl.ksing <= intr.msing + 1)
        if ctrl.verbose
            println("Computing free boundary energies")
        end
        plasma1, vacuum1, total1 = free_run(ctrl, equil, ffit, intr, odet, outp; op_netcdf_out=false) # outp.netcdf_out)
    end
```
So my logic was: if `vac_flag = false`, there's no `Wt` to scale against so just dump the unormalized values after integration, and if `vac_flag = true`, then we need to do the rescaling later in `free_run` and dump then. But looking back at it, perhaps naming them `xi_psi` is incorrect since they aren't scaled to be physical? Let me know if you have better ideas for this. Should I scale it by `Wp` instead?

examples/DIIID_ideal_example/run_and_analyze.ipynb
jhalpern30: You're correct, I had accidentally left the "pre-scaled" in from when I was making the figure for the PR. I'll take a look at adding the units, I think this relates to my below comment on if we should do all the `idcon.f` scalings before dumping.

examples/DIIID_ideal_example/run_and_analyze.ipynb
jhalpern30: This was for scaling `Wt` to plot the DCON summary-like figure, since I didn't actually save the `Wt` rescaling in `Free.jl`, I just used it to rescale `u`. Same as the rest of this, if we add all the rescalings to `Free.jl`, then this can actually be removed

src/DCON/Ode.jl
logan-nc: Yep, I concur. I got turned around before

src/DCON/Free.jl
logan-nc: Ok. This checks out. I vote you make a new issue and/or PR for metric unit normalizations of energies. Let's not do it here and risk mixing things up. For reference, I think the normalizations above are all just done so dW's reported are order 1. That is nice for intuition, but not for doing more physics (like plasma response calculations) where units matter. I'd be happy if our code output the metric values and we can include a `dcon_normalization` for convinience if people want to compare by plotting `et[1]*dcon_normalization` to get something order 1.

logan-nc (approved):

Issue #53, opened 2025-10-13
res_flag = true implementation


jhalpern30:
Implement all DCON code associated with the res_flag = true condition

jhalpern30:
Should have closed this with #81

PR #81, opened 2025-10-31
Res_flag deprecation

jhalpern30:
A true deprecation extravaganza! I didn't realize how deep this went, and it cleans up so much code/a lot of my comments. This PR initially was meant to just deprecate `res_flag`, which, after discussions with @matt-pharr and @logan-nc, appeared to be old code that was a first attempt at resistive physics in DCON but eventually branched off into RDCON and is now unused. Given our plan to overhaul how resistive physics is performed in Julia, this is a first step towards that by providing a clean slate, partially addressing #53.

Here is a list of the deprecated parameters in this PR, and I will go into more detail below for reference if this ever comes up again:

1. `res_flag` (input flag to DCON_CONTROL)
2. `ode_resist_cross` (function)
3. `ode_test` (function)
4. `flag_count` (variable in `Ode.jl`)
5. `singfac_max` (input value to DCON_CONTROL)
6. `msol` (variable in `Ode.jl` and `Sing.jl`)
7. `test` (input to `ode_fixup`)
8. `resist_eval` (function in `resist.f` that wasn't converted yet but left as a TODO)
9. `ResistType` (placeholder Julia type that would have been filled by `resist_eval`)

Most of this PR review will probably be reviewing my logic below for deprecating these things, since most of the changes were just deleting lines of code/replacing variable names. I ran the example cases on both the develop branch and this branch and got the same answers.

### `res_flag`
This is an input flag in the `dcon.toml` file. If true, it did two things: call `ode_resist_cross` when at a rational surface, and initiate some additional logic in `ode_test`. This naturally led to more deprecations, as it becomes clear that without `res_flag`, `ode_test` is only checking to see if we are at the end of integration, which we handle more cleanly in Julia. This logic included the `flag_count` and `singfac_max` variables, which I only see meaningfully used in this `res_flag = true` section of `ode_test`. Within `ode_resist_cross`, there are sections where the variable `msol` is incremented by 1 at each rational surface - everywhere else in the code, this variable is either `mpert`, or `2*mpert` in `sing_mmat`. With this parameter deprecated, that naturally led to the removal of 2, 3, 4, 5, and 6.

### `ode_resist_cross`
Self-explanatory, only called if `res_flag = true`

### `ode_test`
If `res_flag = false`, then the only logic in `ode_test` is
```fortran
c-----------------------------------------------------------------------
c     truncation test: lsode limit
c-----------------------------------------------------------------------
      flag = psifac == psimax .OR. istep == nstep .OR. istate < 0
c-----------------------------------------------------------------------
c     simple return.
c-----------------------------------------------------------------------
      IF(.NOT. res_flag .OR. flag
     $     .OR. ising > msing .OR. singfac > singfac_max)RETURN
```
Which always just returns since `.NOT res_flag = true`, so this function doesn't actually do anything

### `flag_count`
A quick search of the Fortran repo shows that `flag_count` only appears in the bottom of `ode_test` (outside of just setting it to zero), so without `ode_test` it can be removed.

### `singfac_max`
Same as `flag_count`

### `msol`
As mentioned above, the only logic in the code where `msol` is not equal to `mpert` or `2*mpert` is if `ode_resist_cross = true`, where it is incremented by 1 at each rational surface. This allows us to get rid of this variable and explicitly use `mpert` or `2*mpert` in the loop indices/allocations where `msol` was previously used. Furthermore, in Fortran, several arrays were allocated to `2*mpert` but did not actually use the whole thing, I am assuming just to leave space for this incrementing. Allowing for this added some confusing logic, specifically to functions like `ode_unorm` (the most obvious example of this), `ode_fixup`, and `sing_get_ca`. For example, at the beginning of `ode_unorm`:
```fortran
      unorm(1:mpert)=SQRT(SUM(ABS(u(:,1:mpert,1))**2,1))
      unorm(mpert+1:msol)=SQRT(SUM(ABS(u(:,mpert+1:msol,2))**2,1))
```
then
```fortran
         unorm(1:msol)=unorm(1:msol)/unorm0(1:msol)
```
Clearly, if mpert = msol, this becomes extremely simple.

### test
This wasn't actually related to `res_flag`, but I found it when searching for `ode_test`. Its only used for the function `ode_test_fixup`, which is just a debug function and `test` is never used outside of this.

### `resist_eval`
This was a hacked in function that filled in resistive data at the rational surface, and @matt-pharr had said that this won't be needed when we integrate resistive physics better in Julia.

### `ResistType`
This was planned to be filled by `resist_eval`


logan-nc (approved):
Good digging!

PR #83, opened 2025-11-04
Cleaning up some equilibrium code

jhalpern30:
This PR cleans up some code in the equilibrium module that I started when doing the detailed debugging for #73 and decided two split into two separate PRs. This one is pretty much entirely code cleanups that I did along the way, and does not directly fix any major bugs. My primary goal was to get rid of the excessive printing that was left over from the hackathon that would pollute the terminal when running the code (e.g. printing out the current psi value when processing all mpsi surfaces). This PR leaves the equilibrium printing in a minimal state more similar to the Fortran.

Outside of `DirectEquilibrium`, my changes were limited to removing unnecessary lines, adding/improving docstrings, adding flag checking for things like outputting the gse error etc. Within `DirectEquilibrium`, I did a bit more detailed restructuring to make the code more readable such as renaming variables, moving some variables within structs, on top of updating comments and other aesthetics. The only actual change was modifying some loop logic for the `direct_position` function (which makes the final output very slightly different) and changing the `eqfun` splines entries from `|B|^2` to `|B|`, which matches the docstring.

As a check, here is proof that the outputs for both Solovev are condensed but identical between develop and this branch:
This branch:
DIIID
image
Solovev
image
Develop:
DIIID
image
Solovev
image


notebooks/spline_bug_solovev_ideal.ipynb
logan-nc, line 1: Is this an intentional removal of the file?

logan-nc (approved):
Looks good to me. The code is getting cleaner by the day!

notebooks/spline_bug_solovev_ideal.ipynb
jhalpern30: Yes, it probably should have been deleted a while ago

PR #84, opened 2025-11-05
Ode Output Refactoring

### LINKED ISSUE #82: Reduce and rename ode_output_step calls
Following the spirit of #31, #81 included the following TODO:

```
# TODO: this logic can be updated - technically, if we save crit to memory,
    # we could do all these calculations in post processing if a stability calculation
    # is desired. If we keep it here, I'd rename to `check_for_zero_crossing` (what is
    # current `ode_output_monitor`). A lot of `ode_output.f` has been deprecated due to
    # in-memory integration
    ode_output_step(ctrl, equil, intr, odet, outp)
    # Compute total dW in edge region (if `psiedge < psilim`)
    # Same as above, this could technically be done in post processing if desired
    ode_record_edge_dW!(odet, ctrl, equil, ffit, intr)
```

I concur with this renaming. 
Looking closer, I see TODOs in `ode_step` about this `ode_output_step` as well:

https://github.com/OpenFUSIONToolkit/JPEC/blob/0eeba2ad4fb189d3872ea807ab55baddfccdaefe/src/DCON/Ode.jl#L517-L536

I think those are a quirk of the `euler_step` capability that skipped writing some steps to the bin file to save disk space. If I skipped writting steps right next to resonant surfaces, it changed the GPEC resonant field a little because it is calculated from jumps in the derivative of the field right on the left/right of the surface. I also needed the very last step to be in euler.bin or GPEC behavaed oddly since psilim didn't match the eigenmode grids, etc etc. So I had logic to make sure to write first, closest-to-sing-surf, and last step. All this is irrelevant no that we don't do the clunky step by step file IO. So I think we can get rid of these... but we should double check, test, etc.

jhalpern30:
This PR resolves #82 by refactoring how outputs are performed in DCON. This led to a rather large overhaul of many related functionalities, and is another extremely satisfying PR.

## Overhaul of computing crit
This included reorganizing the equivalent of `ode_output_monitor` and `ode_output_get_crit` into `check_for_zero_crossings` and `compute_smallest_eigenvalue`, respectively. We also now do this all after integration, calling `evaluate_stability_criterion` to evaluate all the crit data at once and then optionally dumping to `crit.h5`. This reorganization included deprecating `crit.out` in favor of `crit.h5`, so I decided to rename that file from `OdeOutput.jl` to `FixedBoundaryStability.jl` to more accurately reflect what it is computing. Note that this also deprecated `termbycross_flag`, which kills the integration if a zero crossing occurs, since we do everything post-integration now, along with several `_prev` variables in the crit monitoring. By deprecating `crit.out`, `ode_output_init` became only a write to `euler.h5`, so I also consolidated it and the `euler.h5` write that occurs after integration into one helper function `write_euler_h5` in `Ode.jl`, which further cleaned things up nicely.

I also updated the examples to show a crit vs. psi plot, and verified that it matches what was dumped to crit.out on the develop branch.

## Overhaul of edge truncation logic
Similar to computing crit, I rewrote the functionality to compute the edge dW in the region where `psiedge < psilim` all post-integration, now calling it `findmax_dW_edge`. The function remains mostly the same, except it is now more self contained and doesn't need the checks to see if the wv spline was made already. This is now performed in a single function right after integration, another nice simplification.

To check this, I redid the ARC betascan (which use psiedge logic) and confirmed I achieved the same results

## Adding u as a parameter to `ode_unorm` and `ode_fixup`
Previously, I had to add a copy of `odet.u = integrator.u` and back in the callback since I need odet.u to be updated when performing unorm, output_step, and record_edge, etc. However, with the latter two now gone, I deemed it beneficial to now pass in the u array directly to `ode_unorm` to update, since it removes two copies in this callback with only a little bit of added clutter to the functions. 

## Docstrings
All these new/modified functions have had their docstrings updated

## Utils.jl
I cleaned up this file, removing some unused functions and moving the storage related operations of `Ode.jl` to this file.


src/DCON/Ode.jl
```diff
@@ -494,16 +442,17 @@ the solution at the new point.
...
 
-    # Set up the callback to be run at every step
+    # Callback to be run at every step, handles fixups, tolerances, and data storage
     cb = DiscreteCallback((u, t, integrator) -> true, integrator_callback!)
 
     # Choose integration bounds
+    # TODO: Nik is this needed?! If not, this function becomes extremely clean
+    # I think this is effectively "only integrate to exact equilibrium nodes"
     if ctrl.node_flag # TODO: I can't find any mention of what this flag is for, can we get rid of it?
         while odet.psifac >= equil.sq.xs[ix]
```
jhalpern30, line 457: @logan-nc pinging you on this since I came across it when debugging and this is probably a relevant PR to get rid of it if no one uses this. I can't find any mention of it other than setting it to false by default in the Fortran.

logan-nc:
>I decided to rename that file from `OdeOutput.jl` to `FixedBoundaryStability.jl`
👏👏 

 >By deprecating crit.out, `ode_output_init` became only a write to `euler.h5`, so I also consolidated it and the `euler.h5` write that occurs after integration
👍 👍 

My question on this is why maintain a `crit.h5` different from the euler.h5? They are on the same grid, so wouldn't it be easy enough to include the crit profile in that `euler.h5` write? The `euler.h5` file already has everything else about the stability right? The eigenmodes, but also energies and plasma info like betan etc.? Can we just make it the one stop shop for force free states and stability outputs? Maybe it gets renamed `forcefreebasis_and_stability.h5` or something more descriptive? If we want a split, then I guess there should be `forcefreebasis.h5`, and `stability.h5`? At the very least, rename `crit.h5` to something more descriptive!

src/DCON/DconStructs.jl
```diff
@@ -97,7 +97,6 @@ end
     set_psilim_via_dmlim::Bool = false # previously sas_flag, if true, determines psilim using outermost rational + dmlim
     dmlim::Float64 = 0.2 # % outside the last rational surface to go out to determine dW if set_psilim_via_dmlim is true
     sing_order::Int = 2
-    termbycross_flag::Bool = false
```
logan-nc, line 100: Just to flag this.. we will eventually want to terminate the plasma response calculations if there is a zero crossing. I added this flag after many instances of people (including me) submitting DCON+GPEC jobs, spending lots of time looking at results, seeing some funky stuff, and eventually realizing the whole thing was invalid due to a zero crossing instability. We want to avoid people accidentally blowing through terminal warnings and publishing bad results. Please add an `if` after the new crit calculation function call (but after all the file IO) that terminates the code if there is a zero crossing

src/DCON/Ode.jl
logan-nc: I've never seen this used. I've never even noticed it existed. Remove it.
I suspect this was someone's cruft from a debugging effort decades ago.

jhalpern30:
> > I decided to rename that file from `OdeOutput.jl` to `FixedBoundaryStability.jl`
> > 👏👏
> 
> > By deprecating crit.out, `ode_output_init` became only a write to `euler.h5`, so I also consolidated it and the `euler.h5` write that occurs after integration
> > 👍 👍
> 
> My question on this is why maintain a `crit.h5` different from the euler.h5? They are on the same grid, so wouldn't it be easy enough to include the crit profile in that `euler.h5` write? The `euler.h5` file already has everything else about the stability right? The eigenmodes, but also energies and plasma info like betan etc.? Can we just make it the one stop shop for force free states and stability outputs? Maybe it gets renamed `forcefreebasis_and_stability.h5` or something more descriptive? If we want a split, then I guess there should be `forcefreebasis.h5`, and `stability.h5`? At the very least, rename `crit.h5` to something more descriptive!

If you think this is something that wouldn't be too large to add on to this PR, I could see just having one HDF5 file as output (since they are already quite good at separating outputs) and name the subdirectories. I would also get rid of dcon.out, so a DCON run would only give you a single output file which could be inspected through the built in HDF5 visualization, and the whole DCON_OUTPUT section of the TOML could be gotten rid of. We'd just have a `write_outputs_to_HDF5` flag and a `fname_HDF5` to specify the filename (which would allow customization if running this many times in a script), defaulting to `forcefreebasis_and_stability.h5`. I would also combine the eqdata.h5 that is created in `Main.jl` into this file under a Mercier/Ballooning heading, since most of the entries are already dumped in the splines sections of the euler.h5 anyway.

If this would be too confusing to add to this PR, I can just rename crit.h5 -> stability.h5 and save it for a future PR.

src/DCON/DconStructs.jl
jhalpern30: Makes sense. One idea I have is to default it to true and put this condition right before the end of the `Main` function right before the code would normally print normal termination and then (once integrated) return the DCON solutions to some GPEC interface. I'll make the condition
```julia
if ctrl.termbycross_flag && odet.nzero != 0
    println("Fixed-boundary instability detected with termbycross_flag=true, terminating run before normal completion.")
    return
end
println("Run time: $end_time seconds")
println("Normal termination.")
return DCON_solutions
```
That way, the entire DCON code will run and produce all of its outputs, but it will return nothing if this condition is met so GPEC won't run. This way, the default true behavior doesn't affect the DCON run outside of adding a short erroring out statement at the end of the run. Then, in GPEC, we can add a brief check of "if DCON did not output nothing".

I would like to avoid adding an error throw or early exit condition to this section of the code because its not exactly an error in DCON (which solves for stability among other things, so it did its job), its more so a choice for GPEC. Another alternative is returning some stable/unstable flag to GPEC within the DCON solutions struct, and the GPEC logic itself just checks for unstable, and if so, doesn't run if this termbycross_flag = true. 

Open to thoughts on this

logan-nc: There will be no more DCON, GPEC separation. There will be one `Main.jl`. So yes, an `if odet.nzero != 0` after all the force free states are calculated is all we need. There is no need to re-introduce `termbycross_flag` - we will hardcode it to not bother calculating plasma response `if odet.nzero != 0` and leave it to some future developer argue why they need to override it if they ever do. 

TLDR: Just put this at the end of the current Main
```
# TODO: only continue to plasma response calculations if odet.nzero == 0
```

logan-nc:
Lets merge this since it has some nice restructuring of the callbacks to separate functions. I'll make a separate issue consolidating h5 and deprecating asciis.

logan-nc (approved):
More good cleanup!

src/DCON/DconStructs.jl
logan-nc: did it

PR #86, opened 2025-11-06
DCON - DEPRECATION - Deprecates node_flag

logan-nc:
This didn't seem to do anything. It forces integration to target each equil spline knot location in turn, but kept inrementing to the next until psilim was hit


jhalpern30 (approved):
Looks good

PR #91, opened 2025-11-11
Some general code cleanups

jhalpern30:
This is a PR of code cleanups I encountered while implementing multi-n but were not directly related to the implementation. A significant portion is spent on adding more verbose comments/directly linking code to equations in the DCON paper.

Additional cleanups:

- Deprecating some more unused stuff in `Free.jl`, including the `ahg` file interface, netCDF output, and `vac.bin` output
- Standardizing control logic for Newton/binary searches
- Making the `path` variable default to the current run directory
- Deprecating `m1` and `singfac` within the OdeState struct, they didn't serve a purpose and become confusing for multi-n

### DIFF
--- a/src/DCON/FixedBoundaryStability.jl
+++ b/src/DCON/FixedBoundaryStability.jl
@@ -104,19 +104,21 @@ in the Fortran code.
 """
 function compute_smallest_eigenvalue(psi::Float64, u::Array{ComplexF64,3}, sq::Spl.CubicSpline{Float64})
 
-    # Compute inverse plasma response matrix
-    # TODO: is this actually the inverse?
+    # Compute inverse plasma response matrix W = U₁ * U₂⁻¹ = adj(adj(U₂)⁻¹ * adj(U₁))
     wp_inverse = adjoint(u[:, :, 1])
     temp = adjoint(u[:, :, 2])
     wp_inverse = temp \ wp_inverse
 
-    # Symmetrize to be Hermitian
+    # Enforce that W is Hermitian
+    # TODO: Is this necessary? The DCON paper says W is Hermitian by construction.
+    # Couldn't this mask numerical issues we should be addressing directly?
+    # i.e. we should error out in the Hermitian eigenvalue solver if W is not Hermitian?
     wp_inverse .+= adjoint(wp_inverse)
     wp_inverse .*= 0.5
 
-    # Compute inverse eigenvalues, sort, and return the smallest (with a scale factor)
-    evalsi = eigvals!(Hermitian(wp_inverse))
-    indexi = sortperm(evalsi; by=abs)
+    # Compute eigenvalues, sort, and return the smallest (with a scale factor)
+    eig_vals = eigvals!(Hermitian(wp_inverse))
+    index = sortperm(eig_vals; by=abs)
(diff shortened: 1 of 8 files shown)


src/DCON/FixedBoundaryStability.jl
```diff
@@ -104,19 +104,21 @@ in the Fortran code.
...
     wp_inverse = adjoint(u[:, :, 1])
     temp = adjoint(u[:, :, 2])
     wp_inverse = temp \ wp_inverse
 
-    # Symmetrize to be Hermitian
+    # Enforce that W is Hermitian
+    # TODO: Is this necessary? The DCON paper says W is Hermitian by construction.
+    # Couldn't this mask numerical issues we should be addressing directly?
+    # i.e. we should error out in the Hermitian eigenvalue solver if W is not Hermitian?
     wp_inverse .+= adjoint(wp_inverse)
```
jhalpern30: @logan-nc do you have thoughts on this?

logan-nc: Yeah, I agree from a big picture standpoint. 

It looks like this is in the fortran from back before I brought it to git. I suspect it is in there because someone sometime in the 90s, 00's was running into a weird case where crit was getting hairy for "numerical reasons" (i.e. in known-to-be-stable plasmas) and this cleaned it up. It's a little scary to just remove though, as maybe Glasser added it from the beginning and things get hairy fast without it? Do you know anything about this @parkjk?

I propose @jhalpern30 add a warning/error if the non-hermition residual is significant but leave this in for this PR. If you want to remove it, do a separate PR and document that there is no meaningful change (which I would think their shouldn't be if it's passing the Hermitian tests). 

While we are here though, I really dislike the current form of `compute_smallest_eigenvalue`. This should not require `psi` and `sq`. The `* dVdpsi^2` part should be done in `check_for_zero_crossings!` for transparency and readability.

jhalpern30: > I propose @jhalpern30 add a warning/error if the non-hermition residual is significant but leave this in for this PR. If you want to remove it, do a separate PR and document that there is no meaningful change (which I would think their shouldn't be if it's passing the Hermitian tests).

Sounds good to me. I'll add this update. 
> 
> While we are here though, I really dislike the current form of `compute_smallest_eigenvalue`. This should not require `psi` and `sq`. The `* dVdpsi^2` part should be done in `check_for_zero_crossings!` for transparency and readability.

Agreed, I remembered thinking about the best way to do this when I was rewriting this and think I left it in because its called twice in `check_for_zero_crossings!` so it reduced the lines there, but it does unnecessarily pollute `compute_smallest_eigenvalue`. I'll push this fix

jhalpern30: Did a fairly major update on this function for cleanup, quick summary here:

- Pushed dVdPsi to `check_for_zero_crossings!`
- Simplified formation of `wp_inverse` using right division, added that note to the docstring
- Added the nonhermitian warning along with `hermitianpart!`, Julia's built in operator to do the W = 0.5 (W + W') operation :)
- simplified the eigenvalue return since it could be made into one line pretty easily

logan-nc (approved):
Great!

PR #92, opened 2025-11-12
Consolidating outputs

### LINKED ISSUE #85: Consolidate all force free state and stability outputs into one h5
As discussed in #84, there is no need for separate h5 files with the fixed boundary stability, free boundary stability and eigenmodes, and equilibrium information. This information can all be neatly structured within one file, making it easier to share/organize the outputs of code run instances.

Bonus Points: Add a input param called something like output_file_suffix that gets appended to the default file name. This will allow people to script multiple runs nicely without having to create deep directory structures saving similarly named files.

Optional: Should we store the input structures to this h5? If so, then one needs only share the one file and someone else can immediately rerun the same case. I find that often one has some outputs of a run that get separated from the inputs, causing much frustration if/when trying to restore the run to test a version update or some such.

jhalpern30:
This PR resolves #85 by consolidating all outputs into a single HDF5 file and deprecates `dcon.out`, reducing a lot of clutter throughout the code.

### Structs
- Deprecated the `DconOutput` struct and moved the only two important parameters, `write_outputs_to_HDF5` and `HDF5_filename` to `DconControl`. They default to `true` and `euler.h5`, respectively. We can change the default from `euler.h5` in the future if we rename this module to `ForceFreeStates` or something like that, but for now I like this since its not terrible and indicates to a new user that it replicates `euler.bin`.
- Added a `VacuumData` struct to store the quantities calculated in `Free.jl`. This was needed to dump all outputs in `Main.jl`, and `free_run!` now returns this struct instead of `plasma1, vacuum1, total1`. Going to ping @matt-pharr here since I modeled the struct after what was in the IPEC PR so it shouldn't be too difficult to merge later on. Most likely just adding the additional `g...` arrays that you have.
- Moved the `OdeState` struct from `Ode.jl` to `DconStructs.jl`, which should have been done a while ago. Also added the `crit_store` array to this struct and modified the `evaluate_stability_criterion` function to modify it in place since we are no longer writing `crit.h5` in that function and we need it when we write the HDF5 along with the other ODE data in `Main.jl`

### File output
- Removed all `dcon.out` writes, which also allowed me to remove the helper functions in `Utils.jl` that I was using to open/write/close text files.
- Consolidated all output writing into a single function `write_outputs_to_HDF5` now located in `Main.jl`. This is mostly the same as what was located in several sections of the code prior to this, with a few additional terms/logic to only write if the corresponding flag is true (i.e. `vac_data` if `vac_flag = true`)
- As mentioned in #85, we now directly write the `EQUIL_CONTROL` and `DCON_CONTROL` input files directly to the HDF5 to allow runs to be easily replicated if needed.  As a visual, here is what inputs in the HDF5 looks like
image
There's room to improve/generalize this in the future, but this is just a first attempt. Specifically, if we think it worth it, it would be pretty easy to add functionality to be able to run from an HDF5 output file by converting this group to a dict. Not sure exactly the use case here since you'd already have the outputs, but just an idea.


logan-nc (approved):
Incredible. This cleaned up so much code! It really starting to look like a proper, modern Julia code now 😍

logan-nc:
@jhalpern30 should double check [9bc73f6](https://github.com/OpenFUSIONToolkit/JPEC/pull/92/commits/9bc73f674c4395aab8c7287b498d457eedf5bf6b) since I just did it on the github web interface. 

The conflicts were almost all resolved by accepting this branch's changes. Note this clobbered the naming changes of `et` to `tot_eigvals`, `ep` to `plasma_eigvals`, `tt` to `temp_eigvals` etc in the vacuum parts. As for cases where I didn't just accept this branch's version, there was also an old `op_netcdf_out` argument in `free_run` I removed... I hope that is ok!

jhalpern30:
> @jhalpern30 should double check [9bc73f6](https://github.com/OpenFUSIONToolkit/JPEC/pull/92/commits/9bc73f674c4395aab8c7287b498d457eedf5bf6b) since I just did it on the github web interface.
> 
> The conflicts were almost all resolved by accepting this branch's changes. Note this clobbered the naming changes of `et` to `tot_eigvals`, `ep` to `plasma_eigvals`, `tt` to `temp_eigvals` etc in the vacuum parts. As for cases where I didn't just accept this branch's version, there was also an old `op_netcdf_out` argument in `free_run` I removed... I hope that is ok!

Yep your merge looks good  - there were definitely some overlapping things in the commits that I tried to keep separate but didn't get 100%

PR #102, opened 2025-12-18
Tests improvement

### LINKED ISSUE #101: Required tests for merging are taking too long
The tests in `.github/workflows/test.yml` take nearly 8 minutes to run. Most of the time seems to be because it adds and compiles all the required julia packages every time it runs a test. We need to speed the whole process up to be under 2 minutes to be a reasonable automatic action that people can wait for before merging.

matt-pharr:
See issue #101.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -19,23 +19,18 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up Julia
-        uses: julia-actions/setup-julia@v1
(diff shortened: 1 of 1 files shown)


matt-pharr:
@claude as it stands, the tests are running ok (no cache hit on the first run) but the 'build and test julia package' workflow in test.yaml is testing every single dependency, which is not necessary for us. We just want to run the tests for our own package. How do we do that?

Issue #108, opened 2025-12-19
The Chance 2007 updates to the vacuum code need to be implemented in julia


logan-nc:
The fortran code needs to be updated to use the equivalent of the fortran `vacuum_math.f` version of `aleg` instead of `aleg_old` as it does now. A stub exists in VacuumInternals.jl called `Pn_minus_half_2007` that should hold this implementation. 

This should be benchmarked for agreement with the fortran and `Pn_minus_half_1997`. As described in the fortran `aleg` annotations, the 2007 and 1997 implementations are expected to diverge at large nloc. A benchmark script should create a plot that confirms this behavior.

The new `Pn_minus_half_2007` should be benchmarked for speed compared to the `Pn_minus_half_1997` function.  

Once it is all implemented and checked, `uselegacygreenfunction` should be defaulted to false (and set to false in all example tomls).

logan-nc:
@claude address this

JaeBeom1019:
Closing this issue. We fully addressed this issue in #111, also in GPEC [#229] and have no concerns anymore! (at least for me)(https://github.com/PrincetonUniversity/GPEC/pull/229)  👍

PR #111, opened 2025-12-22
VACUUM - NEW FEATURE - Implement Chance 2007 Legendre function updates

logan-nc:
This commit implements the Chance 2007 updates to the vacuum code as described in J. Comp. Phys 221 (2007) 330-348.

Key changes:
1. Added elliptic_integrals_bulirsch() - Bulirsch's algorithm for computing complete elliptic integrals K(m1) and E(m1) with higher accuracy than the polynomial approximations in the 1997 version

2. Implemented Pn_minus_half_2007() - New Legendre function calculation with:
   - Bulirsch algorithm for base cases P^0 and P^1
   - 32-point Gaussian quadrature for large mode numbers (n*rhohat >= 0.1)
   - Upward recurrence for small mode numbers
   - Better numerical stability at large n compared to 1997 version

3. Changed default in green() from uselegacygreenfunction=true to false, making the 2007 implementation the default

4. Added comprehensive benchmark script (benchmarks/benchmark_legendre_2007.jl) to verify:
   - Agreement between 1997 and 2007 implementations at small n
   - Expected divergence at large n
   - Performance comparison
   - Method selection criteria

5. Updated tests in runtests_vacuum_julia.jl to test both implementations

The 2007 implementation is expected to be more accurate for large toroidal mode numbers where the 1997 upward recurrence becomes numerically unstable.


logan-nc:
@matt-pharr and @JaeBeom1019, I think you know more about this part of the fortran code than anyone. This is claude code's attempt to address #108 - if it's no good, then just say so. I doubt it is 100% correct, but maybe it gets us 90% of the way there? Do you think you can get it over the finish line without too much effort? Just let me know either way

JaeBeom1019:
I think those are almost perfect.
image

One thing missed here is that This vacuum version does not yet implement the full algorithm from the 2007 paper. The complex plane integration (Section 6) for the $n\hat\rho \gg 1$ regime remains to be implemented. 
image

**Current commit(2007 Forced gaussian)** : Gaussian 8 point for $n\hat\rho<0.1$, Gaussian 32 points for $n\hat\rho>0.1$
**2007 paper implement(Standard)** : Gaussian 8 point for $n\hat\rho<0.1$, Gaussian 32 points for $n\hat\rho<1.0$, Complex integrals for $n\hat\rho>1.0$

1. Current Status & Missing FeaturesThis vacuum version does not yet fully implement the complete algorithm described in the M.S. Chance (2007) paper. Specifically, the complex plane integration (Section 6 in paper) optimized for the $n\hat\rho \gg 1$ regime is currently pending integration. I will address this in the upcoming commit.
2. Numerical Strategy by Regime ($n\hat{\rho}$)Based on the convergence properties and efficiency, the integration strategy is divided as follows:$n\hat{\rho} < 0.1$: 8-point Gaussian quadrature.In this regime, the function is smooth enough that a lower-order quadrature is sufficient and efficient.$n\hat{\rho} \ge 0.1$: 32-point Gaussian quadrature.As $n\hat{\rho}$ increases, the integrand becomes more oscillatory/localized, requiring higher precision (32 points) to maintain accuracy.
3. Regarding the High-$n$ Regime ($n\hat{\rho} \gg 1.0$ maybe 10?)Series Expansion (Asymptotic using complex integrals): For very large $n\hat{\rho}$, the power series expansion converges faster than the 32-point quadrature. I have already implemented this logic and can commit it to this branch if deemed necessary.
4. Accuracy Consideration near $n\hat{\rho} \approx 1.0$:According to the paper and numerical tests, the 32-point Gaussian quadrature maintains better accuracy than the series expansion around $n\hat{\rho} \approx 1.0$.Therefore, the series expansion should likely be reserved for cases where $n\hat{\rho}$ is significantly larger than 1.0, or used as a fallback/check.

JaeBeom1019:
I benchmarked with two cases 

**Current commit** : Gaussian 8 point for $n\hat\rho<0.1$, Gaussian 32 points for $n\hat\rho>0.1$
**2007 paper implement** : Gaussian 8 point for $n\hat\rho<0.1$, Gaussian 32 points for $n\hat\rho<1.0$, Complex integrals for $n\hat\rho>1.0$


image


```
==================================================
Running Benchmark for s=5.0, n=170
Note: Standard method uses Series Expansion here.
      GaussianForced uses Numerical Integration.
==================================================

1. Testing Pn_minus_half_2007 (Standard)...

2. Testing Pn_minus_half_2007_GaussianForced (Forced)...

==================================================
Result Summary:
Standard time (median): 979.52 ns
Gaussian time (median): 1133.40 ns
Speedup Factor: Standard is 1.2x FASTER
==================================================
```

Like what paper said, **2007 paper implement** was 1.2 times faster than **Current commit** , but It's just 20% faster with  order of nano seconds. I think we can just move on not adding complex integrals for high n because the difference is quite significant ~10% and 2007 paper agrees that Gaussian 32 is more precise rather than complex integral results.

JaeBeom1019:
Fixed

- runtest toleratnce from 1.e-10 to 1.e-7 because those have different calculation method. (Bulirsch and ellipek)
- Add 32 points gaussian weights and points into global parameters name of GAUSSIANWEIGHTS32 and GAUSSIANPOINTS32 for speed

Also checked

- 2007 method is x5 slower than 1997 (This is natural)
- current Gaussian 32 point integration for loop is faster than broadcasting in Pn_minus_half_2007

image

JaeBeom1019 (approved):
The logic looks solid and has no calculational error.
No concerns from my side 👍

logan-nc:
Thanks for the great tests @JaeBeom1019. If you are happy with it, then I authorize you to go ahead and merge whenever you are ready.

What do you think of Claude's summary? It looks like it might be worth cleaning up a few minor things it suggested like removing unused variables and confusing annotations before we merge. We don't necessarily have to do everything the ai tells us though - so I leave it up to you as to which ones are worth the time to update

JaeBeom1019:
@logan-nc Thanks a lot! Then since I have few leftover points to check, I'll merge it after that. 😃

JaeBeom1019:
Implementation of Vacuum M.S.Chance 2007 updates @logan-nc 👍 


**Comparison with Previous Methods:**

- 1997 Paper: Used 8-point Gaussian quadrature for all n*rhohat.
- 2007 Paper:
  - Noted that 8-point quadrature causes the associated polynomials to diverge for high n*rhohat > 0.1
  - Proposed using 32-point quadrature for$n*\hat\rho > 0.1$.
  - Suggested a Series Expansion (from complex plane integrals) for $n*\hat\rho > 1.0$, as 32-point quadrature is too slow for very high n.
  - Noted that standard elliptic integrals lose precision with large numbers, requiring iterative methods. Replaced them with Bulirsch's algorithms.

**Benchmarking & Implementation using BenchmarkTools.jl results:**

1. Elliptic Integrals for Bulirsch: This update has been applied to both JPEC and GPEC. It has $~1.0e-7$ difference compared with original one. Tried using Julia's ellipk and ellipe functions (though they are different from what is in VACUUM), but I believe we should use the Bulirsch method due to its balance of speed and accuracy.
2. Series Expansion vs. 32-Point Quadrature: Although the Series Expansion is 1.2x faster than the 32-point method, it introduced errors of ~10%. So I decided not to include the Series Expansion. Instead, we use 32-point quadrature for all $n*\hat\rho > 0.1$.
3. Performance:
    - The 32-point method (2007) is 5 times slower than the 8-point method (1997) for vacuum integration. (In JPEC)
    - However, it is only ~20% slower for a full ideal DCON run. (In GPEC)

Issue #54, opened 2025-10-13
kin_flag = true implementation


jhalpern30:
Implement the flag for including kinetic physics. This might be part of a larger inclusion of a PENTRC conversion, and will be a rather large effort

matt-pharr:
@logan-nc this should be closed whenever this work is merged

logan-nc:
closed with #112

PR #113, opened 2025-12-22
DCON - NEW FEATURE - Adds kinetic damping matrix infrastructure

logan-nc:
Implements the Julia equivalent of Fortran's fourfit_kinetic_matrix (method 0) to enable kinetic damping effects in DCON stability analysis.

Changes:
- Added KineticParams struct with minimal parameter (nl)
- Extended FourFitVars with 14 kinetic matrix spline fields
- Created Kinetic.jl with placeholder compute_tpsi_matrices function
- Implemented make_kinetic_matrix function in Fourfit.jl (~240 lines)
- Integrated workflow in Main.jl to call when kin_flag=true
- Added kinetic branch in sing_der! for future ODE implementation

Key design principles:
- Idiomatic Julia: no global variables, clean parameter passing
- Uses LU factorization for non-Hermitian kinetic A matrix
- Placeholder tpsi returns zeros for infrastructure testing
- Throws informative error at ODE integration (prevents silent bugs)

Future work deferred to separate PRs:
- Real PENTRC interface for compute_tpsi_matrices
- Kinetic ODE equations in sing_der!
- Kinetic metric tensor (fmodb)
- Kinetic singular surface crossings


logan-nc:
@mfairborn23 here is claude code's attempt to implement the conversion of the dcon kinetic damping terms we discussed. I told it to include params it needed in a new KineticParams struct and a new table in the dcon.toml but not to convert over all the pentrc params quite yet because I thought it would be overly distracting. 

The idea is to use a dummy function that returns dummy `kwmats_l` and `ktmats_l`. They are zeros for now, but we should make them "small" in the future (maybe sigma * identity matrix?).

Feel free to work off of this or on your own separate branch! I am just flexing my new pro claude account to see what it can do. If this is no good, let me know so I can continue to improve my ai prompting to be more helpful next time around

logan-nc:
@mfairborn23 if this is good enough to work off of,
(a) Comment here to let people know you are working on this (I'll try not to have claude go crazy with new changes)
(b) Add you notes from our meeting laying out the plan for what to include / exclude for now

mfairborn23:
@logan-nc thanks for this! This looks like a great start and I would love to work on this further. 
I was planning to go through and do the kinetic conversion of the matrices, like the w, k, mll, etc matrices and working on the DCON interface part.
We discussed determining some good artificial "W matrices" for ad hoc kinetic damping as well.
We also discussed allowing the user to specify one number (like the determinant) and scanning that number. If a very small value is chosen, we should get the ideal result back and still some singularity. We would look at plots of delta W vs psi_edge.
Please let me know if there is anything you were hoping for me to discuss that I omitted here by accident.

logan-nc:
@mfairborn23 have you gotten the code to work without using any modified `sing*` code? How did it look?

The joy of the kinetic damping is that it removes the singularities. When a user sets `kin_flag=true` in OMFIT (i.e. include the kinetic matrix modification), it automatically sets `con_flag=true` (i.e. continuous integration). Continuous means it never really needs to ID any singular surfaces or manually calculate derivatives on either side of them. You can just feed the usual ode system into the dynamic ODE solver and let it go from core to edge! 

We did all the singular surface identifications and crossings and everything because we could and that brings things in line with ideal DCON and it is necessary when kinetic effects are really small. Here, `ksing_find` looks for places where detF is super small such that the ODE solutions will blow up with near-singular behavior that might mess up the dynamically stepped integrator. Then `sing_der` allows the code to manually make a finite step over those (near)-singular surfaces and restart the dynamic integrator on the other side. This was good to have for numerical studies where we wanted to show convergence to ideal MHD and when we were debugging and sometimes having super small kinetic effects.... but in practice I found that no real tokamaks needed all that once I got all the terms sorted out and normalized correctly. The kinetic terms basically always provide sufficient damping for the dynamically stepped integrator to get through the near-singular surfaces. Also, we could never find a stable algorithm for identifying exactly where |F|~0 is. I remember the newton iterator really struggled because the value has sharp drops to various "low" values. It was a) hard to ID the exact bottom(s) of any of the sharp wells in |F| and b) ambiguous as to which were "too" low.

image

It might be good to chat with @jhalpern30 about the general structure of how we integrate in JPEC to identify the path of least resistance towards getting some initial results. I think you should basically just need to make the fourfit matrix modifications and then tell the `ODEProblem` to `solve`. One gotcha I can think of off the top of my head is that in JPEC we do the `unorm`ing at pre-specified intervals - but you should be able to at least try using the default ideal MHD intervals as reasonable places to `unorm` but just restart the integrator from exactly where it left off (no manual stepping over the rational surface).

This is not to say that all your hard work on `sing*` things is for nothing! It's great to have in the code. A plot of |F| vs ψ is a must-have diagnostic for understanding how much kinetic damping we have / want, for example. Identifying where |F| is small is great to do. But don't bang your head against the `sing_der` part if it is proving too difficult - progress can still be made!

jhalpern30:
Just adding a link here: this PR will close #54

logan-nc:
This branch was too far out of date with the main development to merge smoothly. Instead, lessons learned were transferred to a fresh start from the modern develop branch. The new PR #210 replaces this one

Issue #107, opened 2025-12-19
Vacuum code needs to be fully vetted for with-wall calculations


logan-nc:
The vacuum code is now fully julia as of #24. That PR includes benchmarks that show agreement in the nowall vacuum energy matrix to within less than 1%. Disagreements when a wall is included (specifically, the conformal wall) are larger however. 

This suggests there is a possible bug in the wall calculations. Some todo items that might help check this are,
1. Create a benchmark script that very carefully compares the plasma and wall geometry (x, z) used internally within the compute_vacuum_response function. These were checked by eye, but a more careful plot showing the relative difference in (x,z) points at each theta index would be helpful. In particular, the points should be compared before and after the sub function that attempts to distribute them on an equal arc grid (and the benchmarks should be sure both the fortran and julia are using this feature).
2. Compare a scan of the conformal wall offset distance set by `a` in the WALL namelist. This should be scanned for values of perhaps [0.1, 0.5, 1, 2, 4, 8]. Note that going much higher will result in the fortran using a nowall calculation. 
3. If these do not find any differences in the actual plasma or wall (x,z) or equal_arc logic then the kernel routines need to be combed through, with the top suspect for a bug being an indexing error. @matt-pharr has already done this though, and should weigh in on whether there is a productive route here

logan-nc:
@claude try to address this issue by going through the suggested checks. If you find a discrepancy in the geometries, try to fix it

matt-pharr:
This is a much more thorough issue than I made :P

Here is the original text of mine, which I will delete:

> The accuracy of the vacuum energy module of the code matches Chance's fortran VACUUM very well for low n, giving usable results for under n=12:

<img width="790" height="524" alt="Image" src="https://github.com/user-attachments/assets/3b8f77d3-f3de-49a3-bef4-59409d7974f9" />

> However, this accuracy is not reproduced in the presence of a conformal ideal wall, which exhibits this strange behavior:

<img width="996" height="664" alt="Image" src="https://github.com/user-attachments/assets/364f152e-59fb-4490-b10d-6d0058f880ac" />

> We need to find any potential remaining bugs in VACUUM and also do benchmarking scans for mthvac.

logan-nc:
Closed with #114

PR #114, opened 2025-12-23
Claude/issue 107 20251219 0412

logan-nc:
Addresses #107 by benchmarking the with wall vacuum energy calculations


logan-nc:
The with wall bug was indeed an indexing issue. The issue was in the equal_arc distribution of the points, which has some old logic assuming mth1 and mth2 type grids. I did some plotting I won't commit that showed the result was a wall that didn't close on itself even if you projected one more point out (we should be using [0,1) theta grids). See this:

image

The way it redistributes, the points look to overlay nicely in the first ~5-10 pts, and you only catch this zooming way in to the final points. You can see the old julia did not close. The new Julia is overlaying the fortran, but missing the final 2 fortran points because it obeys the [0,1) exclusive grid the rest of the vacuum code has adopted. 

So now we can run the vacuum accuracy benchmark script for the Solovev example that has the wall and we get the 1e-3 relative errors between mscvac and all-julia calculations like with no wall. It still goes bad at high n, of course.

image

logan-nc:
It does look like there is still an order of magnitude larger error with wall though...

image

so there must still be something off :(

logan-nc:
Finally caught the remaining bug! I had mixed up the mtheta and mtheta+1 grids in one interpolation call. We now get comparable agreement with nowall and reasonable wall distances.

image

The disagreement at very small conformal wall distances comes from the fact that the fortran enforces 2-point mirroring in a way that ends up forcing the 3 most outboard points to all lie on a vertical line. This then propagates as a bump in the wall distance, which you can see if you set a super small distance and zoom in a ton:
image

I think our wall looks more "correct" in that it actually conforms a constant distance from the plasma. 
Our arc lengths between points are also more exactly constant (because we do not interpolate the theta needed for equal arc lengths and then use that to interpolate x,z vs theta). This isn't a big deal, but I found it was more robust when I went to extreme cases.

image

PR #116, opened 2025-12-26
Syntax organizing and notebook cleanup 

### LINKED ISSUE #43: Chease Binary Reader Issue
Image

Although the chease read function, which reads in chease binary files, is implemented in JPEC, we discovered discrepancies between the outputs of both the original GPEC (which matches the JPEC) EQUIL results for Chease binary files and identical Chease ascii files. The GPEC source code for reading in the Chease binary files may have a bug in it. See the attached images for an example of the binary versus the ascii equilibrium from the same Chease run.

Image

### LINKED ISSUE #109: Code organization and cleanup
We already have a good directory structure which keeps all of our code and supporting functionality well-organized. However, someone needs to go through and:

- [ ] Consolidate all the miscellaneous benchmarking scripts
- [ ] Make sure all of these scripts run through properly
- [ ] Further flush out the examples
- [ ] Standardize unit tests
- [ ] Unit tests should be reworked to be quantitative instead of just structural

We also need to implement the vision we have for unit tests and code versioning, though that deserves its own separate issue.

### LINKED ISSUE #110: InverseEquilibrium.jl not working
I’ve been looking into InverseEquilibrium.jl on the develop branch, and it seems like there have been some significant updates recently.

- Undefined Variables: The code calls `sq_xs, new_rz_in,` and `sq_fs,` which are never initialized or are misspelled, causing immediate crashes. And some spelling errors like `rz_` exists.
- Disabled Logic: Critical transformation steps (like integrals or coordinate mapping) are changed into annotations, leaving the core algorithm non-functional.

Is anyone currently tasked with these updates? If this part is still a work in progress, would anyone  know a stable commit or a specific version where InverseEquilibrium.jl is fully operational?

JaeBeom1019:
### Summary of Changes

This PR focuses on code organization, bug fixes across modules, and ensuring all example notebooks run correctly.

Although this branch was originally created for code cleanup, I identified and fixed errors (syntax, zero division, outdated data structures, ...) that are critical for other ongoing developments. Therefore, I would like to merge this PR first.

#### EQUIL (Equilibrium)
* **Fixes:**
  * Restored `InverseEquilibrium` functionality.
  * Patched `lar_run` function: set `z0 = 1.e-4` to avoid zero division errors.
  * Fixed spelling errors and updated old data structure annotations.
* **Dependencies:** Added `DifferentialEquations.jl` to use `AutoFiniteDiff`.
* **Cleanup:** Removed `equil_test_bu.ipynb` (no longer used).
* **Testing & WIP:**
  * Added `INP1_formatted` and `INP1_unformatted` for testing.
  * **WIP:** Ongoing benchmarking for EFIT & CHEASE output matching (Flux currently matches; differences in R remain) with @jmlmir369 .

#### VACUUM
* **Refactoring:**
  * Renamed module `VacuumMod` -> `Vacuum` in DCON.
  * Updated `WallShapeSettings` and `WallGeometry` to be **mutable structs**.
  * Refactored `vac_wall.ipynb`.
* **Cleanup:** Removed obsolete notebooks (`vac_arrays.ipynb`, `vac_vaccal.ipynb`).

#### SPLINE
* **Updates:** Replaced all `spline_setup` and `bicubic_setup` calls with the updated `CubicSpline` implementation.
* **Fixes:** Fixed all `spline_*` examples; verified they run without errors.

#### General / Maintenance
* **Notebooks:** Verified and fixed all `.ipynb` files in notebook directory to ensure they execute correctly.


JaeBeom1019:
@logan-nc Could you please take a look? It mainly addresses existing errors without adding complex new features, so I expect it to be a quick review.

src/Equilibrium/AnalyticEquilibrium.jl
```diff
@@ -1,3 +1,6 @@
+
+using DifferentialEquations
```
logan-nc: It seems this is only used for the Jacobian of the ODE solver. It is not clear to me that a stiff solver is really necessary here, but if it is then I think the faster `autodiff=true` setting should be fine. I suspect someone's AI conversion just suggested this setting without any real reason behind it. Maybe I am wrong, but to warrant adding a new dependency, I would need a separate PR with plots showing why exactly the added accuracy of `autodiff=AutFiniteDiff()` is needed.

Project.toml
```diff
@@ -6,6 +6,7 @@ version = "0.1.0"
 
 [deps]
 DiffEqCallbacks = "459566f4-90b8-5000-8ac3-15dfb0a30def"
+DifferentialEquations = "0c46a032-eb83-5123-abaf-570d42b7fbaa"
```
logan-nc: See comment in lar file where this is used.

Project.toml
```diff
@@ -23,6 +24,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
 DiffEqCallbacks = "4.9.0"
+DifferentialEquations = "7.17.0"
```
logan-nc: See comment in lar file where this is used.

src/Equilibrium/InverseEquilibrium.jl
```diff
@@ -60,9 +58,11 @@ function equilibrium_solver(input::InverseRunInput)
     newq0 = config.control.newq0
 
     me = 3
-    interp = false
-    diagnose_rz_in = false
-    diagnose_rzphi = false
+
+    # Not using anymore
```
logan-nc: Fully remove these as they are not used anywhere else. We do not want to leave cruft like this lying around from al the active development in our early days with the code. It will soon become bloated with cruft, confusing new developers.

src/Equilibrium/ReadEquilibrium.jl
```diff
@@ -120,14 +120,14 @@ function read_efit(config::EquilibriumConfig)
 end
 
 """
-    _read_chease2(equil_in)
+    read_chease2(config)
```
logan-nc: Can we rename this to something more informative than "2"? What does the "2" mean? It means ascii instead of binary, right? Follow the fortran example, which has already renamed this `read_eq_chease_inp1_ascii`.

src/Vacuum/VacuumStructs.jl
```diff
@@ -106,7 +106,7 @@ Struct containing input settings for vacuum wall geometry.
   - `"elliptical"`: Elliptical wall
   - `"dee"`: Dee-shaped wall
   - `"mod_dee"`: Modified Dee-shaped wall
-  - `"filepath"`: Custom wall shape from the file you specify
+  - `"your_file_path"`: Custom wall shape from the file you specify
```
logan-nc: How about `"<path-to-file>"`?

src/Vacuum/VacuumStructs.jl
```diff
@@ -84,7 +84,7 @@ Arrays are of length `mtheta`, where `mtheta` is the number of poloidal grid poi
 - `dx_dtheta::Vector{Float64}`: Derivative dR/dθ at wall
 - `dz_dtheta::Vector{Float64}`: Derivative dZ/dθ at wall
 """
-@kwdef struct WallGeometry
+@kwdef mutable struct WallGeometry
```
logan-nc: We purposefully made these immutable in the Vackathon. Please keep them so. Our reason is that is make the code more clear since the developers know the wall geometry is not being changed as it is passed through the various functions. 

If there is a reason you want it to be mutable and want to change it after its initialization, this is a big enough change that it should have its own PR with a clear reasoning explained so @jhalpern30 and I can understand the need. So please revert these mutable changes in this branch

src/Vacuum/VacuumStructs.jl
```diff
@@ -116,7 +116,7 @@ Struct containing input settings for vacuum wall geometry.
 - `equal_arc_wall::Bool`: Flag to enforce equal arc length distribution of nodes on the wall 
   (recommended unless wall is very close to plasma)
 """
-@kwdef struct WallShapeSettings 
+@kwdef mutable struct WallShapeSettings
```
logan-nc: See earlier comment about these

logan-nc (changes requested):
Thanks for all this @JaeBeom1019!

I am trusting your changes to the equilibrium code for the most part, since I have only been using the solovev or efit options and have not fully followed these other ones. It seems like you've tested them well.

I've requested some changes/reversions in the edits where I think too large of changes were bundled together in this PR though. Adding a dependency, for example, warrants a separate discussion (otherwise, I fear we will soon find ourselves swamped by dependencies that we don't really need). We also purposefully made the geometry structs immutable so that would need a separate discussion as well.

Please respond to the above requests (should all be easy since they are mostly undoing things) and let me know when it is ready to review again.

logan-nc:
oh, and can you also please say what equilibrium is in `notebooks/CHEASE_INP1/EQDSK`? We need to make sure it is not an equilibrium for any device that has user agreements for data access (DIII-D, KSTAR, etc.) since this is an open source code and it will be accessible by anyone

src/Equilibrium/AnalyticEquilibrium.jl
JaeBeom1019: autodiff=true needs another package so I fixed into autodiff=false. I also think that AI did `autodiff=AutoFiniteDiff()`

Project.toml
JaeBeom1019: removed DifferentialEquations.jl dependency! 👍

JaeBeom1019:
@logan-nc I'll revert those changes and clean up the dependencies shortly. It shouldn't take long. Also all files in CHEASE_INP1 directory are from chease examples. If you worry about any data issue, I can replace it with virtual equilibrium like `R0=1m, B0=1T`

logan-nc:
Is the chease code open source? If yes, then using their "official" examples would be good. If no, then it could still be problematic and we should just use our own simple examples as you suggest

src/Equilibrium/ReadEquilibrium.jl
JaeBeom1019: Changed function `read_chease` and `read_chease2` into `read_chease_binary` and `read_chease_ascii`. Also not only `chease` and `chease2` naming, `chease_ascii` and `chease_binary` conventions are now supported to avoid the confusion of checking which one is binary or ASCII. 

image

src/Vacuum/VacuumStructs.jl
JaeBeom1019: Added mutable just for convenience for me. Since you have certain reason, I'll follow your immutable convention.

JaeBeom1019 (commented):
Fully changed what @logan-nc mentioned

- Added new INP1_ascii, INP1_binary, GEQDSK_COCOS_02 for simple equilibrium (B0 = 7.4, R0 = 6.8). 
- Refactorized reading inverse equilibrium
- Removed unused lines or variables and DifferentialEquations.jl dependency
- Replaced function names. read_chease/read_chease2 into read_chease_binary/read_chease_ascii
- Changed Vacuum Structures into immutable

JaeBeom1019:
@logan-nc  Also added unit tests for direct/inverse run

logan-nc (approved):

PR #40, opened 2025-08-06
Code formatting

jhalpern30:
This pull request is related to [this issue](https://github.com/OpenFUSIONToolkit/JPEC/issues/36) and adds a JuliaFormatter.toml file (identical to that in FUSE) which allows for standardization of code formatting across PRs. 

I updated the README to include a section on code formatting (although this can eventually get moved to a developers section in the documentation if desired). I also ran the formatter on the existing files in the main branch to showcase some of the fixes.

PR #120, opened 2026-01-01
Add pre-commit

logan-nc:
This branch adds a pre-commit yml file that cleans up the notebook meta data and outputs so we stop recording silly changes whenever anyone runs a notebook. 

I made it a pre-commit so developers have to opt in by installing the pre-commit functionality and initializing it for their local repo. The alternative is to make it a hook, which would then be forced on anyone making any commit... I am open to thoughts on wether we should switch to this strict approach.

I also added a Julia Formatter. I think it would be valuable to enforce a standard of formatting, which makes the code more coherent and readable. I don't want to be overly prescriptive, but some consistent basics are good imo. I am also open to thoughts on this: should we add more prescriptions? Make it less prescriptive? Not use it?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,40 @@
+# Pre-commit hooks for JPEC repository
+# Install: pip install pre-commit
+# Setup: pre-commit install
+# Run manually: pre-commit run --all-files
+
+default_language_version:
+  python: python3.10
+
+repos:
+  # Jupyter notebook cleaning - removes outputs, execution counts, metadata
+  - repo: https://github.com/kynan/nbstripout
+    rev: 0.7.1
+    hooks:
+      - id: nbstripout
+        name: Strip notebook outputs
+        args:
+          - --extra-keys
+          - |
+            metadata.kernelspec metadata.language_info.version
+            cell.metadata.executionInfo cell.metadata.colab
+            cell.metadata.id cell.metadata.scrolled
+
+  # General file cleanup
+  - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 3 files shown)


jhalpern30:
> I made it a pre-commit so developers have to opt in by installing the pre-commit functionality and initializing it for their local repo. The alternative is to make it a hook, which would then be forced on anyone making any commit... I am open to thoughts on wether we should switch to this strict approach.

I am ok with it being a pre-commit. If someone commits a notebook with outputs still in it, it will be rather obvious during PR review and we can just mention to install the pre-commit package and run from there. 

> I also added a Julia Formatter. I think it would be valuable to enforce a standard of formatting, which makes the code more coherent and readable. I don't want to be overly prescriptive, but some consistent basics are good imo. I am also open to thoughts on this: should we add more prescriptions? Make it less prescriptive? Not use it?

I've been running it here and there on the repo, and I think we definitely need some sort of formatting, especially when it comes to whitespace removal, standardizing spacing in function calls, etc.. I've tried to remember when committing PRs but was not as diligent as I wanted and ended up committing changes that haven't been formatted and then not wanting to do an entirely separate PR to just do formatting, so I think auto-running this is a great idea. I think the requirement should just be to run the Julia formatter to catch bad formatting and not directly prescribe anything major that developers need to consider. This JuliaFormatter.toml file was directly copied from the FUSE repo in #40, and I think it works pretty well to clean up the code

jhalpern30 (approved):

logan-nc:
@matt-pharr @JaeBeom1019 @mfairborn23 @satelite2517 please consider initializing these precommits as per the instructions. it will be helpful if we all confirm to the same formatting

Issue #70, opened 2025-10-21
Switch to a standardized spline package


logan-nc:
Summarizing discussion in #68:

@logan-nc suggested we spend out time using an existing Julia spline package rather than converting the custom splines from fortran to Julia and doing lots of modifications to make them more readible and thread stable.

@bclyons12 listed the most well-known ones to consider:
- [Interpolations.jl](https://juliamath.github.io/Interpolations.jl/)
- [Dierckx.jl](https://github.com/JuliaMath/Dierckx.jl)
- [DataInterpolations.jl](https://docs.sciml.ai/DataInterpolations/stable/)

@logan-nc identified a winner:
Gemini and ChatGPT both recommend Interpolations.jl.
Gemini says
> Yes, you can use Interpolations.jl in parallelized code, such as with Julia's built-in multithreading and distributed computing capabilities. An Interpolations.jl object is read-only after construction, making it safe to use across multiple threads for interpolation lookups. 

ChatGPT warns agains Direckx.jl 
>Avoid Dierckx.jl for threaded evaluation of derivatives. There’s an open issue showing wildly incorrect results when calling Dierckx.derivative from multiple threads; single-thread is fine, but multithreaded derivatives are not thread-safe in practice. 

while recommending Interpolations.jl as
> Pure Julia; very fast construction & throughput on both B-spline and gridded methods; handles multi-D regular and irregularly spaced axes; well-documented boundary conditions/extrapolation; works with modern Julia versions. The docs include benchmark “shootouts” vs Dierckx and others where Interpolations wins across cases. Maintenance stance is “maintained, semver, fixes prioritized,” which signals low surprise from version to version.

The main complaint for Interpolations.jl seems to be the assumption of regular grids, but there is an option to scale the grid that I think would work for us.... except that I don't think we even need it! 
- Our irregular equilibrium psi grids are all just known functional packings of the points so we can spline in regular variable `x` where `x=psin^2` and then easily just eval like`qspl(psin^2)` or even add explicit `x=psin^2; qspl(x)` type stuff in all the functions that happen at a psin value
- GPEC splines the solutions on the dynamically determined force-free solution grid (`u` grid) but only ever does calculations at nodes - so that is just wasted of overhead really.
- Theta grids are uniform. 

## Conclusion
Someone needs to spend a day attempting to place Interpolations.jl splines in for the old fortran ones we currently have wrappers for.
[...]

PR #123, opened 2026-01-03
WIP - Spline Migration

logan-nc:
This branch is for migrating the fortran spline wrappers to pure julia splines

This addresses #70


logan-nc:
This is a work in progress. I just pushed and made the PR to flag for others that I have claude working on it... if anyone else is interested in helping just say so here and we can partition the work a bit

logan-nc:
FYI, the to do list I have claude working on now is below

STAGE 3: Complex Cubic Splines

  Status: Not yet started

  Replace CubicSpline{ComplexF64} with Interpolations.jl for complex-valued splines.

  Currently used in DCON module:
  - amats, bmats, cmats, dmats, emats, hmats - stability matrix splines
  - fmats_lower, kmats, gmats - additional matrix splines
  - wvmat_spline - vacuum matrix spline

  These are all in src/DCON/DconStructs.jl and used for the fixed/free boundary stability calculations.

  STAGE 4: FourierSpline Evaluation

  Status: Not yet started

  Evaluate whether to replace FourierSpline (FFT-based interpolation).

STAGE 5: Benchmark at test

logan-nc:
summary of benchmarking julia vs fortran splines:

Benchmark Results: Fortran vs Julia Spline Implementations

  Performance Comparison

  | Metric       | Fortran Splines | Julia (Interpolations.jl + FFTW) | Improvement |
  |--------------|-----------------|----------------------------------|-------------|
  | Mean Time    | 229.1 s         | 3.9 s                            | 59x faster  |
  | Min Time     | 222.1 s         | 3.6 s                            | 62x faster  |
  | Memory Alloc | 139.7 MB        | 212.9 MB                         | 1.5x more   |

  Eigenvalue Comparison

  | Eigenvalue     | Fortran | Julia | Rel. Diff |
  |----------------|---------|-------|-----------|
  | Lowest (et[1]) | 15.44   | 10.29 | 33%       |
  | et[2]          | 144.2   | 125.6 | 13%       |
  | et[3]          | 405.5   | 327.0 | 19%       |

  Root Cause of Eigenvalue Difference

  The investigation revealed that the Fortran version had incorrect spline derivatives:

  - P (pressure) derivative in Fortran: Constant at -5.37e-02 everywhere (incorrect)
  - P derivative in Julia: Varies correctly from -0.107 to ~0 across the profile

  The Fortran spline wrapper was returning a constant average slope instead of proper pointwise derivatives. The Julia implementation using ForwardDiff computes correct derivatives.

  Conclusion

  1. Speed: The pure Julia implementation is ~60x faster than the Fortran wrapper version
  2. Memory: Uses ~50% more memory (acceptable tradeoff for the speed gain)
  3. Correctness: The Julia version produces more accurate derivatives than the old Fortran wrappers
  4. Eigenvalue difference: The 33% difference in the lowest eigenvalue is due to the Fortran version computing incorrect pressure derivatives, not a bug in the Julia migration

  The migration to pure Julia splines is a significant improvement in both performance and numerical accuracy.

jhalpern30:
This is super cool! I was skimming through it and two quick comments (that might get outdated before this is ready for actual review)
1) We'll need to make sure to replace `::Any = nothing` in the structs with their actual type/instantiation.
2) Is there a reason you're using the ForwardDiff package instead of the built-in Interpolations.gradient and Hessian?

logan-nc:
This effort has been taken over be #137

PR #124, opened 2026-01-03
Julia speedups

logan-nc:
@claude summarize this branch for the PR


logan-nc:
@claude summarize this branch for the PR

logan-nc:
@jhalpern30 this 2x speedup seems promising. The ai also seems to have hinted that parallelizing chunked integration would be within its abilities (i.e. basically stride??) but I decided to leave that large of a change until after #121 is merged into develop.

I personally think the change to `OdeState` makes it much more difficult to read as a non-julia expert. The AI really liked it, but it can't be that big of a factor, can it? What do you think? I'd be fine reverting that or keeping it based on your preference.

logan-nc:
@jhalpern30 this is ready for review and I promise not to add anything else! I couldn't help reprofiling and adding the last two minor improvements. But let's get this merged and then your big refractor of the integration bounds in before we dig any deeper

src/DCON/DconStructs.jl
```diff
@@ -225,6 +225,7 @@ A mutable struct containing control parameters for stability analysis, set by th
     write_outputs_to_HDF5::Bool = true
     HDF5_filename::String = "euler.h5"
     force_wv_symmetry::Bool = true
+    save_interval::Int = 10  # Save every Nth ODE step (1 = save all, 10 = save every 10th). Always saves near rational surfaces.
```
jhalpern30: This is effectively `euler_stride` from the Fortran, yeah? I like this name better since its more clear, but for consistency should we have somewhere in the source/documentation that makes it clear this performs the same function? Also, why default to 10? In the Fortran, euler_stride was defaulted to 1. I'm ok leaving it at 10, but just wanted to double check we want this as a default

src/DCON/DconStructs.jl
```diff
@@ -347,64 +348,119 @@ and a small set of temporary matrices and factors used to compute singular-layer
       + `tmp::Matrix{ComplexF64}` - Workspace matrix for EL derivative calculations with shape `(numpert_total, numpert_total)`.
       + `Afact::Union{Cholesky{ComplexF64, Matrix{ComplexF64}}, Nothing}` - Cholesky factor
       + `singfac_vec::Vector{Float64}` - Vector of m-nq factors
+      + `chol_workspace::Matrix{ComplexF64}` - Pre-allocated workspace for in-place Cholesky
 """
-@kwdef mutable struct OdeState
+mutable struct OdeState
```
jhalpern30: Agreed that this looks a lot messier and I prefer using the @kwdef, which was a recommendation by Brendan in the first place. @bclyons12, any input here? Claude seems to think that using the @kwdef creates overhead that is removed when we use the explicit constructor here (which I'm not convinced it does, since from what I can tell @kwdef is really just a helper macro to do the same thing as the explicit constructor but more clearly). 

We could also just benchmark the code with and without the explicit constructor change here to see if it gives any speedup

src/DCON/Fourfit.jl
```diff
@@ -206,6 +206,10 @@ function make_matrix(equil::Equilibrium.PlasmaEquilibrium, intr::DconInternal, m
     imat = zeros(ComplexF64, 2 * intr.mband + 1)
     imat[mid] = 1 + 0im
 
+    # Pre-allocate workspace for matrix solve operations to avoid allocations in loop
+    temp1 = Matrix{ComplexF64}(undef, intr.numpert_total, intr.numpert_total)
```
jhalpern30: Thoughts on calling these a_inv_dmat and a_inv_cmat or something similar to provide even more clarity on that the temporary matrices are? Unless its used for multiple purposes, I like the practice of naming even temp matrices something more descriptive in case we need to use them/print outputs

src/DCON/Fourfit.jl
```diff
@@ -302,8 +306,9 @@ function make_matrix(equil::Equilibrium.PlasmaEquilibrium, intr::DconInternal, m
         # TODO: Fortran threw an error if factorization fails for A/F due to small matrix bandwidth,
         # Add this check back in if we implement banded matrices
         amat_fact = cholesky(Hermitian(amat, :L))
-        temp1 = amat_fact \ dmat
-        temp2 = amat_fact \ cmat
+        # Use in-place linear solve to avoid allocating temporaries
```
jhalpern30: AI always did this for me too when adding updates - I don't think we need to add a comment that we are doing an in-place solve/multiplication/copy every time we do, it clutters things up a bit. There's a few of those in this PR that can be removed

src/DCON/EulerLagrange.jl
```diff
@@ -329,17 +331,38 @@ function integrator_callback!(integrator)
...
+    # 1. First few steps of integration (ensures we capture point right after rational/axis)
+    # 2. Every Nth step (save_interval)
+    # 3. Near the end of integration segment (ensures we capture point right before next rational)
+
+    # Check if we're near the end of this integration segment
+    psi_range = abs(integrator.sol.prob.tspan[2] - integrator.sol.prob.tspan[1])
+    psi_remaining = abs(integrator.sol.prob.tspan[2] - integrator.t)
+    near_end = psi_remaining < 0.05 * psi_range || psi_remaining < 1e-4
+
+    # Check if we're at the beginning (first 2 steps capture the point right after rational)
```
jhalpern30, line 401: Overall, I like this change, and the comments make this logic clear. Its not clear to me why we need the first 2 steps to capture the point right after the rational, can you add something extra here? Is it for a derivative?

src/DCON/EulerLagrange.jl
```diff
@@ -547,23 +570,29 @@ function transform_u!(odet::OdeState, intr::DconInternal)
...
             for jsol in 1:intr.numpert_total
                 if mask[jsol]
                     temp[ksol, jsol] = odet.fixfac[ksol, jsol, ifix]
                 end
             end
-            # Matrix multiplication gauss = gauss * temp
-            gauss[:, :, ifix] .= gauss[:, :, ifix] * temp
+            # In-place matrix multiplication: gauss = gauss * temp
+            mul!(gauss_buffer, view(gauss, :, :, ifix), temp)
+            gauss[:, :, ifix] .= gauss_buffer
```
jhalpern30, line 649: I think this can be made `@views copyto!(gauss[:, :, ifix], gauss_buffer)` just like the modifications you made to `Free.jl`

jhalpern30: Actually, a lot of the broadcasting operations below can use `copyto!` as well. For the sake of consistency, I would make sure the notation in this function and `Free.jl`/`Ode.jl` are identical. 
Based on some googling, it seems that the `.=` notation is effectively the same as `copyto!` in terms of runtime performance, and it really becomes user preference for clarity. I don't want to add too much work here so I see two paths:
minimal change: make the notation consistent in these PR changes, either always using broadcasting `.=` or the `copyto!`
more work: do a short benchmark in terms of allocations/runtime with all `.=` operations and `copyto!` operations. Replace all with whichever is best, or if identical, whichever you determine to be more readable

src/DCON/Sing.jl
```diff
@@ -728,7 +728,9 @@ function sing_der!(du::Array{ComplexF64,3}, u::Array{ComplexF64,3},
         kmat = reshape(odet.kmat, intr.numpert_total, intr.numpert_total)
         gmat = reshape(odet.gmat, intr.numpert_total, intr.numpert_total)
 
-        odet.Afact = cholesky(Hermitian(amat))
+        # Use in-place Cholesky with pre-allocated workspace to avoid allocation
+        copyto!(odet.chol_workspace, amat)
+        odet.Afact = cholesky!(Hermitian(odet.chol_workspace))
```
jhalpern30: I might be wrong here so this needs double checking, but I am pretty sure since we don't need `amat` anywhere later in this function, we can get rid of the chol_workspace and overwrite amat instead by just using amat as the workspace for the Cholesky factorization, i.e.
`odet.Afact = cholesky!(Hermitian(amat))`
This is more readable, removes the extra temp variable, and removes an additional copy operation from amat to chol_workspace.
I ran a quick test by just commenting out the chol_workspace copyto line and using the cholesky! line above, and the outputs are at least the same (as expected since amat is not reused). The allocations and runtime were near identical

jhalpern30 (changes requested):
Looks great! Just a few requested cleanups, mostly pertaining to optimizations that were added that I don't think add any actual optimization and just reduce clarity, and it should be ready to go

src/DCON/Sing.jl
jhalpern30: Confirmed you can just reuse the existing `amat` workspace. I'll add this in my upcoming changes

src/DCON/DconStructs.jl
jhalpern30: Confirmed that this adds no improvement for either time or allocations. Will remove in my upcoming changes

src/DCON/EulerLagrange.jl
jhalpern30: Verdict here - tossing `copyto!` in favor of `.=` since they're identical. Also discovered you can just wrap a for loop in `@views` to avoid having to write it every time, i.e.
```julia
@views for istep in jfix:kfix
    # This is u1->u4 in Fortran
    mul!(gauss_buffer, odet.u_store[:, :, 1, istep], transforms[:, :, ifix])
    odet.u_store[:, :, 1, istep] .= gauss_buffer
    mul!(gauss_buffer, odet.u_store[:, :, 2, istep], transforms[:, :, ifix])
    odet.u_store[:, :, 2, istep] .= gauss_buffer
    mul!(gauss_buffer, odet.ud_store[:, :, 1, istep], transforms[:, :, ifix])
    odet.ud_store[:, :, 1, istep] .= gauss_buffer
    mul!(gauss_buffer, odet.ud_store[:, :, 2, istep], transforms[:, :, ifix])
    odet.ud_store[:, :, 2, istep] .= gauss_buffer
end
```

jhalpern30:
Summary of my changes:
I removed the unnecessary fluff from this PR that wasn't contributing to speedups and was just mystifying the code. This included some of the copyto! calls and OdeState constructor. For reference, these were my @btime results:
Starting point from this PR
  5.947 s (2932516 allocations: 4.69 GiB)
Without chol_workspace
  5.972 s (2944474 allocations: 4.68 GiB) (VERDICT: remove)
With @kwdef Odestate constructor
  5.665 s (2944476 allocations: 4.68 GiB) (VERDICT: remove)
Removing optimized coeffs calcs in Free.jl
  6.079 s (2949564 allocations: 4.78 GiB) (VERDICT: minimal, but not negligible, so keep this)
Removing Gauss buffer in transform_u 
  6.737 s (2956411 allocations: 4.92 GiB) (VERDICT: minimal, but not negligible, so keep this, but replacing `copyto! with `.=` maintains sees no change)

src/DCON/EulerLagrange.jl
jhalpern30: @logan-nc almost all of the prep work on this branch is done now, I just need your input on this last thing here. When you get a chance, give this comment a quick look, and also check out issue 1. from Claude below and just confirm that that's not an issue (I don't think it is, but is worth a look). 

After this, its ready to get merged in

PR #135, opened 2026-01-09
Replacing delta with nu and other VacuumInputs cleaning

### LINKED ISSUE #134: Replace delta inputs with nu in vacuum interface
Currently, the vacuum code accepts the plasma limiter geometry in the form of `r, z, delta` from DCON, where `delta = -nu /qa` is defined in eq. 98 of the Chance paper and appears to be somewhat vestigial from the PEST code. However, everywhere in the vacuum code we just multiply it by `qa` and so we really only need `nu`. It would improve clarity/understanding of the code to replace the `delta` input with `nu = f[3]` in `set_vacuum_inputs`, and then form the Fourier functions used in the transforms as `m * theta - n * nu`, directly matching equations 115-118 of Chance's paper. As part of fixing this issue, we can also clean up the `initialize_plasma_surface` function by better documenting the above and also removing unnecessary functions from the `PlasmaGeometry` struct `cnqd, snqd, sinlt, coslt` and better naming `snlth, cslth`

jhalpern30:
Closes #134

### DIFF
--- a/src/DCON/Free.jl
+++ b/src/DCON/Free.jl
@@ -8,6 +8,7 @@ and returns a `VacuumData` struct containing the data needed for perturbed equil
 and data dumping.
 
 ### TODOs
+
 Check if normalize is ever false, currently always true, and if not, remove related logic
 """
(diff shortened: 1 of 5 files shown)


jhalpern30:
@claude please review and summarize this PR

logan-nc (approved):
Nice cleanup ✅

PR #137, opened 2026-01-14
Spline Migration using FastInterpolations

logan-nc:


logan-nc:
@claude summarize this branch's PR for me

logan-nc:
Here are the benchmark comparison results (DIIID-like example):

  Performance Comparison
```   
  ┌──────────────┬────────────────┬────────────────┬────────┐
  │    Metric    │ Develop Branch │ Feature Branch │ Change │
  ├──────────────┼────────────────┼────────────────┼────────┤
  │ Mean runtime │ 6.93s          │ 5.41s          │ -22%   │
  ├──────────────┼────────────────┼────────────────┼────────┤
  │ Min runtime  │ 5.42s          │ 4.38s          │ -19%   │
  ├──────────────┼────────────────┼────────────────┼────────┤
  │ Eigenvalue   │ 1.701          │ 1.709          │ +0.5%  │
  └──────────────┴────────────────┴────────────────┴────────┘
```
  Summary

  The pure Julia BicubicSpline implementation is ~20% faster than the Fortran-based splines on the develop branch.

  Note: There's a small eigenvalue difference (1.701 vs 1.709, ~0.5%). This is likely due to subtle numerical differences between the Hermite bicubic interpolation in the pure Julia implementation vs the Fortran spline library. Both values indicate stable modes (positive total energy).

logan-nc:
10 points for anyone who figures out where the remaining stability difference comes from. my suspicion is the fourier splines, since we use FFTW and I know DCON uses an fft_flag that toggles different behaviors.

I really thought it was the bicubic psi_in for forming the equilibrium surfaces but updating that to follow the Fortran didn't change things at all.

src/Splines/FastCubicSpline.jl
```diff
@@ -0,0 +1,326 @@
...
+| evaluate (1 pt) |      5.1 ns   |        9.3 ns     | Single point             |
+| deriv1 (1 pt)   |      5.2 ns   |        9.3 ns     | Single point             |
+| deriv2 (1 pt)   |      4.9 ns   |        8.1 ns     | Single point             |
+| deriv3 (1 pt)   |      4.6 ns   |        6.2 ns     | Single point             |
+| Monotonic loop  |      4.7 ns/pt|        8.9 ns/pt  | Best case for cached     |
+| Random loop     |     14.5 ns/pt|        8.8 ns/pt  | FastCubicSpline1D wins   |
+
+**Summary**: CubicSpline1D with cached interval search is ~2x faster for monotonic
+access patterns (typical in ODE integration). FastCubicSpline1D is faster for random
+access but lacks multi-quantity support and cached interval optimization.
```
logan-nc: @mgyoo86 here is the thin wrapper attempt and summary of performance. I ended up using the heavier SplineAdaptor.jl implementation because it caches the last index, speeding the Monotonic loops.

mgyoo86:
Hello @logan-nc,

I found a bug in the `CubicSpline1D` implementation and committed a quick fix. This may be one of the reasons why this PR yields different results.

## Bug Fix: `extrap` Boundary Condition

Fixed a bug in `CubicSpline1D` related to extrapolation boundary conditions.

### Summary

| BC Type | Status |
|---------|--------|
| **Natural BC** | Already matched `FastInterpolations` ✅ |
| **Extrap BC** (with provided slope) | Had incorrect results due to a bug — now fixed ✅ |

### Comparison on $f(x) = x^3$

Cubic spline interpolation should reproduce $x^3$ exactly when correct boundary derivatives are provided. The `_estimate_endpoint_derivative_series` function uses Lagrange interpolation to estimate endpoint derivatives, which gives the exact first derivative ($3x^2$) for polynomial functions up to degree 3.

Therefore, with properly computed boundary conditions, `CubicSpline1D` must match the analytical result for this case.

### **Before fix:**
before_fix

### **After fix:**
after_fix


---
### Note on Periodic BC

The current periodic BC implementation is not a true periodic spline and has some incorrect behavior. Unlike the extrap bug above, fixing this properly is non-trivial, so it's not addressed in this PR.

For now, I think we can continue using the current implementation and plan to address this when integrating `FastInterpolations.jl` more thoroughly:
- **1D cubic** is already available with better performance and accuracy than `CubicSpline1D`
- **2D bicubic** is not yet supported, but will be developed in the coming weeks

Once bicubic support is ready, we can cleanly integrate both 1D and 2D splines at once, which would be a better path than patching the current implementation piece by piece.

logan-nc:
Thanks for the great benchmark and deep bug fix @mgyoo86!

The DIIID-like example now benchmarks as follows:
```
  ┌───────────────────┬─────────┬──────────────────┬──────────────────┐                                                   
  │      Metric       │ develop │ spline_migration │        et[0]     │                                                   
  ├───────────────────┼─────────┼──────────────────┼──────────────────┤                                                   
  │ Eigenmode energy  │ 1.7005  │ 1.7058           │ +0.31%           │                                                   
  ├───────────────────┼─────────┼──────────────────┼──────────────────┤                                                   
  │ Integration steps │ 911     │ 861              │ -5.5% (50 fewer) │                                                   
  ├───────────────────┼─────────┼──────────────────┼──────────────────┤                                                   
  │ Runtime (warmed)  │ 6.62s   │ 3.94s            │ -40.5% faster    │                                                   
  └───────────────────┴─────────┴──────────────────┴──────────────────┘   
```

So we are indeed a bit closer to the fortran spline result. Still not quite *exactly* the same but perhaps the remaining small discrepancies are from the 2D splines.

logan-nc:
FYI for anyone following, I've had claude completely convert the code to directly use FastInterpolations everywhere instead of using wrappers that replicate the old fortran api.

Things were running really slow for some reason, but after restarting my computer and instructing claude to do proper JIT warmups, here is the speed comparison:

```
Comparison: Feature vs Develop                                                                                          
  ┌─────────┬───────────────────┬───────────────────┬────────┐                                                            
  │ Metric  │ Feature (d816d95) │ Develop (bb65a5b) │ Change │                                                            
  ├─────────┼───────────────────┼───────────────────┼────────┤                                                            
  │ et[1]   │ 1.7072            │ 1.7005            │ +0.4%  │                                                            
  ├─────────┼───────────────────┼───────────────────┼────────┤                                                            
  │ Steps   │ 864               │ 911               │ -5.2%  │                                                            
  ├─────────┼───────────────────┼───────────────────┼────────┤                                                            
  │ Runtime │ 2.12s             │ 2.73s             │ -22%   │                                                            
  └─────────┴───────────────────┴───────────────────┴────────┘  
```

logan-nc:
@claude give this another review. Look carefully for any instances where indexing might have gotten messed up. The periodic splines, for example, must all be made on closed grids now (where the last x,y match the first x,y)

logan-nc:
@claude how is it now?

mgyoo86:
@logan-nc 
I’m happy to see `FastInterpolations.jl` getting more tightly integrated into `JPEC` 🙂

I’ve been extending FastInterpolations.jl into a general N-dimensional interpolation package. The core ND cubic interpolation pieces (including bicubic) are mostly done, and I’m now polishing/refining the APIs, including JPEC-inspired APIs :).

I haven’t had a chance to go through all the details of this PR yet, but the direction looks good to me as-is.

Once I finalize and release the ND CubicInterpolation, I’ll take a closer look at this PR and see whether the new APIs can help make the adapter cleaner and more efficient 🙂

logan-nc:
Thanks @mgyoo86 ! 
Yeah, I have it all working with your API now and am just reading through to look for ai oddities.

Bicubic splines would be great!

Is there any chance of getting support for complex y values though? that is the clunkiest thing in our use of Fast interpolations right now.

My helper functions make the extrap boundary condition pretty nicely for now, but let me know if you implement that natively too 🙏

mgyoo86:
@logan-nc 
Thanks for the great suggestion!
I’ve been also thinking about adding native support for complex values, but it might require a major refactor of the type system in FastInterpolations.jl.
Still, I believe it should be technically feasible and worth trying.
For now, I’ll finish up what I’m currently working on first, and then I’ll move on to implementing complex support!

logan-nc:
This is happening! It needs to happen now to prevent divergence from other development.
It is a good thing! The code is now 100% julia! 🎉 🎉 

It does change a lot though. So everyone should double check their work. 
@JaeBeom1019 please double check the inverse equilibrium 
@jhalpern30 see how the splines changed in vacuum and use this api in your 3D code
@mfairborn23 be aware of these big changes. If you are creating a lot of new splines, you may want to merge develop into your branch sooner rather than later

🤞  everything looks smooth! I tried my best. As a reminder, it's all julia, and significantly faster. The DIII-D-like example stability is almost unchanged. I did my best to test as I went. Good luck out there.

PR #143, opened 2026-02-03
Complex regularization of singularities instead of enforcing ideal crossing

matt-pharr:
This is a WIP, which may or may not work out well and therefore may be discarded. 

The goal is to add an epsilon `ϵ_res ~ sqrt(η)` such that we can set resonant singfacs = m - nq + i ϵ_res in the vicinity of rational surfaces to try and effectively allow island opening.


matt-pharr:
@claude review this PR

matt-pharr:
@claude what is forcefreestates?

matt-pharr:
This is a test comment to make sure Claude does not respond when it is not mentioned

matt-pharr:
@logan-nc ran some Claude tests here and it is looking good

PR #119, opened 2025-12-30
Top level organization of unified Perturbed Equilibrium code

logan-nc:
This branch has already merged in the changes from #118, and supersedes that PR.

I think it should be merged quickly, and not necessarily wait for full PerturbedEquilibrium functionality / benchmarking. This is because the organizational changes will quickly cause many conflicts if others are working on feature branches from the old develop. Let's get the organization reviewed by @matt-pharr and merged, then keep going on the details of the perturbed equilibrium

PR #145, opened 2026-02-06
3D Vacuum Part 1: Reorganizing vacuum file structure and force_wv_symmetry bugfix

jhalpern30:
This is part of a series of PRs for 3D vacuum that I am breaking up to be more manageable. Starting off is two things.

TL;DR - I changed file names + moved some functions around, and fixed a small bug. @logan-nc this should be a quick review, basically just if you agree with my file naming.

## File reorganization

Removes `Vacuum...jl` naming convention, since this isn't what we did for DCON and isn't required for Julia standard. Furthermore, `VacuumInternals.jl` was massive and contained a bunch of unrelated functionality.

New proposed structure. NO CODE WAS CHANGED in this process, only moved from file to file.

- `DataTypes.jl`
  - Holds: input/geometry structs and geometry initialization helpers.
- `MathUtils.jl`
  - Holds: general math utilities (Fourier transforms, interpolation/splines, arc‑length reparameterization) after extracting the kernel. For now, also the GPEC vacuum field functions, which don't work and can be moved later.
  - distribute_to_equal_arc_grid was moved from `DataTypes.jl` to here
- `Kernel2D.jl`
  -  new home for 2D kernel routines formerly in VacuumInternals
  - Holds: 2D toroidal Green’s‑function kernel, and all related functionality for toroidal Green's functions, elliptical integrals, and Legendre polynomials.
- (Coming soon) `Kernel3D.jl`
  - 3D analogy of `Kernel2D.jl`

## BUGFIX - force_wv_symmetry was not working properly
I chose to include this in this PR since it slightly changes the outputs of the vacuum code, and will make comparing PRs to develop annoying if I tried to do them all at once.

The original fortran for this section is
```fortran
      IF ( lsymz ) THEN

         do 601 l1 = 1, jmax1
            do 600 l2 = l1, jmax1
               vacmat(l1,l2) = 0.5 * ( vacmat(l1,l2)+vacmat(l2,l1) )
               vacmti(l1,l2) = 0.5 * ( vacmti(l1,l2)-vacmti(l2,l1) )
               rmatr(l1,l2) = 0.5 * ( rmatr(l1,l2)+rmatr(l2,l1) )
 600        continue
 601     continue
         do 621 l1 = 1, jmax1
            do 620 l2 = l1, jmax1
               vacmat(l2,l1) = vacmat(l1,l2)
               IF ( l1 /= l2 ) vacmti(l2,l1) = - vacmti(l1,l2)
               IF ( L1 == L2 ) vacmti(l2,l1) = 0.0
               rmatr(l2,l1) = rmatr(l1,l2)
 620        continue
 621     continue
      ENDIF
```
However, our Julia conversion was
```julia
    force_wv_symmetry && begin
        for l1 in 1:mpert
            for l2 in l1:mpert
                vacmat[l1, l2] = 0.5 * (vacmat[l1, l2] + vacmat[l2, l1])
                vacmti[l1, l2] = 0.5 * (vacmti[l1, l2] - vacmti[l2, l1])
            end
        end
    end
```
we clearly left out the second do loop here. Adding this second loop in Julia is equivalent to just calling the `hermitianpart!()` function, which allows a very simple line
```julia
force_wv_symmetry && hermitianpart!(wv)
```

As documentation that these do the same thing:
image
and we can see that these matrices are in fact Hermitian now as expected, which is not occurring in the existing code.


logan-nc (approved):
This is going to cause me headaches for #119 but you are doing the right thing by chunking work into digestible PRa so I am going to reward that and merge - dealing with the headache on the other end.

Please note that an issue should be opened that requests we move as many basic "mathematical utilities" to a shared top level directory. I've created it in #119, but after that is merged the task should be to go through each directory (ForceFreeStates, Vacuum, etc.) and pull things out as appropriate. Part of this will be migrating some things. For example, we don't need to reinvent FFTs for vacuum - we can use a standard package.

PR #155, opened 2026-02-13
Citations

logan-nc:
This branch adds more specific citations.
I let Claude read the foundational papers and put specific citations where possible. 
It didn't find too many, which is good - didn't go crazy making things up. 
I also standardized the citation style, which cleans up the various Chance references we already had there.


jhalpern30 (approved):
Looks good. Note I also added full deprecation of is_closed_toroidal since this PR referenced some equations specific to that and it was removed from the WallGeometry struct so I think it only partially implemented anyway

PR #158, opened 2026-02-13
Use FastInterpolations Splines with no adaptations

### LINKED ISSUE #148: 2D splines should use hints if possible
I implemented the FastInterpolations 2D splines directly with no wrapper. 
I didn't use hints because I wasn't sure how for these 2D ones. There are plenty of cases where multiple splines are called at the same point, so shared hints would be great. 

@mgyoo86, do these work with hints?

logan-nc:
This PR has been made possible by the many great feature adds to FastInterpolations from @mgyoo86!

We no longer need a `src/Splines` submodule at all. The whole `src` can use FastInterpolations directly, as all our needs are covered by them now. This PR closes #148.


test/runtests_spline.jl
```diff
@@ -1,68 +1,17 @@
...
-    # @testset "BicubicSpline" begin
-    #     ... (tests removed)
-    # end
-end
-
-# NOTE: BicubicSpline has been removed - use native FastInterpolations.cubic_interp instead
-# @testset "BicubicSpline - PeriodicBC" begin
-#     ... (tests removed)
-# end
+# NOTE: Splines module has been removed - all spline functionality now uses FastInterpolations directly
```
jhalpern30: There's no need for this file to even be here anymore - this is just unit testing spline things that I assume are tested in fastinterpolations

jhalpern30 (approved):

jhalpern30:
@claude remove all unit testing for the splines

jhalpern30:
@claude I approve the removing spline unit tests commit, please do it

PR #159, opened 2026-02-14
Goodbye to the fortran

jhalpern30:
This PR removes all existing mention of the Fortran code (except in my private vacuum_fortran branch). This includes deleting the vacuum source code, fortran build/compiling, mentions of Fortran in the documentation, wrapper functions, etc.

I think I got everything, but there's likely some areas that I missed out on and we may be removing small bits here and there for a while. Also, I think the setup instructions + CLAUDE.md/copilot instructions will need an overhaul, but our documentation on that is lacking a bit anyway so I will reserve that for a future PR.


logan-nc (approved):
good

Issue #142, opened 2026-01-28
Fine vacuum gridding leads to numerical instability


matt-pharr:
Vacuum does not automatically adjust the number of grid points it counts as "near the singularity" for the purpose of separating the singularity into a non-singular and an analytic term. This means that when the number of poloidal grid points gets sufficiently high, we start getting points close enough to the singularity that they start blowing up and the accuracy starts to suffer. This was originally noticed by Stuart Benjamin, I believe. A solution to this may be to adjust the number of designated singular points based on the number of poloidal grid points.

jhalpern30:
@matt-pharr is there a known equilibrium/resolution where this bug appears and/or a script that reproduces and plots it? i.e. "Solovev ideal example when mtheta > 2048". Ok if not, but don't want to repeat work tracking one down

matt-pharr:
@jhalpern30 not off the top of my head but you could ask Stuart Benjamin as I think he tracked this down at some point.

logan-nc:
@jhalpern30 it seems like things are working robustly now... without a concrete example of something going bad, I'd argue for closing this rather than keep it around as a ghost. Do you agree or is there a concrete example of something going wrong to address?

jhalpern30:
Agreed - I thought this was a bigger issue than it appears that it is, I took some test cases and ran up to like mtheta = 8096 and never observed this, so I think its an edge case. If we end up finding a legitimate reason I can add in a fix, but not worth the effort rn

PR #160, opened 2026-02-16
3D Vacuum Part 3: External Gaussian quadrature and cleaned up lagrange interpolation interface

### LINKED ISSUE #129: Explore the use of an external Gaussian Quadrature package
The Vacuum component and the in-progress Galerkin integrator for the Newcomb Euler-Lagrange equation both use gaussian quadrature to compute certain integrals. There are two main packages for this in Julia; [GaussQuadrature.jl](https://juliapackages.com/p/gaussquadrature), and [FastGaussQuadrature.jl](https://github.com/JuliaApproximation/FastGaussQuadrature.jl). Each of these packages should be explored and their advantages/disadvantages considered to potentially exploit any speedup/simplification opportunities.

jhalpern30:
Tl;DR - clarified and extracted Lagrange interpolation and Gaussian quadrature logic (+ used an external package instead of hardcoding coefficients), removed some unnecessary allocations 

This is part 3 (and the last non-3D related) PR for VACUUM developments made on my 3D vacuum branch. This closes #129 by using an external Gaussian quadrature package (FastGaussQuadrature) instead of hardcoding the coefficients. On top of that I also made a clear lagrange interpolation interface and reduced some allocations that were being made in the tight loops, such as in the source dx_dtheta and dz_dtheta calls that were being repeated in every observer point.

Note, I did NOT add the ability for the user to vary either the Lagrange interpolation order of Gaussian quadrature order, since I don't see that as super necessary right now and is difficult to do in a type-stable/compiler-optimizable way. However, this extracts the sections that are related to each of these numerical methods in a very clear way such that we could do this without much difficulty in the future. For example, this might be useful when tackling #142. 

I was able to reduce the allocations a bit on this branch - it didn't seem to affect the runtime much, but a win is a win. Output eigenvalues stayed the same in these tests.
```
DIIID
Vacuum.compute_vacuum_response (n=1): 0.153s, 52.55 MiB alloc
Solovev
Vacuum.compute_vacuum_response (n=1): 2.042s, 450.23 MiB alloc

Develop
DIIID
Vacuum.compute_vacuum_response (n=1): 0.148s, 54.79 MiB alloc
Solovev
Vacuum.compute_vacuum_response (n=1): 2.092s, 480.05 MiB alloc
```


jhalpern30:
@claude summarize and review this PR

src/Vacuum/Kernel2D.jl
```diff
@@ -1,48 +1,74 @@
...
+end
+
+"""
+    get_lagrange_stencils(gaussian_points)
+
+Return cached 5-point Lagrange stencils keyed by Gaussian order. Initializes
+and caches the stencils on first use for a given order.
+"""
+function get_lagrange_stencils(gaussian_points::AbstractVector{<:Real})
+    order = length(gaussian_points)
```
logan-nc: I will approve this as it works and is beautiful... but as a general note I'd say it is perhaps more beauty than brawn. If really trying hard to optimize for a hot loop, it seems like we shouldn't be bothering with recomputing the `length` over and over again. It's fixed, right? maybe 8 or 32? could be passed directly as an arg. Similarly, I suspect all the logic to do something on the first get is overkill if in sudo code,
```
x = 8
something_dict = Dict()
do i=0,100
    y = i * get_something(x) # fills dict on first call
end
```
could more readably and efficiently be
```
x = 8
something = set_something(x)
do i=0,100
    y = i * something
end
```

but maybe I am not seeing some magical use case here (I did just start the review after all). The `length` and `get` calls are presumably hyper optimized low level things, so I guess they aren't the worst to call in hot loops. Just pointing out they aren't nothing.

src/Vacuum/Kernel2D.jl
```diff
@@ -102,30 +128,43 @@ function kernel!(
...
+    # Precompute 5-point Lagrange stencils for the 8-point Gaussian nodes.
+    stencils_left, stencils_right = get_lagrange_stencils(GL8.x)
+    sing_idx = zeros(Int, 5)
+    stencil = zeros(5)
+
+    # Precompute source derivatives on the theta grid once used in Simpson integration
+    # The Gaussian singular-panel points are off-grid, so those still use spline evaluation directly.
+    dx_dtheta_grid = Vector{Float64}(undef, mtheta)
+    dz_dtheta_grid = Vector{Float64}(undef, mtheta)
+    @inbounds @simd for i in 1:mtheta
```
logan-nc: The `@simd` here implies to me this is a performance critical loop. However, for `@simd` to help loop iterations must be independent (no iteration depends on previous results). FastIterpolations is already highly optimized, and I think using their native api is better than using a manual loop with `@simd`. For example, using hints in loops makes the spline evals O(1) instead of O(N):
```
hint = Ref{1}
for i in 1:mtheta
   y1 = itp1(x[i], hint=hint)
   y2 = itp2(x[i], hint=hint)
end
```

Even better, you can just evaluate the full array directly (no manual loop):
```
y1 = itp1.(x)
```
Even better-better, you can make a SeriesInterpolant (see https://projecttorreypines.github.io/FastInterpolations.jl/stable/guides/api_selection/):
```
d1xy_itp = cubic_interp(theta0, [x0, y0])
dx_dtheta_grid, dy_dtheta_grid = d1xy_itp(thets_grid)
```

@mgyoo86, do we need to set `search=LinearBinary()` and provide a hint when we evaluate for a monotonic vector or this this already optimized for internally?

logan-nc (approved):
This looks quite nice. 
I only had two comments. One is truly just a comment - I doubt it will impact performance a ton. The second is about splines - this could actually impact performance so please consider updating the spline usage. Since it looks like it works now though, I am ok if you want to merge this PR and do the spline changes in a separate PR.

src/Vacuum/Kernel2D.jl
jhalpern30: I wouldn't say the `get_lagrange_stencils` is in a hot loop. Its called at the head of the kernel function, which is only called once (or four times with walls, but we cache it even then). The actual hot loop is inside the Gaussian integration, which is just indexing into the stencils 
```julia
# Get the stencil weights for the Gaussian point
stencil .= leftpanel ? stencils_left[ig] : stencils_right[ig]
...
@inbounds for stencil_idx in 1:5
    greenfunction[j, sing_idx[stencil_idx]] += G_n * stencil[stencil_idx] * GL8.w[ig] * dtheta
end
```
So the current state of the code is actually your second code block already:
```julia
stencils_left, stencils_right = get_lagrange_stencils(GL8.x)
do i=obs_points, src_points, etc.
   y = i * stencil
end
```
However, Claude has some other good suggestions below for this interface that I'm going to add.

src/Vacuum/Kernel2D.jl
jhalpern30: Great suggestion here. For now, I'm going to replace this section with 
```julia
dx_dtheta_grid = d1_spline_x.(theta_grid)
dz_dtheta_grid = d1_spline_z.(theta_grid)
```
(so the even better option). I did not know series interpolants were a thing, so that's definitely interesting. However, I'm going to wait on this since I don't see the point of doing a single series interpolant - I think if we implement it, we could just make all splines a [x, z] spline to simplify things. This would extract nicely to 3D for a [x, y, z] spline, so I already have collapsed the x, z grids into a single r = [x, y, z] grid in 3D in #163

logan-nc: OK for now. From the api link I had above, the SeriesInterpolant,
> - Uses unified matrix storage with SIMD-optimized point-contiguous layout.
> - 10-120× faster for scalar queries due to cache locality.

This is very similar to the Fortran spline optimizations. I broke out the splines in ForceFreeStates because there were a ton of evals when only one thing was needed (e.g. evaluation of all 4 profile splines when only q is needed). For things like this where we are always going to need all the functions at every evaluation point, the SeriesInterpolants are definitely the way to go in the end!

logan-nc:
🎉

PR #131, opened 2026-01-07
Vacuum response for 3D geometries

jhalpern30:
WIP - extended the existing vacuum code from 2D geometries to 3D geometries

PR #163, opened 2026-02-20
3D Vacuum Part 4: Extension of the VACUUM code to 3D geometries + unit test cleanup

jhalpern30:
New version of #131 since the merge conflicts with develop were awful so starting from scratch. This PR extends the axisymmetric VACUUM code to 3D geometries, but currently only has capabilities to run with axisymmetric 3D geometries. Nonaxisymmetric surfaces will be introduced in a future PR.

There are several main focuses of this PR

- the 3D vacuum code itself
- overhaul of the vacuum unit tests with an addition of 3D vacuum unit tests
- Introduction of an in-place version of the `compute_vacuum_response` function and a cleanup of how we call the function in 2D. We now create the input struct, and the looping over n and views into the larger matrices is handled internally to the vacuum module

I was able to use some clever tricks and multiple dispatch to have both the 2D and 3D approaches use the same `compute_vacuum_reponse` function. The 3D code is run if `nzvac>1` in the TOML file, and 2D code is run if `nzvac=1 (default)`. Branching/dispatch within Vacuum is further handled internally based off this parameter. 

## Major questions for @logan-nc
- I removed the `cn0` functionality, since I didn't see how it would fit in multi-n and it didn't seem to be being used. Was this needed?
  - (EDIT: based on the discussion below, I think DCON also breaks for n = 0 (there are some places where we divide by n) - so I think this is another long deprecated path that can be removed)
- I am unsure of how to handle the `n<0` logic that was carried over from the Fortran, I'm not sure if it does anything.  Should I try to do this for all n<0 modes in the nlow:nhigh range for 3D or multi-n? From DataTypes.jl:
```julia
    # Extract plasma surface geometry at this psi
    r, z, ν = extract_plasma_surface_at_psi(equil, ψ)

    # TODO: this was in Free.jl - is this general for GPEC too?
    # Invert values for n < 0
    if n < 0
        ν .= -ν
        n = -n
    end
```
## Minor notes/comments
- I consolidated the `apply_kernelsign` logic into a few lines without the function call
- The full size `grad_green` and `green_temp` arrays are now only allocated as large as they need to be depending on if a wall is present. Prior to this, we'd allocated 4x the space we needed, and then take a view into the matrix. In 3D, this is a very nontrivial allocation, and this is more efficient. For now, the `grri/grre` matrices preserve the zero padding to the number of rows if there is no wall to avoid confusing in perturbed equilibrium down the line
- Small reorganization of vacuum files - `equal_arc_grid` and `extract_surface_at_psi` now make up utilities along with two other inline functions I needed, and the untested `compute_vacuum_field` sub functions are in `Field.jl`
- I modified the Fourier transform functions. One of the better cleanups is replacing `m00` and `l00` with `row_offset` and `col_offset` keyword arguments

## Things not implemented
- interface with multi-n in the perturbed equilibrium code. I believe the green's matrices are now all in a form that can be handled easily, but I did not go digging too deep here to make it work since the physics outputs from this module are still inaccurate
- interface for nonaxisymmetric equilibria. This would have been too messy for this PR


jhalpern30:
@claude summarize and review this PR. Address all issues relating to possible bugs, clarity, unclear functionality, performance issues, etc.

jhalpern30:
@logan-nc I am going to try to get this cleaned up and ready for review tomorrow - I still need to fix a few small things, add plots for benchmarking, look at Claude's suggestions, etc. but considering this is a big PR I'll just ping you now in case you have time to skim through this prior to then for any larger suggestions. No rush on this, just wanted to get my base code merged in since I did a big cleanup of it today and its fairly clean. I'll ping again once its ready for a final review

jhalpern30:
# Benchmarks
## Accuracy
### Verifying 2D results agree with develop (identical up to all digits shown here)
Solovev ideal example
```
Computing free boundary energies
[ Info: Calculating conformal wall shape 8.52e-02 m from plasma surface.
[ Info: Re-distributing wall points to equal arc length spacing (assumes closed, toroidal wall).
Least Stable Eigenmode Energies:
  Plasma = -1.102e+01 +2.883e-06i
  Vacuum = +1.058e+01 +9.860e-19i
  Total  = -4.372e-01 +2.883e-06i
Free-boundary mode unstable for n = 1.
Writing saved data to jpec.h5

============================================================
Run time: 3.999e+00 seconds
Normal termination.
============================================================
```
DIIID-like ideal example
```
Computing free boundary energies
[ Info: Using no wall
Least Stable Eigenmode Energies:
  Plasma = -3.471e-02 +1.912e-05i
  Vacuum = +1.742e+00 -5.194e-17i
  Total  = +1.707e+00 +1.912e-05i
All free-boundary modes stable for n = 1.
Writing saved data to jpec.h5

PERTURBED EQUILIBRIUM START
```
### Verifying 3D results converge to 2D
For the Solovev ideal example 3D case (mtheta = 128, nzeta = 64)
```
Computing free boundary energies
[ Info: Average grid spacing [m]: dθ=0.0375, dζ=0.0789, aspect ratio=2.1
[ Info: Calculating conformal wall shape 8.52e-02 m from plasma surface.
Least Stable Eigenmode Energies:
  Plasma = -1.102e+01 +2.909e-06i
  Vacuum = +1.058e+01 +2.883e-17i
  Total  = -4.377e-01 +2.909e-06i
Free-boundary mode unstable for n = 1.
Writing saved data to jpec.h5

============================================================
Run time: 1.968e+01 seconds
Normal termination.
============================================================
```
And here is the convergence of the 3D case to 2D versus nzeta for the three different coordinate systems
image

## Performance using @btime
### Develop
```
  2.899 s (12820733 allocations: 1.45 GiB)
```
DIIID-like ideal example 
```
  1.879 s (8425481 allocations: 1.78 GiB)
```
### This PR (negligible change in runtime, small decrease in allocations, likely due to in-place compute_vacuum_response)
Solovev ideal example
```
  2.983 s (12820789 allocations: 1.37 GiB)
```
DIIID-like ideal example
```
  1.895 s (8425492 allocations: 1.66 GiB)
```

jhalpern30:
I believe I might understand this logic: 
```julia
# TODO: this was in Free.jl - is this general for GPEC too?
# Invert values for n < 0
if n < 0
    ν .= -ν
    n = -n
end
```
From a comment I saw while reviewing #169: 
`assert n >= 1 "PnQuadEntry is only defined for n ≥ 1 (Γ(1/2 - n) diverges at n = 0)"`
This implies that this logic is here because the toroidal Green's function formulas are not defined for an n<0 case, so even though we can run DCON with n<0, we include this hack to run VACUUM with n<0. This works because the places `n` appears are:
- Green's and gradient Green's function. I think these are equal for n = -n, at least analyzing the equation below from Chance 1997, so this flip does not affect those outputs.
- Fourier transforms, where n appears as `exp(i(m * theta - n * nu))`. Therefore, if we flip n = -n, we also flip nu = -nu, which gives us `exp(i(m * theta - n * nu)) = exp(i(m * theta - (-n) * (-nu)))` so we preserve the correct transform. 

@matt-pharr I believe you are the expert of this section of the code. Does my logic make sense to you? 

What this means for me: this sign flip is a consequence of using the 2D toroidal Green's functions, as such it is NOT needed for the 3D implementation. But it is needed in 2D for cases where n<0, so it needs to be handled properly in multi-n cases. I think the way to handle this is: since the net effect of flipping both n and nu on `compute_fourier_coefficients` is nothing, we only need to pass in the absolute value to the kernel
```julia
kernel!(grad_green, green_temp, plasma_surf, plasma_surf, abs(n))
```
But this might be too simple or else it would have been done in the fortran. Unless this is another one of those "I only want to hack the interface between the codes and not the code itself", where Glasser made this to only modify DCON and not VACUUM. I will do a test to confirm this before I add it.

image

jhalpern30:
Ok, this branch has been merged with main - I think I have preserved all of @mgyoo86's performance optimizations during this process. Using btime on my laptop:
```
My branch
DIIID
  860.908 ms (385228 allocations: 612.08 MiB)
Solovev
  1.447 s (202765 allocations: 493.78 MiB)

Develop
DIIID
  1.057 s (385155 allocations: 620.14 MiB)
Solovev
  1.213 s (202687 allocations: 495.20 MiB)
```
The variation between btime runs seemed to be larger than the variation in between the two branches, so I think we're good.

logan-nc:
> Unless this is another one of those "I only want to hack the interface between the codes and not the code itself", where Glasser made this to only modify DCON and not VACUUM. I will do a test to confirm this before I add it.

This reads as if you think it unlikely, but I'd say this should be the default assumption. Hacks like this are a major reason for the rewrite as a unified code! 

You had also mentioned surprise at not caching the P_n recursion due to love of globals. Remember the authors of the DCON and VACUUM code were different. DCON has a lot of globals and careful opti.izations for things like fast splines. VACUUM had a relatively more straight forward job that was done relatively quickly - so was perhaps less obsessively optimized.

logan-nc:
@claude review this and let me know if all the original review points have been addressed. In particular, does the SingularCoupling logic correctly support multiple n?

logan-nc (approved):
Looks great! 

I resolved merge conflicts and cleaned up a few straggler PR review action items. @jhalpern30 has 🟢  to merge

matt-pharr:
> @matt-pharr I believe you are the expert of this section of the code. Does my logic make sense to you?

@jhalpern30 Yes this reads correctly to me.

jhalpern30:
> I resolved merge conflicts and cleaned up a few straggler PR review action items. @jhalpern30 has 🟢 to merge

Final changes: implementing logic from #180  - removing the n < 0 section and making vacuuminput immutable now that we never have to consider flipping the sign of nu. Also made the PATCH_RAD resizing robust in both directions instead of just toroidal

PR #164, opened 2026-02-21
(perf): Tidy FastInterpolations API for type stability

mgyoo86:
## Motivation

JPEC currently suffers from high heap allocation pressure -- GC time alone accounts for ~20% of total runtime. This PR is the **first in a series** of optimization passes aimed at reducing allocations and GC overhead across the codebase. The focus here is on eliminating type instabilities in the FastInterpolations API surface, which cause dynamic dispatch and small heap allocations on every interpolant call in hot loops (equilibrium solver, ODE integration, stability matrices).

## Summary

FastInterpolations v0.2.13 introduced the `AbstractExtrap` type hierarchy (`ExtendExtrap`, `WrapExtrap`, etc.) as a type-stable replacement for the existing Symbol-based `extrap` keyword (`:extension`, `:wrap`). The old Symbol API is still supported for backward compatibility, but this PR proactively migrates all JPEC call sites to the new struct-based idioms. A further API cleanup may come with v0.3.0.

Along the way, this PR also tidies JPEC's interpolation call sites to follow more idiomatic FastInterpolations patterns: one-shot evaluation, Range-typed grids, and shared options tuples.

### 1. Typed extrapolation via `AbstractExtrap`

```julia
# Before: Symbol-based, runtime dispatch (internal branching on each call)
extrap = :extension
extrap = (:extension, :wrap)

# After: struct-based, compile-time dispatch (compiler specializes & inlines)
extrap = ExtendExtrap()
extrap = (ExtendExtrap(), WrapExtrap())
```

- **`extrap`**: `AbstractExtrap` subtypes replace Symbol keywords, enabling compiler specialization and inlining.
- **`bc`**: Scalar `bc=CubicFit()` auto-broadcasts to both dimensions, replacing the redundant `bc=(CubicFit(), CubicFit())`.

### 2. One-shot interpolation (Vacuum)

```julia
# Before: creates intermediate interpolant object, then broadcasts
result = cubic_interp(xs, fs; bc=PeriodicBC()).(query_pts)

# After: single call, no intermediate object
result = cubic_interp(xs, fs, query_pts; bc=PeriodicBC())
```

Avoids allocating the intermediate interpolant when it's only used for a single batch evaluation.

### 3. Range grids for uniform axes

```julia
# Before: collected Vector, O(log n) binary search per lookup
grid2d = (psi_nodes, collect(theta_nodes))

# After: Range stays as Range, O(1) arithmetic index lookup
grid2d = (psi_nodes, theta_nodes)  # theta_nodes::StepRangeLen
```

May introduce machine-precision level differences compared to the `Vector` path.

### 4. Parameterized `FourFitVars{S, Opts}`

Previously `FourFitVars` held its `CubicInterpolant` fields with abstract type parameters, so every access required dynamic dispatch and heap allocation. This PR parameterizes the struct as `FourFitVars{S, Opts}` where `Opts` captures the concrete interpolation options tuple, making the interpolant types fully concrete and enabling type-stable placeholder construction via `@kwdef` defaults.

### 5. Shared interpolation options

Extracted repeated `bc`/`search`/`extrap` keyword arguments into shared `NamedTuple` variables (`opts2d`, `itp_opts2d`, `flux_opts`) to reduce repetition and ensure consistency across call sites.

## Modules touched

| Module | Key changes |
|---|---|
| **Equilibrium** | Typed extrap, shared opts, Range grids |
| **ForceFreeStates** | Parameterized `FourFitVars{S,Opts}`, centralized `itp_opts` |
| **Vacuum** | One-shot interpolation API |
| **PerturbedEquilibrium** | Typed extrap in `SingularCoupling` |
| **JPEC (main)** | Updated imports, typed extrap for `locstab` |

## Known limitation

`OdeState.wvmat` placeholder still uses the zero-arg `_empty_series_interp_complex(n)` (not type-stable). Intentional -- `wvmat` is always overwritten before use and is not on a hot path. Full type-parameterization of `OdeState` would require significant structural changes for minimal benefit.

## Benchmark Comparison

Measured with `@timev` on each example, averaged over 10 loops, comparing `develop` (`48ede89`) vs this PR. Single-threaded BLAS, warmed JIT, HDF5 output disabled for DIII-D. GC time varies — included for reference only.

**`DIIID-like_ideal_example`**

| | Allocations | Alloc size | Avg runtime (N=10) | GC % |
|---|---|---|---|---|
| `develop` (`48ede89`) | 8.545 M | 2.017 GiB | 3.711 s | 27.5% |
| This PR | 5.504 M | 1.948 GiB | 2.988 s | 19.3% |
| **Total change** | **-35.6%** | **-3.4%** | **-19.5%** | 27.5% → 19.3% |

**`Solovev_ideal_example`**

| | Allocations | Alloc size | Avg runtime (N=10) | GC % |
|---|---|---|---|---|
| `develop` (`48ede89`) | 12.957 M | 1.498 GiB | 5.965 s | 23.4% |
| This PR | 9.931 M | 1.428 GiB | 4.611 s | 15.7% |
| **Total change** | **-23.3%** | **-4.7%** | **-22.7%** | 23.4% → 15.7% |


jhalpern30 (approved):
These all look great too. Seems like all changes are specific to the splines so I'll trust your expertise and documented improvements there, I see no issues with the interface to the rest of the code. I'll make sure to keep the one-shot API in mind if I see other applications for it. Thanks!

mgyoo86:
@jhalpern30 
Regarding the new changes from the merge of 'develop' to this PR, 
I've made a very small change for FastInterpolations interpolant usage for vector query.
```julia
# Before: used broadcast operation, which is unnecessary
dx_dtheta_grid = d1_spline_x.(theta_grid)
dz_dtheta_grid = d1_spline_z.(theta_grid)

# After: use pure API that can take the vector query
dx_dtheta_grid = d1_spline_x(theta_grid)
dz_dtheta_grid = d1_spline_z(theta_grid)
```

`FastInterpolations`'s interpolants can accept vector inputs natively, so broadcasting is redundant and could be less efficient. 

I'll update PR #165 as well to stay consistent with develop.

logan-nc (approved):
Looks great. Thanks for all the improvements!

src/Vacuum/DataTypes.jl
```diff
@@ -196,9 +196,11 @@ function PlasmaGeometry(inputs::VacuumInput)
     # Interpolate arrays from input onto mtheta grid
     θ_in = range(0.0, 2π; length=length(inputs.r)) # VacuumInput uses [0, 2π] grid
     θ_out = range(; start=0, length=inputs.mtheta, step=2π/inputs.mtheta) # VACUUM uses [0, 2π) grid
-    x = cubic_interp(θ_in, inputs.r; bc=PeriodicBC()).(θ_out) # no endpoint handling needed!
-    z = cubic_interp(θ_in, inputs.z; bc=PeriodicBC()).(θ_out)
-    ν = cubic_interp(θ_in, inputs.ν; bc=PeriodicBC()).(θ_out)
+
+    # Use one-shot API with PeriodicBC
+    x = cubic_interp(θ_in, inputs.r, θ_out; bc=PeriodicBC()) # no endpoint handling needed!
```
logan-nc, line 201: @mgyoo86 do these not need `extrap` kwargs because you know theta_out is the same range of theta_in [0,2pi]? 

Is there any performance boost to not using extrapolation? Most of our splines never actually get extrapolated

src/ForceFreeStates/EulerLagrange.jl
```diff
@@ -352,7 +352,7 @@ function integrate_el_region!(odet::OdeState, ctrl::ForceFreeStatesControl, equi
     # Advance differential equation from psi_start to psi_end
     rtol = compute_tols(ctrl, intr, odet, chunk.ising) # initial tolerances
     prob = ODEProblem(sing_der!, odet.u, (chunk.psi_start, chunk.psi_end), (ctrl, equil, ffit, intr, odet, chunk))
-    sol = solve(prob, BS5(); reltol=rtol, callback=cb)
+    sol = solve(prob, BS5(); reltol=rtol, callback=cb, save_everystep=false, save_end=true)
```
logan-nc, line 355: @jhalpern30 this is fine and we should have been using it all along. It turns out that by default, ODE solvers store enough extra information at each step to interpolate the solution continuously between steps and the `sol` it returns is actually callable like `sol(psifac)`. But we never use this feature! We construct our own `odet.u` using manual callbacks like a bunch of dinosaurs from fortran times. We should probably take a hard look at this and see if doing everything with the native saving features would be better, but for now we definitely don't need to be double saving in `sol` and `odet`

src/Vacuum/DataTypes.jl
mgyoo86: With `PeriodicBC()`, `extrap` is internally forced to `WrapExtrap()`, so queries are always wrapped into the domain.

About performance — extrapolation adds very little overhead. It's just simple boundary checks or coordinate wrapping, not extra spline computation. 
For vector queries with `PeriodicBC`, there's also a fast path: we check min/max of the query vector first, and if everything is inside the domain, wrapping is completely skipped in the loop, which is the exactly the case for above example.

For your scenario where queries are always inside the domain, I think it is a good idea to add a new kwarg that lets users assert this explicitly, skipping even the min/max check. I will consider adding this feature in a future update.

logan-nc:
@mgyoo86 will this be ready to merge into develop once #167 is merged here? Or do you have more features that we should wait for?

@jhalpern30's major vacuum code developments like #163 could probably benefit from these kinds of optimizations as well. I think the clearest thing to do might be to merge this into develop asap, keep merging the vacuum developments as they are and then add an additional optimization PR to the end of @jhalpern30's PR series that applies the best practices established here to the new 3D vacuum code. Do you agree @jhalpern30 and @mgyoo86?

mgyoo86:
@logan-nc 
I merged #167 into this PR, and I believe it is now ready to be merged into `develop`.

I also have another optimization PR ready, but it is independent of these memory allocation PRs (#164, #165, and #167). So I will open that new PR against `develop` once this PR is merged.

jhalpern30:
> @jhalpern30's major vacuum code developments like #163 could probably benefit from these kinds of optimizations as well. I think the clearest thing to do might be to merge this into develop asap, keep merging the vacuum developments as they are and then add an additional optimization PR to the end of @jhalpern30's PR series that applies the best practices established here to the new 3D vacuum code. Do you agree @jhalpern30 and @mgyoo86?

Agreed. The status of the 3D code is still in flux so I don't want to spend too much time optimizing it right now, but once its closer to its final state it could definitely use the pooling.

logan-nc:
@aoifeturner please also take a look at this PR's performance improvements. In particular, check out the pooling introduced by #167.

PR #165, opened 2026-02-22
(perf): Reduce allocations across modules

mgyoo86:
## Summary

Follow-up to #164 (FastInterpolations type stability), this PR continues the optimization pass by targeting **heap allocations**: unnecessary ODE output storage, missing `@views`, allocating linear solves, and type instabilities.

### 1. ODE solver: `dense=false`, `save_everystep=false`

```julia
# Before
sol = solve(prob, BS5(); reltol=1e-6, abstol=1e-8)

# After: skip dense interpolant / skip per-step storage
sol = solve(prob, BS5(); ..., dense=false)
sol = solve(prob, BS5(); ..., save_everystep=false, save_end=true)
```

`dense=false` for fieldline integration and LAR solver. `save_everystep=false` for ballooning and Euler-Lagrange ODEs where only the final state is needed.

### 2. In-place linear algebra: `lu!`/`ldiv!`

```julia
# Before
grri .= grad_green_interior \ grri

# After
F = lu!(grad_green_interior)
ldiv!(F, grri)
```

Applied in Vacuum (Green's function inversion) and Sing.jl (singular matching).

### 3. `@views`, pre-allocation, and in-place patterns

`@views` on array slices across hot paths, hoist loop-invariant allocations, and reorder FourFit evaluation to write directly into pre-allocated ODE buffers instead of through `vec(reshape(...))`.

### 4. Type stability fixes

Return type annotation on `direct_fieldline_int` to work around DiffEq `solve()` inference depth limit caused by deeply-nested FastInterpolations type parameters. Fix closure variable capture in `direct_position!` to prevent `Core.Box` wrapping.

## Benchmark

Measured with `@timev`, averaged over 10 loops, single-threaded BLAS, warmed JIT. GC time varies between runs.

**`DIIID-like_ideal_example`**

| | Allocations | Alloc size | Avg runtime (N=10) | GC % |
|---|---|---|---|---|
| `develop` (`48ede89`) | 8.545 M | 2.017 GiB | 3.711 s | 27.5% |
| PR #164 | 5.504 M | 1.948 GiB | 2.988 s | 19.3% |
| This PR | 3.402 M | 0.922 GiB | 2.566 s | 14.4% |
| **Total change** | **-60.2%** | **-54.3%** | **-30.9%** | 27.5% → 14.4% |

**`Solovev_ideal_example`**

| | Allocations | Alloc size | Avg runtime (N=10) | GC % |
|---|---|---|---|---|
| `develop` (`48ede89`) | 12.957 M | 1.498 GiB | 5.965 s | 23.4% |
| PR #164 | 9.931 M | 1.428 GiB | 4.611 s | 15.7% |
| This PR | 7.910 M | 0.889 GiB | 4.246 s | 8.9% |
| **Total change** | **-39.0%** | **-40.7%** | **-28.8%** | 23.4% → 8.9% |


jhalpern30 (approved):
Everything looks great here, especially the improved performance. 

I think my only open questions/comments are:

- Just want @logan-nc to confirm that the `save_everystep=false` logic is ok. I think yes since we are manually saving using our own `save_interval` parameter, so just saving the last step should be ok.  
- @mgyoo86, what's the benefit of wrapping `apply_green_function` and `Pn_minus_half...` functions with an in-place `!` version if we don't actually use them? Every call to `Pn_minus_half_1997` is still allocating a `P` vector, so I don't see the benefit here. Is there something I'm unaware of with Julia's optimizations that is occurring?

mgyoo86:
@jhalpern30 Yes, you are right. The current in-place `!` functions don't have any effect yet. I had a larger set of changes prepared internally and split them into multiple PRs; these ended up on this side of the split. However, providing the `!` interface upfront is standard Julia practice though, and I'll be putting them to use in the follow-up PR (coming very soon), so I think we can keep them as-is.

On the claude bot's feedback — applied the `mul!` change in `apply_green_function!` to eliminate the temporary allocation from the matrix-vector product. 

This PR was intentionally scoped to low-invasive fixes and refactors. The more substantial allocation reductions — including the `Pn_minus_half` hot path and wiring up the `!` variants — are coming in the next PR.

jhalpern30:
> @jhalpern30 Yes, you are right. The current in-place `!` functions don't have any effect yet. I had a larger set of changes prepared internally and split them into multiple PRs; these ended up on this side of the split. However, providing the `!` interface upfront is standard Julia practice though, and I'll be putting them to use in the follow-up PR (coming very soon), so I think we can keep them as-is.
> 
> On the claude bot's feedback — applied the `mul!` change in `apply_green_function!` to eliminate the temporary allocation from the matrix-vector product.
> 
> This PR was intentionally scoped to low-invasive fixes and refactors. The more substantial allocation reductions — including the `Pn_minus_half` hot path and wiring up the `!` variants — are coming in the next PR.

Gotcha, that makes sense. Then this is all good with me, I'll resolve the conflicts with develop and you can merge with #164 whenever you're ready. NOTE: since `nonsing_idx` is no longer used as of #160, I removed that optimization on those operations that existing in this PR during the merge commit [here](https://github.com/OpenFUSIONToolkit/JPEC/pull/165/commits/292fb29b598a1404617e2e27e4b607ea36f320c2)

And it looks like Claude agrees with my thoughts on the `every_savestep` logic
> save_everystep=false, save_end=true on EulerLagrange is correct — the integrator_callback! handles all custom saving via odet.u_store, and save_end=true is the right choice when the final state is read back via sol.u[end] after the solve.

PR #168, opened 2026-02-24
Name Change to GPEC

logan-nc:
This branch provides an example of what the changes will look like when we properly name this package to finally fully address #34 . 
In this instance, I've implemented the default, which is to carry over the Generalized Perturbed Equilibrium Code (GPEC) name. This examples shows some important points, like how the actual Julia package should be CamelCase and not too long to write out in words rather than an acronym. The executables, `toml`s, hdf5 outputs, etc. can all use the acronym and self-references in annotations / documentation can use it (as can papers, of course). I think this strikes a good balance, but others are welcome to weigh in if they would like to tip the scales a little more towards using more CamelCase or acronyms here or there.

Ultimately, we await @parkjk's approval to actually use this name. It is fine if @parkjk prefers to keep the Julia and fortran code names different. If so, then the Unified Perturbed Plasma Equilibrium Response code (UPER) might be the leading contender from #34. I suppose this could be `PerturbedPlasmaEquilibrium.jl`  in the Julia registry. 

- Rename Julia module: src/JPEC.jl → src/GeneralizedPerturbedEquilibrium.jl
- Update Project.toml package name to GeneralizedPerturbedEquilibrium
- Rename executable script: jpec → gpec
- Rename configuration files: jpec.toml → gpec.toml (examples, benchmarks, test data)
- Rename default output file: jpec.h5 → gpec.h5
- Update all using/import statements, docstrings, and comments throughout src/, test/, docs/, benchmarks/
- Update CLAUDE.md, README, and agent config files


logan-nc:
@parkjk agreed! We are a go.

PR #169, opened 2026-02-24
(perf): Cache n-dependent computations in `Pn_minus_half_2007!`

mgyoo86:
## Summary

`Pn_minus_half_2007!` is the dominant hot path of the entire JPEC pipeline (~50–70% of total runtime). It is called millions of times per run, but was recomputing expensive n-dependent values — especially `sinh` transcendentals, called tens of millions of times redundantly — on every single call, even though `n` never changes within a vacuum run.

This PR introduces a thread-safe cache (`PnQuadCache`) that computes all n-dependent values once on first use and serves all subsequent calls from memory. The result is a **~7× speedup** on the function itself, translating to **~1.4× end-to-end on DIIID-like** and **~2× on Solovev** examples.

### `src/Vacuum/PnQuadCache.jl` (new)

A thread-safe, per-`n` cache that pre-computes and stores all n-dependent intermediate values in a `PnQuadEntry` struct:
- `sinh`/`cosh` vectors
- Gamma normalization constants 
- Quadrature weights

The cache uses an inline/noinline split for fast access: a single atomic integer compare (lock-free) on the hot path, and a `SpinLock`-guarded `Dict` on first population. Since `n` is constant per run, the fast path hits ~100% of calls.

### `src/Vacuum/Kernel2D.jl`
`Pn_minus_half_2007!` updated to read from `PnQuadEntry` instead of computing n-dependent values inline. The only remaining per-call computation is the s-dependent `sqrt` for the quadrature denominator, which cannot be cached.

### `test/runtests_vacuum_julia.jl`
New regression tests confirms the numerical equivalence (rtol<1e-15) with the pre-optimization reference values.

## Benchmark Results

**`Pn_minus_half_2007!` benchmark** — `n = 1`, `s = 1.0001`:

| | Median time |
|---|---|
| `develop` | 583 ns |
| **This PR** | **82 ns** |
| **Speedup** | **~7×** |

**Example end-to-end benchmark** — 10 warm runs, min time (GC < 1% at min), single-threaded BLAS:

| | `develop` (minimum) | **This PR** (minimum) | Speedup |
|---|---|---|---|
| DIIID-like | 2.104 s | **1.512 s** | **~1.4×** |
| Solovev | 3.609 s | **1.850 s** | **~2×** |

**CI benchmark** — GitHub Actions `Build and Test Julia Package`:

| | Wall time | Speedup |
|---|---|---|
| `develop` | 12m 27s | |
| **This PR** | **3m 55s** | **~3.2×** |


logan-nc (approved):
Another wonderful speedup. @aoifeturner please take note of this as a great example of performance improvement opportunities!

I leave the final approval and merging of this to lead vacuum developer @jhalpern30

logan-nc:
@mgyoo86, @aoifeturner is a Columbia CS student joining this project to help us improve performance. If you'd be willing to share what tools you used to profile the code and identify opportunities like this, I am sure @aoifeturner would benefit greatly from your Julia wisdom!

mgyoo86:
I pushed a small fix addressing the claude bot's review:
- @test || @show — removed
- Thread safety — no issue; added a comment explaining why the plain Ref is safe (atomic sentinel acts as release fence).
- `NTuple` for `PnQuadEntry` fields — tried it, measured ~20% slower due to larger size of NTuple copy on hot-path. Keeping the current design.

mgyoo86:
> @mgyoo86, @aoifeturner is a Columbia CS student joining this project to help us improve performance. If you'd be willing to share what tools you used to profile the code and identify opportunities like this, I am sure @aoifeturner would benefit greatly from your Julia wisdom!

Hi @aoifeturner, hope the following helps you dig into the Julia code more easily!

For profiling, I mostly use `@profview` and `@profview_allocs` with VSCode Julia Extension — they show flamegraphs which is convenient for quick profiling. When I need more detail, Julia's built-in `Profile` module with `PProf` works well.

One tool I'd especially recommend is [`Infiltrator.jl`](https://github.com/JuliaDebug/Infiltrator.jl). You can use `@infiltrate` macro anywhere and it pauses execution with a local REPL — very handy for interactive debugging.

I also have some personal scripts/macros to ease my profiling/debugging workflow. Feel free to start a discussion anytime if you want to dig into any of this!

src/Vacuum/PnQuadCache.jl
```diff
@@ -0,0 +1,124 @@
...
+
+# ── Quadrature node constants (integration limits: xl=0, xu=5) ───────────────
+# Uses gausslegendre(32) directly so this file is self-contained
+# (included before Kernel2D.jl where GL32 is defined).
+# _PN_TG02[ig]   = tg0² = (agaus + x[ig] * bgaus)²
+# _PN_WANUMR[ig] = w[ig] * tg0 * exp(-tg0²)
+const _PN_AGAUS = 2.5
+const _PN_BGAUS = 2.5
+const (_PN_TG02, _PN_WANUMR) = let
+    x32, w32 = gausslegendre(32)
```
jhalpern30, line 27: These are precomputed at the top of Kernel2D.jl:
```julia
const GL32 = gausslegendre_rule(Val(32))
```
Please either use that here or delete it since this is the only place it is used. I have no preference, whatever you think is clearer for the implementation

src/Vacuum/Kernel2D.jl
```diff
@@ -539,58 +540,38 @@ function Pn_minus_half_2007!(P::AbstractVector{Float64}, s::Real, n::Int)
...
     if n * rhohat >= 0.1
 
-        # Integration limits
-        xl = 0.0
-        xu = 5.0
-
-        # Transform to integration interval
-        agaus = 0.5 * (xu + xl)
-        bgaus = 0.5 * (xu - xl)
+        entry = get_pn_quad_cache(n)
```
jhalpern30: Why is this called `entry`? This variable name doesn't make a lot of sense to me at first glance.

jhalpern30 (changes requested):
Looks great! Only two minor comments, and I can merge afterwards.

@logan-nc I am surprised caching the Legendre terms wasn't a Fortran feature given the love of globals. I think ultimately, this section still needs a cleanup since it still has the Fortran style, extremely short/unclear variable names, but that's out of the scope of the PR. I was just reminded of it as I read through.

src/Vacuum/PnQuadCache.jl
mgyoo86: I cleaned-up the GL32 definition as it is now dead-code as you pointed out.

src/Vacuum/Kernel2D.jl
mgyoo86: `entry` is the conventional name for a cached item, but I renamed it to `pn_cache` to make its role clearer at the call site.

mgyoo86:
I pushed minor-fix commits to address @jhalpern30 and @claude bot's comments.

jhalpern30 (approved):

jhalpern30:
Claude is really having a field day on this PR...

@logan-nc is it easy to make it only review once automatically, and after that as requested? No worries if not, but it clutters up the PR.

@mgyoo86 Thanks for addressing the changes, this looks great. Hadn't seen the `entry` notation before but that makes sense. I'll merge this

logan-nc:
We had it set up such that claude should only do reviews when explicitly requested in a PR comment. obviously something has broken that in the recent flood of merges... I'll try to get it back to sanity levels

aoifeturner:
> > @mgyoo86, @aoifeturner is a Columbia CS student joining this project to help us improve performance. If you'd be willing to share what tools you used to profile the code and identify opportunities like this, I am sure @aoifeturner would benefit greatly from your Julia wisdom!
> 
> Hi @aoifeturner, hope the following helps you dig into the Julia code more easily!
> 
> For profiling, I mostly use `@profview` and `@profview_allocs` with VSCode Julia Extension — they show flamegraphs which is convenient for quick profiling. When I need more detail, Julia's built-in `Profile` module with `PProf` works well.
> 
> One tool I'd especially recommend is [`Infiltrator.jl`](https://github.com/JuliaDebug/Infiltrator.jl). You can use `@infiltrate` macro anywhere and it pauses execution with a local REPL — very handy for interactive debugging.
> 
> I also have some personal scripts/macros to ease my profiling/debugging workflow. Feel free to start a discussion anytime if you want to dig into any of this!

Hi @mgyoo86 ! Nice work, seems like a fantastic speed up. Thanks for your words of wisdom with tools. I’ll definitely take you up on your offer to chat about your personal scripts and macros sometime soon- seems like it could be really beneficial.

PR #172, opened 2026-02-26
Restrict Claude review 

logan-nc:
Restrict Claude review to explicit comment requests

Trigger changed from pull_request_review:submitted to issue_comment:created with a job condition requiring '@claude review' in the comment body and the comment to be on a PR. Prevents auto-reviews on every submitted review or after subsequent commits/comments.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -1,24 +1,16 @@
 name: Claude Code Review
 
 on:
-  # pull_request:
-  #   types: [opened, synchronize]
-    # Optional: Only run on specific file changes
-    # paths:
-    #   - "src/**/*.ts"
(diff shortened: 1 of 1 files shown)


logan-nc:
There should be no review here until someone says the magic words

logan-nc:
Even despite multiple comments

logan-nc:
✅ 

Ok. And then we should get a review if we explicitly ask:

@claude review

logan-nc:
And then we should be able to comment and/or submit reviews and it won't trigger a new ai review. A new ai review will require another explicit ask. @jhalpern30 please submit a PR review to confirm

jhalpern30 (approved):

logan-nc:
it works!

PR #173, opened 2026-02-26
Perf/green gamma prefactor

logan-nc:
## Summary

   - Precompute `2√π · Γ(1/2 - n)` once per `kernel!` call instead of recomputing it on every call to `green`
   - Fix benchmark script to use `jpec.h5` output and run benchmarks in isolated subprocesses

   ## Problem

   `green()` previously computed `2 * sqrt(π) * gamma(0.5 - n)` on every invocation. Since `n` is the toroidal mode number — constant across an entire `kernel!`
   call — this was an expensive special function evaluation (~logarithms + rational polynomial) being repeated ~262,000 times per run (4 kernel blocks × 256²
   source/observer pairs), contributing ~11% of wall-clock time by sampling.

   ## Solution

   Add `gamma_prefactor::Float64` as an **optional positional argument** to `green`, defaulting to `2 * sqrt(π) * gamma(0.5 - n)`. This keeps the public API and
   all existing tests unchanged. `kernel!` and `_calculate_potential_chi` compute the prefactor once before their inner loops and pass it in.

   ## Performance

   Profiled on `examples/Solovev_ideal_example` (post-JIT warmup):

   | | Before | After |
   |---|---|---|
   | Run time | 2.95 s | 1.54 s |
   | Speedup | — | **~47% faster** |
   | `_gamma` profile samples | 92 (11%) | 0 |

   The speedup substantially exceeded the ~11% predicted from profiler sampling — `gamma()` stalls the instruction pipeline in ways that cascade beyond what
   wall-clock sampling alone captures.

   Numerical results are identical: ODE step counts (85/129/137) and eigenmode energies (`et[1] = -4.372e-01`) are unchanged.

   ## Test plan

   - [x] All 112 vacuum unit tests pass (`test/runtests_vacuum_julia.jl`)
   - [x] Solovev example produces identical eigenmode energies before and after
   - [x] ODE integration step counts unchanged

### DIFF
--- a/benchmarks/benchmark_git_branches.jl
+++ b/benchmarks/benchmark_git_branches.jl
@@ -40,8 +40,8 @@ julia benchmarks/benchmark_git_branches.jl --example examples/DIIID-like_ideal_e
 # Output Metrics
 
 For each branch/commit, reports:
-- Eigenmode energy (et[1]) from euler.h5
-- Integration steps from euler.h5
+- Eigenmode energy (et[1]) from jpec.h5
+- Integration steps from jpec.h5
 - Runtime (averaged over warm runs)
 - Git commit hash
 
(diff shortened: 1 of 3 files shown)


logan-nc:
@aoifeturner please profile this branch and the develop branch to confirm this improvement is real and not just a quirk of my laptop's CPU getting thrown off by other tasks during the testing.

@jhalpern30 you are just assigned in hopes that you skim quickly and tell me if this breaks something for you

jhalpern30:
> @jhalpern30 you are just assigned in hopes that you skim quickly and tell me if this breaks something for you

No issues from me, I barely touched numerics of the 2D stuff. The built-in Claude test script for runtime and outputs is cool.

aoifeturner:
I ran a few profiling tests on them both. For all tests, I used examples/DIIID-like_ideal_example.

New Runtime (Median across 5, ignoring first): 2.077s
Old Runtime (Median across 5, ignoring first): 2.309
~10.1% improvement with the precompute

Also generated flame graphs for both, was only able to find these gamma calls in the old runs (should be on my slides). Wanted to be sure so also ran @profile tests and dumped them into files and compared the files.
@profile 
OLD RUN: green() calls SpecialFunctions.gamma, with the profiler sampling execution inside gamma and _gamma, confirming that the gamma prefactor is being recomputed within the hot path. 
NEW RUN: Profiling the optimized branch shows no gamma() calls in the green() hot path. This confirms the prefactor is precomputed and reused, eliminating repeated special-function evaluations in the kernel loop.

HOWEVER, I will say that my runtime was a little inconsistent. There were a few runs where the old run outperformed the new run. This is likely due to fluctuations of my poor wifi at home, but I will say my 10% is rather different from your initial 47% so I may run some more tests later.

logan-nc:
Great @aoifeturner! I think your numbers sound more reasonable actually - I suspect my tests probably had a slow develop branch run due to other things running on my laptop. Seems like we all agree this is a solid win. I'll go ahead and merge!

PR #174, opened 2026-02-27
Replace ad-hoc Newton/bisection loops with Roots.jl

logan-nc:
## Summary

- Adds `Roots.jl` as a dependency and replaces all manually-coded Newton and bisection iteration blocks across `ForceFreeStates` and `Equilibrium` with `Roots.jl` calls, reducing boilerplate and improving robustness.
- Upgrades the hot-path `direct_refine` function from Newton (quadratic convergence) to Halley's method (cubic convergence), reducing the number of expensive `direct_get_bfield!` calls per flux surface refinement step.
- Fixes two bugs in `equilibrium_separatrix_find!` introduced during Fortran→Julia conversion.

## Changes by commit

| Commit | Module | Change |
|--------|--------|--------|
| `e7a2dd3` | ForceFreeStates | Replace Newton/bisection in `Free.jl`, `Sing.jl`, `EulerLagrange.jl` with `Roots.Newton` / `Roots.Brent` |
| `d75ed835` | Equilibrium | `find_separatrix_crossing` → `Roots.Brent`; `direct_fieldline_int` starting-R → `Roots.Newton` |
| `144788f7` | Equilibrium | `direct_refine` upgraded from Newton to Halley's method (`derivs=2`) |
| `beac6bda` | Equilibrium | `equilibrium_separatrix_find!` midplane crossing → `Roots.Newton` |
| `a50ba671` | Equilibrium | Fix spurious theta-wrapping and wrong convergence criterion in separatrix extrema loop; rename variables for clarity |

## Bug fixes (commit `a50ba671`)

Two bugs in `equilibrium_separatrix_find!` introduced during the Fortran→Julia conversion of `equil_out.f`:
1. **Spurious theta-wrapping** — not present in Fortran; unnecessary because the rzphi splines have periodic extrapolation and the extrema sit at θ ≈ 0.25/0.75.
2. **Wrong convergence criterion** — used `1e-12 * y_period` instead of `1e-12` (the normalised θ grid has period 1).

## Test plan

- [ ] Run `runtests_solovev.jl` — analytical equilibrium exercises `equilibrium_separatrix_find!` and `direct_refine`
- [ ] Run `runtests_ode.jl` — exercises ForceFreeStates Newton paths (`Sing.jl`, `EulerLagrange.jl`, `Free.jl`)
- [ ] Run `runtests_fullruns.jl` — end-to-end check
- [ ] Benchmark `direct_refine` Halley vs Newton on DIIID-like example using `benchmarks/benchmark_git_branches.jl`


logan-nc:
This should be a straightforward little PR to skim. In general, I thought all the manually coded newton iterations throughout the code might be likely bug opportunities and not super optimized. Hopefully, this standardizes and improves them.
I want to be a good citizen and get a second opinion on whether it is helpful or not before merging though.

jhalpern30 (approved):
Looks good, a nice cleanup that removes a lot of hardcoded numerical methods that are implemented better elsewhere. Seems like now there are just two hardcoded in Equilibrium in `direct_refine` and `equilibrium_separatrix_find!`. I'm assuming these weren't condensed because they're more complicated, but just noting in case it was an oversight.

I think the main comment I have is that there should be some output to confirm replacing Newton's method with the hardcoded Halley's method actually improves runtime. Since this is hardcoded and is a more complex numerical method, I'd say it moves away from robustness, so it isn't worth it unless the speedups are noticeable - while it may reduce the number of calls, the calls are also more expensive. It looks like this was on Claude's test plan but I don't see any results. 

If it does provide improvement, it can be cleaned up by using Roots.jl - based on the discussion on their [docs page](https://juliamath.github.io/Roots.jl/stable/reference/#Roots.Halley), you can provide a function that returns f, f', and f'' which we could easily do here to make the logic more clear

logan-nc:
Thanks @jhalpern30

The `equilibrium_separatrix_find` loop has a lot of calculations in it so it doesn't consolidate to one line well. I tried pulling them out into a separate function and then using roots.jl on that but found the resulting code longer and actually harder to understand because it involved fancy cache things. In the end I reverted it all and just updated the variable names and annotations. I think this is the cleanest way to leave it for now.

The `direct_refine` is not implemented because the AI admitted there is a bit of overhead to the Roots.jl calls that shouldn't be put into a hot loop.

I'll double check the Halley's method more explicitly before merging

logan-nc:
Summary of speeds in refining the equilibrium field line tracing:

                                                                                                                                                                                                                                                                                                                                                     
  ```
┌───────────────────────────┬───────────┬──────────┬─────────┬─────────┬────────────────────┐                                                                                                                                                                                                                                                      
  │      Implementation       │ Mean Time │ Std Dev  │   Min   │   Max   │     vs Fastest     │                                                                                                                                                                                                                                                    
  ├───────────────────────────┼───────────┼──────────┼─────────┼─────────┼────────────────────┤                                                                                                                                                                                                                                                      
  │ Manual Halley             │ 0.049 s   │ ±0.002 s │ -       │ -       │ baseline (fastest) │                                                                                                                                                                                                                                                      
  ├───────────────────────────┼───────────┼──────────┼─────────┼─────────┼────────────────────┤
  │ Roots.jl Newton (current) │ 0.054 s   │ ±0.001 s │ 0.053 s │ 0.056 s │ +10%               │
  ├───────────────────────────┼───────────┼──────────┼─────────┼─────────┼────────────────────┤
  │ Manual Newton (original)  │ 0.061 s   │ ±0.002 s │ 0.059 s │ 0.065 s │ +24%               │
  └───────────────────────────┴───────────┴──────────┴─────────┴─────────┴────────────────────┘
```

The Roots.jl newton ended up being fast enough plus it is consistent with other uses and clean to read. So I went with that. I've confirmed the DIIID example results are unchanged

**Comment by logan-nc** [comment]
DIIID example confirming the usual result:
```
Least Stable Eigenmode Energies:
  Plasma = -3.471e-02 +1.911e-05i
  Vacuum = +1.742e+00 -1.801e-16i
  Total  = +1.707e+00 +1.911e-05i
```

PR #177, opened 2026-02-28
ForceFreeStates: Hybrid adaptive callback + ldp-spaced saveat Gaussian reduction

logan-nc:
## Summary

Replaces the develop branch's per-step `DiscreteCallback` (doing tolerance switching, norm checking, and dynamic solution storage at every step) with a cleaner hybrid architecture:

- **Tolerance switching removed**: Single `eulerlagrange_tolerance` replaces `tol_r`/`tol_nr`/`crossover`. Develop's adaptive tolerance was the primary mechanism for controlling ODE step count; this branch removes it in favour of a uniform tolerance set to the desired near-rational precision.
- **Solution storage**: ODE built-in `saveat` with **ldp-spaced save points** (`sin²(i/(N-1)·π/2)`) per chunk — clusters points near both ends where solutions grow fastest. Exact pre-allocation from chunk structure; no dynamic `resize_storage!`.
- **Gaussian reduction**: Adaptive `DiscreteCallback` that fires only when the column growth ratio exceeds `ucrit`, sampled every K=10 accepted steps to reduce per-step O(np²) norm overhead.
- **One chunk per inter-rational region**: Removed `n_subchunks_per_region`/`edge_chunk_fraction` sub-chunking; callback handles conditioning adaptively.

### Key algorithmic improvements

**Growth-ratio sorting (correctness fix for multi-n)**: The callback baseline `unorm0` tracks column magnitudes at the last crossing or last reduction. Sorting by `current_norm / unorm0` (growth since baseline) rather than absolute norm correctly identifies the resonant column after a crossing, when it starts small but grows fastest. The develop branch's absolute-norm sort can zero the wrong column in multi-n cases.

**Duplicate fixstep guard**: On stiff problems (many accepted ODE steps between consecutive saveat points), multiple K=10 callbacks can fire with the same `n_saved`, producing the same `fixstep_val`. A second `apply_gaussian_reduction!` at the same fixstep creates an empty region in `transform_u!`, corrupting the transformed solution. The guard `odet.ifix > 0 && odet.fixstep[odet.ifix] == fixstep_val && return` prevents this.

**Post-integration ud_store**: `ud_store` is computed by a single post-integration `sing_der!` pass over stored steps, rather than inline in the callback. Enables correct ud after `transform_u!` applies to `u_store`.

**`u_store` memory layout**: Restructured from `(np, np, 2, nsteps)` to `(np, np, nsteps, 2)` so HDF5 component output `u_store[:,:,:,k]` is a single contiguous block, and `FieldReconstruction` inner-loop stride is halved.

**`ResponseMatrices.jl` index fix**: `extract_boundary_displacements` was using the old `u_store[:,:,1,step]` convention (causing `BoundsError` with `compute_response=true`); corrected to `u_store[:,:,step,1]`.

### Performance

Benchmarked on Solovev ideal example (mpert=32, no HDF5 output, 6 warm runs at `eulerlagrange_tolerance=1e-6`):

| Branch | Median | Best-3 avg |
|---|---|---|
| `perf/remove_callback` | 2.73s | 2.61s |
| `develop` | 2.81s | 2.71s |

Within noise — no regression.

### Removed

- `DiscreteCallback` `integrator_callback!` (tolerance switching, inline storage, per-step norm checking)
- `compute_solution_norms!`, `new`, `q_store` from `OdeState`/`ForceFreeStatesControl`
- `resize_storage!` (replaced by exact pre-allocation)
- `n_subchunks_per_region`, `edge_chunk_fraction` TOML parameters
- `tol_r`, `tol_nr`, `crossover` TOML parameters
- `BRANCH_STATUS.md` development tracking file
- `DiffEqCallbacks` dependency

### Added

- `ucrit` — Gaussian reduction threshold (max/min growth ratio, default 1e4)
- `save_npoints_per_chunk` — number of ldp-spaced saveat points per integration chunk
- `chunk_step_offset` field in `OdeState` — tracks global step offset for callback fixstep computation
- `unorm0` baseline vector (was `unorm0` scalar in develop) — per-column growth baseline
- `numunorms_init` pre-allocated fixup storage (replaces dynamic `ifix` tracking)

## Test plan

- [x] 44 ODE unit tests (`test/runtests_eulerlagrange.jl`)
- [x] Solovev single-n full run regression
- [x] Solovev multi-n full run regression
- [x] DIIID eigenvalue: `et[1] = 1.707` (matches develop)
- [x] Equilibrium, Vacuum, Fourier unit tests
- [x] `ResponseMatrices.jl` index order corrected for `compute_response=true` path


logan-nc:
@jhalpern30 I don't think this will get merged, but it was an interesting little journey I took over the weekend, and I thought you might be interested in the summary.

The initial plan was to completely remove the callback and use the inbuilt ODE `solve` solution saving. This worked beautifully in terms of cleaning up the code and performance... but the results were no good due to the solution imbalances causing loss of numerical accuracy. I tried applying gaussian reductions at the end of each chunk and even introducing sub-q-window chunking but in the end I decided the dynamic ucrit is the most robust way to ensure good numerics.

Along the way, I did learn that you give a `save_at` parameter to the `solve` call, which then means you can know the exact size of your solution array and avoid all the dynamic resizing code... but I am not 100% convinced this is worth it. Do you have any thoughts?

Things that also happened along the way that I DO like are (1) going to a single, clearly named tolerance instead of tol_r and tol_nr (note that this removes one of the old callback tasks that checked which tolerance should be used at every step) and (2) reordering the solution dimensions. If `save_at` is not convincing, it might be best to just start a new branch and PR for those two things.

Please note I made no attempts to clean up AI code here since I wasn't 100% sold on the final state of things from a big picture perspective. So if you look are any file changes, take them with a grain of salt for now

logan-nc:
superseded by #185

PR #180, opened 2026-03-02
Removing n < 1 code from JPEC

jhalpern30:
As discussed with @logan-nc and @matt-pharr and in #163 , there was some code in the Fortran in the DCON + VACUUM interface that did not make sense as it was treating n < 1, which we don't run with in 2D. Specifically, there was:
1) the `cn0` parameter in `compute_vacuum_response` which handled n = 0 runs
2) the `n < 0` logic in the `VacuumInput` data type

After discussion, we determined we don't need these in 2D, so this PR removes them. It also adds control logic in `JPEC.jl` that makes sure all toroidal modes are > 1.

It also introduces two small bugfixes I found along the way:
- added a missing trunctation to the `mlow` logic that was leading the value as a float
- Only saving the `di0` values at the rational surfaces if any rational surfaces exist in the HDF5. The current setup was erroring out if there were no rational surfaces


jhalpern30:
@claude review

jhalpern30:
@claude please fix the bug causing the unit tests to fail

jhalpern30:
@claude the unit tests are still failing, fix them

jhalpern30:
@logan-nc the tests are failing for some unrelated FastInterpolations issue, but this is good for review

logan-nc (approved):
Looks good. I added a little annotation just to make note of what we removed in case it ever comes up.

I also found the FastInterpolations API had breaking changes in a recent update. I will create a new PR for those fixes

PR #182, opened 2026-03-04
Update FastInterpolations API for v0.3.1 compatibility

logan-nc:
## Summary

FastInterpolations v0.3.1 introduced three breaking API changes that caused CI test failures on the \`no_neg_n\` branch (and will affect any branch using \`>=0.2.13\`):

- \`LinearBinary()\` renamed to \`LinearBinarySearch()\`
- \`NaturalBC()\` renamed to \`ZeroCurvBC()\`
- \`deriv=<Int>\` keyword argument replaced with \`deriv=DerivOp(<Int>)\`

This PR updates all 14 affected files and bumps the compat lower bound from \`>=0.2.13\` to \`>=0.3.1\`.

## Test plan

- [x] All 112 vacuum unit tests pass locally (was 97 pass + 3 errors before)
- [x] Full test suite passes locally
- [ ] CI passes


logan-nc:
@jhalpern30 and I found our github workflow unit tests were failing in a recent PR review (#180). It turns out FastInterpolations made a breaking minor version update to 0.3.1 and our toml's use of the `>=` picked these up to ill effect. 

This PR is my attempt to update to the new FastIterpolations API. I would appreciate @mgyoo86's review though to make sure we are getting it right and to let us know if there are any other subtle changes we might be missing.

mgyoo86:
@logan-nc 

This one was on me — the CI break came from an overly loose `[compat]` entry I added earlier for `FastInterpolations` mistakenly, which allowed the resolver to pull in `v0.3.x` and its breaking API changes. My intent was to stay on the `0.2.x` series so that breaking changes wouldn’t be introduced implicitly.

For a quick fix, I’ve reverted the code changes for now and pushed a minimal fix that only pins compat back to 0.2.13.
```
[compat]
# Before
FastInterpolations = ">=0.2.13"

# After
FastInterpolations = "0.2.13"
```

Separately, `FastInterpolations` has already moved on to `v0.4.0` as it’s been evolving quickly with incoming feedback. I’ll follow up with a dedicated PR to update JPEC for the newer API. As the `JPEC` is the first-class support target of `FastInterpolations`, I’ll make `JPEC` will use `FastInterpolations` in the best way :)

mgyoo86:
Just as an FYI: in Julia, packages in the `v0.x.y` range are considered “pre-1.0,” meaning the API is still evolving and breaking changes can happen more frequently. In this phase, the minor version (`x`) effectively acts like the major version (i.e., 0.2 → 0.3 can be breaking), while the patch version (`y`) is expected to stay non-breaking (bug fixes, small features, etc.).

Julia’s version/compat rules treat `0.x` specially for this reason: if you specify compatibility starting at `0.2` (e.g., `0.2.13`), it will not automatically upgrade to `0.3`. I mixed this up earlier and wrote an overly loose [compat], which is why 0.3+ got pulled in and caused the break.

logan-nc:
Ok. Thanks for the clean fix and clear explanation @mgyoo86 

Based on the wonderful improvements we've seen from FatInterpolations, I expect we will regularly want to update to the latest minor releases even while it is still in major version `0.*` development. I understand that we may not *need* to keep up with every minor release, so this is ok for now. How do you recommend we decide if/when to update to the newer API? Just wait until you tell us to? Will you keep us in mind for such things?

mgyoo86:
@logan-nc 

Julia package manager is pretty good for handling this version compatibility.

Let me first clarify that, at the pre-1.0 phase, i.e., `v0.x.y`, `x` is the major version, `y` is the minor version.

As we just fixed the version [compat] to "v0.2.y",
this means that Julia package manager will only update the minor version (y), keeping the same major version (2).

This means that you can get newest features the package provides as long as there were no breaking chagnes.
Once the pacakge made the breaking changes, the major version jumps.

In this situation, Julia package manager keep you using the old major version (2), but let you know that there is the new major version!

For example, see the following result when you type `status` or `st` in short in the `pkg>` mode:
image

You can invesitgate what the new version is by typing `status --outdated`, 
image

From this information, you can know what the newest version, and see the documentations from the package and move on to the newest version when you ready :)

PR #183, opened 2026-03-04
Calin/imas integration

### LINKED ISSUE #138: Adding IMAS capabilities to GPEC
In order to integrate the existing code into FUSE, we'll need to add an input capability for IMAS data structures. I had planned to do this a while ago, but got caught up in other tasks so I am creating an issue to document my plan.

From a conversation from Brendan and I:
1. Add a "imas" input type to the equilibrium inputs. This will act in a similar way to other input types, extracting the necessary data from the data structure and passing it into the general equilibrium processing scripts
2. Make this repo depend on IMASdd.jl
3. Initialize DCON from a dd. You could use https://github.com/JuliaFusion/EFIT.jl/blob/master/src/geqdsk_imas.jl as an example. The imas2geqdsk shows how to get the data you need from the dd, and since you can run from a gEQDSK, that should be everything you need.
4. To test, can use EFIT.jl geqdsk2imas to create a dd from an existing test gEQDSK file, so then you could do a one-to-one comparison. Or EFIT.jl has a test_dd.eq.json file you could use to start, as shown in https://github.com/JuliaFusion/EFIT.jl/blob/master/test/runtests.jl
5. After the run, we should write the linear stability data back to the IMAS file

calin1989:
Work in progress IMAS integration


jhalpern30:
Replaced by #202

PR #185, opened 2026-03-06
ForceFreeStates: simplify EulerLagrange integration (unify tolerances, remove callback)

logan-nc:
Replaces #177.

## Summary

- **Unify tolerances**: `tol_r`, `tol_nr`, `crossover` → single `eulerlagrange_tolerance = 1e-7`. Removes `compute_tols()` and the per-step `integrator.opts.reltol` assignment. The 10× difference between `tol_r`/`tol_nr` in examples is not meaningful since BS5 adapts step size near singularities regardless.
- **Remove `DiscreteCallback`**: Replaces `solve()` + `DiscreteCallback` with a manual `init`/`step!` loop in `integrate_el_region!`. Deletes `integrator_callback!()` entirely.
- **Fix `near_start` correctness bug**: The old code used `length(integrator.sol.t)` to count steps-in-segment, but with `save_everystep=false` this was always ≤2, making `near_start` always true and causing every step to be saved. Now tracked with a plain `steps_in_segment` counter.
- **Update all TOML configs**: 8 files updated (examples, benchmarks, test data).

## Test plan

- [ ] `julia --project=. test/runtests.jl` passes
- [ ] DIIID-like example: `et[1] ≈ 1.707`
- [ ] Solovev regression examples match expected outputs


logan-nc:
## Test results ✅

All tests pass on this branch:

| Test suite | Result |
|---|---|
| FastInterpolations-based Splines | ✅ (0 tests, skipped) |
| FourierCoefficients | ✅ 5/5 |
| Vacuum.jl Unit Tests | ✅ 112/112 |
| JPEC.Equilibrium Unit Tests | ✅ 8/8 |
| Solovev unit tests | ✅ 24/24 |
| ODE Tests | ✅ **85/85** (includes new `eulerlagrange_tolerance` field test) |
| Test Sing Functions | ✅ (0 tests, skipped) |
| Full ForceFreeStates runs | ✅ **2/2** (Solovev + multi-n, 22.6s) |

**Note**: A follow-up commit was needed. The first implementation used a manual `init()`/`step!()` loop, which caused a hang because [`BS5` is FSAL](https://docs.sciml.ai/OrdinaryDiffEq/stable/explicit/bs/): after `apply_gaussian_reduction!` modifies `integrator.u` in-place between steps, the cached last-stage derivative becomes stale, causing the adaptive step sizer to repeatedly shrink `dt → 0`.

The fix reverts to `solve()` + `DiscreteCallback` (the proven pattern) and fixes the `near_start` bug using a `Ref` closure counter instead of `length(integrator.sol.t)`, which was the original bug (#177).

logan-nc:
## Additional commit: improved near-boundary save heuristic

**Problem with the old saving logic** (two separate issues):

1. **Inconsistent criteria** — `near_start` used step count (`<= 2 steps`) while `near_end` used a psi fraction (5% of `psi_range`). Apples and oranges.
2. **Wrong coordinate** — solution variation near a rational surface scales with singfac = `n(q − q_res)`, not ψ-distance. With a flat q profile (low magnetic shear), a 5% ψ band is a tiny q interval, potentially under-resolving the approach. With steep q (high shear), the same 5% ψ is a large q interval, saving far more steps than needed.
3. **Endpoint not guaranteed** — the final step was saved only if the near_end heuristic happened to catch it, fragile against large adaptive steps.

**New logic** (`near_q_frac = 0.05`, local variable for easy future exposure):

```julia
q_start = equil.profiles.q_spline(chunk.psi_start)
q_end   = equil.profiles.q_spline(chunk.psi_end)
q_range = abs(q_end - q_start)

near_start = abs(odet.q - q_start) < near_q_frac * q_range || steps_in_segment[] == 1
near_end   = abs(odet.q - q_end)   < near_q_frac * q_range
```

Plus an unconditional endpoint save after `solve()` as an explicit guarantee (belt-and-suspenders; near_end still catches it in normal operation since `odet.q ≈ q_end` at the final step).

logan-nc:
## Benchmark results vs `develop` (DIIID-like example)

| Metric | `develop` @ 6655d6c4 | `callback_cleanup` @ b1435e1c | Δ |
|---|---|---|---|
| `et[1]` (eigenmode energy) | 1.2106 | 1.2106 | **0.00%** ✅ |
| Runtime (warmed avg) | 62.59 s | 59.17 s | **−5.5%** (faster) ✅ |
| Integration steps (total) | 131 | 722 | +451% ⚠️ see note |

### Ideal stability ✅
`et[1]` matches bit-for-bit. The ODE solver produces the same eigenmode.

### Runtime ✅
~5% faster, despite the save logic change. Note `save_interval` was also tuned from 10 → 3 in the DIIID example during this work, which increases saved steps but the runtime improvement comes from eliminating `compute_tols()` and the per-step `integrator.opts.reltol` assignment.

### Step count ⚠️
The 451% increase in total integration steps is unexpected. On `develop`, the near_start bug caused every step to be saved (making `save_interval=10` ineffective), so the saved-step count was high. The total-step count increase from 131 → 722 in the HDF5 output warrants a closer look at what metric the benchmark tool is reading and whether the q-based near-boundary zones are creating unexpectedly dense saves.

### Δ' difference ⚠️
The perturbed equilibrium output differs between branches:
- `develop`: Δ' = **+1.189e+11** at q=2
- `callback_cleanup`: Δ' = **−1.757e+10** at q=2 (sign flip, ~7× smaller magnitude)

This suggests `ResponseMatrices.jl` is sensitive to the density/placement of saved solution points, which changed with the near_start/near_end fix. The Δ' from the ideal DCON integration feeds directly into singular coupling. **This should be investigated before merge** — either by establishing which branch gives the physically correct value, or by confirming the Δ' from `develop` was already wrong due to the near_start bug inflating solution density near the axis.

logan-nc:
I am going to be a medium-bad developer and merge this as soon as it passes the tests even if no one reviews it. Reviewers were mostly tagged just to notify those folks as an FYI. 
I'm on a push to close out PRs and the web of interdependent conflicts is growing exponentially, so I need to knock down the numbers by hitting merge.

matt-pharr:
@logan-nc did you have any thoughts on the diverging Delta' values introduced here?

logan-nc:
@matt-pharr the Δ's being quoted are from PerturbedEquilibrium resonant metics. The PerturbedEquilibrium code is just a skeleton here - actual, correct, plasma response calculations are in #196 to be merged soon (hopefully this week, but I keep getting tempted to add one more output each day 😅 )

PR #186, opened 2026-03-07
VACUUM - BUG FIX - Fix thread workspace allocation for Julia 1.12+

mgyoo86:
## Summary

This is a Julia version-dependent bug: `Threads.nthreads()` and `Threads.threadid()` behave differently starting in Julia `1.12`, causing a BoundsError in a manual thread-local workspace indexing. CI only tested against Julia 1.11, so this bug was not caught. This PR fixes the bug and adds a version matrix (lts, 1.11, 1.x) to the CI workflow to test the code on multiple julia versions, catching this kind of version-dependent bugs earlier.

## Problem

In `Vacuum` module, `compute_3D_kernel_matrices!` creates a manual thread-local workspaces using `Threads.nthreads()` and indexes them with `Threads.threadid()`. This works on Julia 1.11 where both values start at 1, but breaks on Julia 1.12+:

| | Julia 1.11 | Julia 1.12+ |
|---|---|---|
| `maxthreadid()` | 1 | 2 |
| `threadid()` inside `@threads` | 1 | 2 |
| `nthreads()` | 1 | 1 |

On 1.12+, the main thread (ID 1) belongs to the `:interactive` pool and the default pool worker starts at ID 2. This causes `BoundsError` when accessing `workspaces[threadid()]` with an array sized by `nthreads()`.

## Fix

Minimal fix: use `Threads.maxthreadid()` instead of `Threads.nthreads()` for workspace allocation, ensuring the array covers all possible thread IDs.

## Notes for future work

1. **`threadid()`-based indexing is fragile.** Thread IDs are not guaranteed to be contiguous or start at 1 across Julia versions, so directly using `threadid()` for workspace indexing is generally discouraged.
   - `AdaptiveArrayPools` can handle this more robustly. I'll follow up with a separate PR.

2. **BLAS thread contention.** The `@threads` loop calls `mul!` internally. When multiple Julia threads call BLAS routines concurrently, each call may utilize up to BLAS.get_num_threads() threads, leading to significant oversubscription (e.g. (4 Julia threads) x (8 BLAS threads per call) = up to 32 concurrent threads on a 8-core machine), significanlty slowing down the performance. 
 In `FUSE` we set `BLAS.set_num_threads(1)` globally to avoid this problem, but that's not always desirable since it slows down other `mul!` calls outside threaded regions. A scoped alternative:
   ```julia
   old_blas = BLAS.get_num_threads()
   BLAS.set_num_threads(1)
   try
       Threads.@threads :static for idx_obs in 1:num_points
           ...
       end
   finally
       BLAS.set_num_threads(old_blas)
   end
   ```
   This would need performance benchmarking to confirm the tradeoff.

## CI changes

- Added Julia version matrix (`lts`, `1.11`, `1.x`) with `fail-fast: false`
- Added `julia = "1.10"` compat to `Project.toml`
- Broadened stdlib compat bounds from pinned `1.11.0` to `"1"`


Project.toml
```diff
@@ -28,6 +28,7 @@ TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
+julia = "1.10"
```
logan-nc, line 31: Why is this julia version pinned to old `"1.10"`? I've been using 1.11 and the github workflows have been using 1.11. So I would have thought we'd pin ourselves to 1.11 if anything

Project.toml
```diff
@@ -37,15 +38,15 @@ FastGaussQuadrature = "1.1.0"
 FastInterpolations = "0.2.13"
 HDF5 = "0.17.2"
 JLD2 = "0.6.3"
-LinearAlgebra = "1.11.0"
+LinearAlgebra = "1"
```
logan-nc, line 41: The rhyme and reason for which of these packages were rounded to the major version only is not clear to me. Why are we changing the toml at all in this "minimal fix" PR? Is it somehow necessary for the matrixed tests? Why only these and not the others?

logan-nc (commented):
Can @mgyoo86 clarify the toml changes before we merge?

Project.toml
mgyoo86: TL;DR
It is safe to use "1.10" here and it is a common choice for many pacakges. 
Let me explain more details.

Previously, there was no julia compat entry in JPEC's Project.toml, so any julia version was allowed to use with JPEC.
This is why I encountered this version-dependent bug silently since I'm using julia "1.12" not "1.11".

Here, `julia = "1.10"` means that it allows any julia version within `[1.10.0, 2.0.0)` range, so julia "1.10", "1.11". and "1.12", and future releases (before the new major version starting with "2")  will be available.

The common purpose of the [compat] section is to set the lowest version of pakages that are compatible with JPEC, such that when JPEC is combined with other pacakges at the downstream, the julia can safely resolve the highest-version of packages that are compatible with all given pakages.

If you want to "pin" to julia "1.11" only, you can use `julia="~1.11"`, or more specifially, `julia="=1.11.3"`, for example. You can find more details about the compat rules from [here](https://pkgdocs.julialang.org/v1/compatibility/#Caret-specifiers).

The reason I set the minimum julia version to "1.10" is that, it is the current LTS (long-term support) version which is the promise that the julia community will support this "1.10" version for a longer period (like several years) than others (like "1.11").
Some people, who prefer the stabilitiy & same experience than enjoying newest features, may use this LTS version, instead of the latest one.
For this reason, this `julia="1.10"` is a common choice when developing package to make the pacakge availabe for wider audience, unless there is techincal problem or compatibility issue.

As the current JPEC can run on julia "1.10" well, so I think it is safe to keep it for now.
If there will be any breaking changes that JPEC needs specific new features supported from julia >= 1.11, then we can update the compat to `julia="1.11"`.

Project.toml
mgyoo86: This is just for the convinence. Although the standard libraries (like the LinearAlgebra or Printf) are shipped with julia, their concrete version numbers does not necessary match the same version with julia. 

Thanks to the julia package resolver, we can simply set the major version of standard libraires (which is "1" for now), and then set the julia [compat] (like `julia="1.10"`) then the julia pakcage manager automatically resolve the highest version available.

logan-nc:
Thanks for explaining things to the julia newbies @mgyoo86

PR #187, opened 2026-03-08
EQUIL - NEW FEATURE - Add efit_arclength and efit_by_inversion equilibrium solver strategies

logan-nc:
## Summary

Implements two new equilibrium solver strategies alongside the existing `\"efit\"` geometric-angle field-line ODE, selectable via `eq_type` in `gpec.toml`. All three paths are fully parallel and the original `\"efit\"` path is unchanged.

### Strategy A: `\"efit_arclength\"` (`DirectEquilibriumArcLength.jl`)
- Replaces the geometric-angle ODE with an arc-length parameterization: `(dR/ds, dZ/ds) = (∂ψ/∂Z, −∂ψ/∂R) / |∇ψ|`
- The flux surface is followed by construction — eliminates the denominator singularity `Bz·cos(η) − Br·sin(η) → 0` near elongated plasma tops/bottoms
- No `direct_refine` Newton correction needed at every step
- ODE terminated by `ContinuousCallback` on `Z = zo` upward crossing; output resampled to the same 5-column `y_out` format as `direct_fieldline_int` via angle unwrapping

### Strategy B: `\"efit_by_inversion\"` (`DirectEquilibriumByInversion.jl`)
- Traces flux surface contours using `Contour.jl` (pure-Julia marching squares)
- Selects the plasma-interior closed curve via ray-casting (`_point_in_polygon`); resamples to a uniform-θ grid via cubic spline on angle-sorted vertices
- Near-axis fallback for `psifac < 5e-3`: circular approximation `R ≈ ro + √psifac·(rs2−ro)·cos(θ)`
- Constructs an `InverseRunInput` and delegates to the existing `equilibrium_solver(::InverseRunInput)` (same path used by CHEASE)

### Benchmarking scripts (`benchmarks/`)
| Script | Purpose |
|--------|---------|
| `equil_method_comparison.jl` | Runtime and accuracy at default parameters; q cross-method comparison; roundtrip error; q monotonicity |
| `equil_psihigh_scan.jl` | Robustness as `psihigh → 1` (near separatrix); last successful psihigh per method |
| `equil_psilow_scan.jl` | Near-axis accuracy; q0 extrapolation; roundtrip error for ψ < 0.10 |
| `equil_numerical_params.jl` | mpsi/mtheta sweeps (all methods); Contour.jl `refine` sweep (strategy B) |

**Note on accuracy metrics**: EFIT q-profile is only used as reference for ψ < 0.95. In the far edge (ψ > 0.95), methods are compared against each other using self-consistency metrics (roundtrip error, q monotonicity, Jacobian consistency) — because in a diverted tokamak `q → ∞` as `ψ → 1`, and EFIT polynomial extrapolation gives a wrong finite `qa`.

### New dependency
- `Contour.jl` (`uuid d38c429a-6771-53c6-b99e-75d170b6e991`, v0.6.3) added to `Project.toml`

## Status
Implementation is complete but **not yet run/tested**. Before merging:
- [ ] Run `equil_method_comparison.jl` on `DIIID-like_ideal_example` to verify all three `eq_type` options load and produce sensible results
- [ ] Run `equil_psihigh_scan.jl` and `equil_psilow_scan.jl` to evaluate robustness
- [ ] Fix any issues surfaced by the benchmark runs
- [ ] **Remove `docs/direct_equilibrium_improvements_plan.md`** (session-continuity document only, not for production)

## Test plan
- [ ] `julia --project=. benchmarks/equil_method_comparison.jl examples/DIIID-like_ideal_example`
- [ ] `julia --project=. benchmarks/equil_psihigh_scan.jl examples/DIIID-like_ideal_example`
- [ ] `julia --project=. benchmarks/equil_psilow_scan.jl examples/DIIID-like_ideal_example`
- [ ] `julia --project=. benchmarks/equil_numerical_params.jl examples/DIIID-like_ideal_example`
- [ ] Existing test suite: `julia --project=. test/runtests.jl` (original `\"efit\"` path must be unaffected)


logan-nc:
## Benchmark Results (after bug fixes)

All three methods now verified against the DIIID-like example.

### Method comparison (default params: mpsi=128, mtheta=256, psilow=1e-4, psihigh=0.993)

| Method | Roundtrip max | vs EFIT q (ψ<0.95) max | Runtime (warmed) |
|--------|--------------|----------------------|-----------------|
| efit | 3.15e-04 | 4.63e-04 | 88ms |
| efit_arclength | 2.51e-04 | 6.02e-04 | 93ms |
| efit_by_inversion | **4.19e-06** (75× better) | 3.22e-03 | 673ms |

Cross-method q agreement (full domain): |Δq| < 6e-03 for all pairs.
All three methods succeed up to **psihigh=0.999** with zero q-monotonicity violations in the outer 10%.

### Numerical parameter sensitivity (efit_by_inversion)

| Contour refine factor | Roundtrip max | Runtime |
|----------------------|--------------|---------|
| 2× (default EFIT nw) | 1.98e-05 | 190ms |
| 4× (default) | 6.16e-06 | 673ms |
| 8× | 1.17e-06 | 2.95s |

Roundtrip error is **independent of mpsi and mtheta** (dominated by Contour.jl grid resolution), scaling only with the refine factor.

### Near-axis psilow scan

`efit_by_inversion` q0 extrapolation vs psilow:
- psilow ≤ 0.01: q0 ≈ 1.20-1.23 (matches efit's ~1.21) ✓
- psilow > 0.05: q0 ≈ 11 ✗ — InverseEquilibrium's inner-boundary extrapolation is unreliable when the grid doesn't extend close to the axis. **Recommendation: use psilow < 0.01 with efit_by_inversion.**

### Bugs fixed in this follow-up commit (297d489b)

1. **Theta convention**: `rz_in_ys` must be in turns [0,1] not radians [0,2π]
2. **Near-axis q accuracy**: removed eager circular fallback — now uses Contour.jl for all surfaces (fallback only if truly unresolvable)  
3. **NaN crash**: removed duplicate closing vertex from Contour.jl before building cubic spline
4. **Roundtrip formula**: `psi_norm = 1 - psi_in/psio` (was inverted)
5. **EFIT q parsing**: replaced manual EFIT file re-parsing with `sq_in` spline (column 3 = q)

logan-nc:
## Updated benchmark results (post InverseEquilibrium optimizations)

Latest commit optimizes the `InverseEquilibrium` hot path (items 1–8: eliminate redundant `sq_in` calls, replace per-knot spline evaluation loops with direct `.y` nodal data access, in-place buffer reuse, broadcast fusion, monotone hints). Benchmarked on DIIID-like example, 8 threads, median of 3 warm runs:

| Method | Runtime (8 threads) | Roundtrip max error |
|--------|---------------------|---------------------|
| `efit` | 0.086 s | 3.15e-04 |
| `efit_arclength` | 0.093 s | 2.51e-04 |
| `efit_by_inversion` | **0.188 s** | **4.19e-06** |

`efit_by_inversion` is now ~0.188 s on 8 threads (down from ~0.336 s before this commit, ~44% faster), while retaining 75× better roundtrip accuracy than `efit`.

logan-nc:
## Benchmark scan results (DIIID-like example, 8 threads)

### psihigh robustness scan (`equil_psihigh_scan.jl`)

All three methods succeed up to **psihigh = 0.999** with zero q monotonicity violations throughout.

| psihigh | efit rt_edge | arclength rt_edge | by_inversion rt_edge | q(end) |
|---------|-------------|-------------------|----------------------|--------|
| 0.980 | 2.03e-04 | 2.52e-04 | 3.42e-06 | ~4.91 |
| 0.990 | 2.28e-04 | 2.67e-04 | 3.84e-06 | ~5.26 |
| 0.995 | 1.75e-04 | 2.00e-04 | 3.04e-06 | ~5.58 |
| 0.999 | 1.87e-04 | 3.10e-04 | 4.12e-06 | ~6.25 |

`efit_by_inversion` roundtrip error stays 50–100× better than the ODE methods all the way to the separatrix. q diverges as expected (physically correct for a diverted tokamak).

---

### psilow near-axis accuracy scan (`equil_psilow_scan.jl`)

| psilow | efit q0 | arclength q0 | by_inversion q0 | by_inversion rt_inner |
|--------|---------|--------------|-----------------|----------------------|
| 1e-01 | 1.199 | 1.199 | **11.393** ❌ | 1.35e-04 |
| 5e-02 | 1.206 | 1.207 | **11.745** ❌ | 5.84e-05 |
| 1e-02 | 1.209 | 1.209 | 1.225 ✓ | 7.14e-06 |
| 1e-03 | 1.210 | 1.210 | 1.230 ✓ | 6.00e-06 |
| 1e-04 | 1.210 | 1.210 | 1.205 ✓ | 5.33e-06 |

**`efit_by_inversion` requires psilow ≤ 0.01.** For psilow ≥ 0.05, the innermost grid surfaces are too far from the axis for InverseEquilibrium's boundary extrapolation, causing a spurious q0 ~ 11. For psilow ≤ 0.01, q0 extrapolation agrees with the other methods (~1.21–1.23) and inner roundtrip error is better (5–7e-06 vs 3–6e-05).

---

### Numerical parameter sensitivity (`equil_numerical_params.jl`)

**mpsi and mtheta sweeps**: roundtrip error is insensitive to both for `efit_by_inversion` (stays at 6.16e-06 across all tested values) — accuracy is set by the Contour.jl grid, not by the SFL grid resolution.

**Contour grid refinement sweep (efit_by_inversion):**

| refine | runtime | roundtrip max |
|--------|---------|---------------|
| 2 | 0.09 s | 1.98e-05 |
| **4 (default)** | **0.18 s** | **6.16e-06** |
| 8 | 0.88 s | 1.17e-06 |

Default refine=4 is the sweet spot: 3× better than refine=2 for 2× cost. refine=8 costs 5× more for only another 5× gain.

logan-nc:
## Grid clipping optimization

Clips the fine Contour.jl grid to the plasma bounding box (separatrix + 5% margin) instead of the full EFIT domain. The bounding box is free — obtained by tracing the outermost needed flux surface on the already-stored EFIT coarse nodal data with a single Contour.jl call, no extra spline work.

For DIIID-like (where the plasma fills ~68% of the EFIT domain area):
- Grid: 1028×1028 → **877×798** (~32% fewer points)
- Runtime: 0.188s → **0.096s** (~49% faster, 8 threads)
- Roundtrip accuracy: unchanged (4.6e-06)

`efit_by_inversion` is now only ~2× slower than `efit` (0.096s vs 0.047s) while retaining 70× better roundtrip accuracy. The fallback (full EFIT domain) is used automatically if the coarse contour trace fails.

logan-nc:
## Final benchmark summary (all optimizations combined)

**Runtime on DIIID-like example, 8 threads, median of warm runs:**

| Method | Runtime | Roundtrip max error |
|--------|---------|---------------------|
| `efit` | 0.047 s | 3.15e-04 |
| `efit_arclength` | 0.056 s | 2.51e-04 |
| `efit_by_inversion` | **0.096 s** | **4.58e-06** |

`efit_by_inversion` is now ~2× slower than `efit` while retaining ~70× better roundtrip accuracy.

**Optimization history for `efit_by_inversion`:**

| Stage | Runtime (8 threads) |
|-------|---------------------|
| Initial implementation | ~0.34s |
| + parallelism, hints, mutating resample | ~0.19s |
| + InverseEquilibrium hot path (items 1–8) | ~0.19s |
| + separatrix bounding-box grid clipping | **0.096s** |

The grid clipping (clips fine grid to separatrix bbox + 5% margin using free coarse nodal data) was the largest single gain: 32% fewer grid points → ~49% faster.

logan-nc:
## psihigh scan with et[1] (free-boundary stability eigenvalue)

Ran `benchmarks/equil_psihigh_scan.jl` on 8 threads (DIIID-like example, mpsi=128, mtheta=256, refine=4).

**Bug fixed this session:** `InverseEquilibrium.jl` had a pre-existing bug in the `eqfun` computation (magnetic field and metric coefficients). Two errors:
1. `rzphi_fs` was populated with pre-SFL-transformation spline nodal values (`spl.y[:, 1:4]`). The `eqfun` computation needs post-SFL values (deta with SFL correction, ν function, SFL Jacobian). Fixed by rebuilding the spline from the modified `spl_fs` at the SFL theta positions and evaluating at uniform theta.
2. `jacfac` in the eqfun loop used `fx_rzphi[4]` (the ψ-**derivative** of `rzphi_jac`) instead of `f_rzphi[4]` (the Jacobian **value**). This caused B≈72 T instead of B≈1.66 T, and et[1]≈−2.4M.

After the fix, `efit_by_inversion` B field matches `efit` to 5 significant figures at representative test points.

---

### et[1] vs psihigh (all methods succeed through psihigh = 0.9999)

| psihigh | `efit` | `efit_arclength` | `efit_by_inversion` |
|---------|--------|------------------|---------------------|
| 0.980  | +2.2016 | +2.2005 | -150.96 |
| 0.985  | +2.1559 | +2.1564 | +2.1549 |
| 0.990  | +1.2430 | +1.2161 | +1.0997 |
| 0.993  | +1.2106 | +1.2417 | -44.34  |
| 0.995  | +1.1845 | +1.2102 | +1.2554 |
| 0.996  | +1.1679 | +1.1443 | +0.5062 |
| 0.997  | +1.1451 | +1.1139 | +1.1303 |
| 0.998  | +1.1182 | +1.0423 | +1.2692 |
| 0.999  | +0.7232 | -1689   | -498    |
| 0.9995 | -59.25  | +2.1462 | -3221   |
| 0.9999 | -591684 | -18.3M  | -14419  |

**Key observations:**
- `efit` and `efit_arclength` give consistent, stable et[1] through psihigh ≈ 0.998 (range +1.0 to +2.2), indicating robustly stable equilibrium
- Near-separatrix breakdown at psihigh ≥ 0.999 is expected: q→∞ causes the EFIT q-profile extrapolation to become unphysical, so rational surfaces proliferate
- `efit_by_inversion` shows occasional noisy et[1] values (e.g., −150 at psihigh=0.980, −44 at 0.993). These likely reflect sensitivity of the FFS eigenvalue to the specific q-profile near a rational surface, since the inverse solver produces a slightly different equilibrium geometry. The physical et[1] values (~1.0–1.3) match the ODE methods where the result is clean.
- Zero q-monotonicity violations at the edge for all methods at all psihigh values ✓

**Conclusion:** All three methods successfully reconstruct the equilibrium and compute stable eigenvalues through psihigh=0.998. The methods agree on the physics (positive et[1], decreasing as psihigh → separatrix) for well-conditioned psihigh values.

logan-nc:
## Progress update: sinh-stretched Z grid, InverseEquilibrium bug fix, and psihigh scan results

### Commits pushed in this update

- **`98de4d06`** — Benchmark scripts: fix broken symlink in psihigh scan (`abspath` required for symlink targets); update spline comparison to use 8× oversampled LDP-distributed eval grid (previous 500-pt uniform grid was 13× too coarse to capture inter-knot ringing)
- **`ab03a1cf`** — **BUG FIX**: Remove spurious near-axis Lagrange extrapolation in `InverseEquilibrium.jl`

The earlier commits on this branch (from the previous session) implement:
- **Sinh-stretched Z grid** in `DirectEquilibriumByInversion.jl`: concentrates Contour.jl marching-squares resolution near x-point(s) without grid stitching. `Z(v) = z_lo + ΔZ·sinh(β·v)/sinh(β)` with `β_z=2.0` default.
- **Topology classifier** (`classify_topology`): detects `:limited`, `:sn_lower`, `:sn_upper`, `:double_null` to know which end(s) of Z to stretch toward.
- **Round-trip validation**: warns if any test point has `|ψ_reconstructed − ψ_target| > 2e-3`.

---

### Bug: spurious near-axis Lagrange extrapolation (now fixed)

`InverseEquilibrium.jl` had a Fortran-ported line that overwrote `deta[1,:]` (the angle offset at the innermost flux surface) with a Lagrange-3 polynomial extrapolated to `r²=0` (the magnetic axis). In the Fortran `inverse.f` this was correct because the innermost surface *was* the axis. In Julia, `rz_in_xs[1] = psilow > 0`, so the line was overwriting valid data with a physically inconsistent axis-extrapolated value — causing a cubic spline spike near `psilow` visible in the Jacobian, angle offset, and eqfun plots. Removed entirely; the comment block explains the reasoning.

The fix also removed the spike visible in `rzphi_spline_4_Jacobian.png` and `rzphi_spline_2_angle_offset.png` for the `efit_by_inversion` method.

---

### psihigh scan results (DIIID-like example, after all fixes)

Round-trip errors for `efit_by_inversion` are now uniformly **~2×10⁻⁵**, roughly **6–10× better** than `efit`/`efit_arclength` (~2×10⁻⁴). Zero q-monotonicity violations across all psihigh and all methods.

**et[1] (free-boundary stability eigenvalue) — positive = stable:**

| psihigh | efit | efit_arclength | efit_by_inversion |
|---------|------|----------------|-------------------|
| 0.980   | +2.20 | +2.20 | **+2.36** |
| 0.985   | +2.16 | +2.16 | **−35.8** ❌ |
| 0.990   | +1.24 | +1.22 | **+1.37** |
| 0.993   | +1.21 | +1.24 | **+1.52** |
| 0.995   | +1.18 | +1.21 | **+1.23** |
| 0.996   | +1.17 | +1.14 | **−6323** ❌ |
| 0.997   | +1.15 | +1.11 | **+1.27** |
| 0.998   | +1.12 | +1.04 | **+1.24** |
| 0.999   | +0.72 | −292 ❌ | **−5712** ❌ |
| 0.9995  | −59.2 ❌ | −6444 ❌ | **−161** ❌ |
| 0.9999  | −5.9×10⁵ ❌ | −3.1×10⁶ ❌ | **−6235** ❌ |

Key observations:
- `efit` is stable and consistent from 0.980–0.999; diverges at 0.9995+ (near-separatrix physics expected)
- `efit_arclength` is similar to `efit` through 0.998, then diverges at 0.999+
- `efit_by_inversion`: round-trip accuracy is excellent at all psihigh, but **et[1] is still sporadic** at psihigh = 0.985, 0.996, 0.999, 0.9995, 0.9999. The pattern is non-monotonic (0.997 and 0.998 are fine, flanking the bad 0.996 and 0.999 values).

The root cause of the remaining `efit_by_inversion` instability is **unknown**. The contour tracing is accurate (round-trip ~2e-5), so the issue lies downstream — possibly in how InverseEquilibrium constructs the SFL coordinates for specific psihigh values, or in sensitivity of the ODE integrator to subtle spline discontinuities at the edge.

---

### Spline comparison at psihigh=0.999 (figures in `benchmarks/equil_spline_comparison_psihigh0p999/`)

After the InverseEquilibrium fix, the deep-core spike in Jacobian and angle offset is **gone**. Remaining differences:

- **rzphi splines** (r², angle offset, toroidal shift, Jacobian): `efit_by_inversion` and `efit` agree well across most of ψ; small differences in outermost ~5% due to slightly different contour tracing near the x-point arms.
- **metric1**: Still shows O(1)–O(10) discrepancies between methods at off-axis θ. This is confirmed to be a **permanent feature of two independent SFL constructions** (not a bug) — even at psihigh=0.997 (which gives correct et[1]), metric1 shows the same level of disagreement. Not a diagnostic for the et[1] instability.
- **1D profiles** (F, P, dV/dψ, q): Essentially identical between all three methods — confirms the 1D physics is consistent.
- **Flux contours**: Visually identical between efit and efit_by_inversion throughout the plasma.

---

### Remaining TODOs for next session

1. **Root cause of remaining efit_by_inversion et[1] instability** — The sporadic bad et[1] values (at 0.985, 0.996, 0.999) occur despite excellent round-trip accuracy. Possible investigations:
   - Print/plot the `deta` angle offset array at bad vs. good psihigh values — look for large inter-surface jumps
   - Check if the outermost Contour.jl contour at bad psihigh values has anomalous shape (arms that nearly touch, self-intersecting contours, etc.)
   - Try increasing `nz` or `nR` in `make_stretched_z_grid` for the fine grid
   - Look at whether the spline of `deta` vs. ψ has a large jump near psihigh (would indicate the outermost contour is qualitatively different from its neighbors)

2. **psihigh=0.980 et[1] = +2.36 for efit_by_inversion vs. +2.20 for efit** — slightly elevated but positive; worth monitoring

3. **Regression testing** — InverseEquilibrium extrapolation removal should be verified on CHEASE and Solovev paths (these also call `InverseEquilibrium`)

4. **`inverse_extrap` function** is now unused — consider removing in a cleanup pass

5. **β_z tuning** — Currently hardcoded to 2.0. Consider exposing in `gpec.toml` under `[Equilibrium]` if users need to tune it for different topologies or equilibria with extreme x-point positions.

6. **near-separatrix instability** — All three methods show et[1] < 0 at psihigh ≥ 0.9995. This may be physical (near-separatrix regime where q→∞ and the code's finite-domain approximations break down) or a numerical issue — worth understanding.

logan-nc:
@claude review this PR. look for bugs that might explain the sporadic efit by inversion results. Based on the core interpolation issue we found, it seems the inverse equilibria from chease and other codes may have assumed surfaces were provided for psin [0,1] instead of [psilow, psihigh]. if this is the case we need a different behavior for our contour.jl path -either key word arguments telling the inverse solver the psi range of the contours or perhaps an entirely new solver. give your recommendation

logan-nc:
## Session status summary — picking up from here

### What was accomplished this session

**Goal**: Fix near-axis ringing in `efit_by_inversion` by improving the global grid construction.

**Key change committed**: `EQUIL - IMPROVEMENT - Unify R/Z grid construction via make_multi_stretched_grid` (commit `8d37eeda`)

The core refactor replaces two separate grid functions with a single `make_multi_stretched_grid` that concentrates grid points near an arbitrary list of points (interior or boundary). Both `make_stretched_r_grid` and `make_stretched_z_grid` are now thin wrappers. The Z grid now concentrates at **both** the magnetic axis (`zo`) and x-point boundaries, which was the user's stated requirement ("both the R and Z grids need to be smoothly concentrated at 2 x-points and the axis").

**Architecture of new function**:
- `make_multi_stretched_grid(lo, hi, centers, n, β)`: splits domain at each interior concentration point; each sub-interval uses `_sinh_both_ends` (fine at both ends) or `_sinh_left`/`_sinh_right` (one-sided toward the concentration point)
- `make_stretched_r_grid(r_lo, r_hi, ro, n, β)` → `make_multi_stretched_grid(..., [ro], ...)` — unchanged API, same result
- `make_stretched_z_grid(z_lo, z_hi, zo, n, topology, β)` → `make_multi_stretched_grid(..., [zo, z_lo?, z_hi?], ...)` — **new `zo` parameter**; all callers updated

**Result**: The old merge hack (sort+vcat+dedup of two full-size grids) is eliminated. Global grid is now 1099×1005 (down from 1099×2009) with `dZ=1.23mm` at the axis — better efficiency AND better axis resolution. Round-trip edge error: `2.6e-6`. No zoom pass needed at `psilow=1e-3`.

---

### Still-open work items

The user's original concern about transition ringing was driven by earlier iterations (zoom_extend_factor, adaptive condensation). With the unified grid, the near-axis resolution is now naturally good from the global grid. However **the plot images from the benchmark haven't been manually inspected** — recommend running the benchmark and visually checking:

```bash
julia --project=. -t4 benchmarks/equil_spline_comparison.jl examples/DIIID-like_ideal_example 0.996
```

Plots go to `benchmarks/equil_spline_comparison_psihigh0p996/`. Look specifically at:
- `flux_contours_combined.png` — grid overplot should show fine Z spacing near BOTH x-point AND midplane
- `1d_profile_4_q*.png` — q profile should be smooth in the core (no ringing near psilow)

The `metric1` edge error (`~2`) is a known pre-existing issue with `efit_by_inversion` at off-axis angles near the separatrix — not new to this session.

### Branch state
- Branch: `direct_equilibrium_improvements` 
- Latest commit: `8d37eeda`
- No outstanding merge conflicts; CI should be green

logan-nc:
## psihigh scan results (post `make_multi_stretched_grid`, commit `8d37eeda`)

All runs: DIIID-like example, 4 threads, warmed.

---

### et[1], q(psihigh), and runtime vs psihigh

| psihigh | q efit | q arc | q inv | et[1] efit | et[1] arclength | et[1] by_inv | t efit | t arc | t inv |
|---------|--------|-------|-------|-----------|-----------------|--------------|--------|-------|-------|
| 0.980 | 4.91 | 4.92 | 4.92 | +2.202 | +2.203 | **+2.200** ✓ | 0.04s | 0.05s | 0.83s† |
| 0.985 | 5.06 | 5.06 | 5.06 | +2.059 | +2.157 | **+2.156** ✓ | 0.07s | 0.05s | 0.33s |
| 0.990 | 5.26 | 5.26 | 5.26 | +1.371 | +1.223 | **+1.239** ✓ | 0.11s | 0.05s | 0.29s |
| 0.993 | 5.43 | 5.42 | 5.43 | +1.123 | +1.230 | **+1.198** ✓ | 0.12s | 0.06s | 0.33s |
| 0.995 | 5.58 | 5.58 | 5.58 | +1.118 | +1.223 | **+1.179** ✓ | 0.11s | 0.06s | 0.33s |
| 0.996 | 5.68 | 5.68 | 5.68 | +1.292 | +0.995 | **+1.150** ✓ | 0.08s | 0.06s | 0.32s |
| 0.997 | 5.80 | 5.80 | 5.80 | +1.167 | +1.085 | **+1.140** ✓ | 0.04s | 0.05s | 0.33s |
| 0.998 | 5.97 | 5.98 | 5.97 | +1.219 | +1.164 | **+1.122** ✓ | 0.05s | 0.05s | 0.29s |
| 0.999 | 6.24 | 6.26 | 6.24 | +0.736 | +3.017 | **+1.049** ✓ | 0.04s | 0.07s | 0.25s |
| 0.9995 | 6.52 | 6.54 | 6.51 | −71 ❌ | +4.70 | **−10623** ❌ | 0.05s | 0.10s | 0.23s |
| 0.9999 | 7.14 | 7.24 | 7.09 | −5.7e5 ❌ | −2.0e6 ❌ | **−6768** ❌ | 0.09s | 0.08s | 0.21s |

†First `efit_by_inversion` run includes residual JIT cost; steady-state is ~0.25–0.33s.

---

### Key observations

**The sporadic `efit_by_inversion` et[1] instability is resolved.** The `make_multi_stretched_grid` refactor (commit `8d37eeda`) fixes the Z grid concentration at both the magnetic axis and x-points, which was causing poor Contour.jl contour quality at certain psihigh values. Previously bad cases are now clean:

| psihigh | Previous et[1] | Current et[1] |
|---------|---------------|---------------|
| 0.980 | −150.96 ❌ | +2.200 ✓ |
| 0.985 | −35.8 ❌ | +2.156 ✓ |
| 0.996 | −6323 ❌ | +1.150 ✓ |

`efit_by_inversion` now gives **consistently stable, physically reasonable et[1] through psihigh=0.999**, in good agreement with the other methods.

**Near-separatrix breakdown** (psihigh ≥ 0.9995) affects all methods — this is physical (q→∞ causes rational surface proliferation) and not a bug. `efit_by_inversion` breaks down sooner than `efit_arclength` at this extreme edge, but all three methods fail within one step of each other.

**Roundtrip accuracy** stays 10–20× better for `efit_by_inversion` (1.5–3.5e-5 vs ~2e-4) through all psihigh values including 0.9999.

**Runtime**: `efit_by_inversion` runs ~0.25–0.33s warmed (~6× slower than `efit`). Interestingly, runtime *decreases* toward psihigh=0.9999 (0.33s → 0.21s) because the separatrix bounding-box grid clipping gets more aggressive as the traced domain shrinks, processing fewer Contour.jl grid points.

---

### psihigh=0.9999 spline comparison

Also ran `equil_spline_comparison.jl` at the extreme psihigh=0.9999:

- **Flux contours**: all three methods overlay perfectly — geometry is clean at this extreme psihigh
- **q-profile**: bulk (ψ < 0.99) is identical across all methods; final-surface q diverges (7.09–7.24) as physically expected for a diverted tokamak  
- **Jacobian / dV/dψ**: clean throughout the bulk; single-surface divergence at the boundary is physical (dV/dψ → ∞ near separatrix)
- **|B|**: methods agree to < 6e-3 T at edge, < 5e-5 T in core
- **metric1**: large O(10–50) discrepancy at off-axis θ near psihigh — this is the pre-existing known issue from two independent SFL constructions; not a regression and not diagnostic of et[1] accuracy (methods agree on et[1] at nearby psihigh values where metric1 shows the same discrepancy)

Figures saved locally in `benchmarks/equil_spline_comparison_psihigh0p9999/`.

logan-nc:
Here are the benchmark plots showing impressive agreement for the 3 techniques all the way out to `psihigh=0.99995`! Cool! Note how far I had to zoom in to see the differences. 

flux_contours

profile_q
profile_dVdpsi

eqfun_metric2
eqfun_metric1
eqfun_Bmag
rzphi_jacobian
rzphi_nu
rzphi_angle_offset
rzphi_rsquared

logan-nc:
If we push out to 0.99999 the arc length approach fails with x point issues, but the contour method produces smoother splines in the far edge than the original one 👀 

profile_q
flux_contours

logan-nc:
With six 9's arc length comes back? Both are smoother than the original:
profile_q

logan-nc:
@StuartBenjamin might be interested in this :)

logan-nc:
I think this is ready for review. My hope with the work was to find a new construction method that was more robust out to high psihigh values very close to one, and thus high q values. It looks like each method has pros and cons. There are definitely some good aspects of the new ones, so they are worth merging in as options. No one is a clear silver bullet, but maybe if they stay around a bit someone will make a magic tweak that turns it into one 🤞

jhalpern30:
@logan-nc this is definitely a large and very cool PR - I can probably get to a review in the next few days. Two things that I'm going to request at first glance:
1. The documentation fail will need to be fixed before we merge
2. A lot of additional functionality seems to be added to the equilibrium code but the only unit tests I see added are the InverseEquilibrium ones for CHEASE and LAR. There should be some new ones for your new efit methods too

logan-nc:
## Status update — adaptive β_z + grid sensitivity benchmark

### `adaptive_xpt_bz` (new function in `DirectEquilibriumByInversion.jl`)

Physics-based β_z computation for `make_stretched_z_grid`, replacing the hard-coded default of `2.0`:

1. Traces the LCFS at psihigh using the coarse nodal ψ grid (cheap, no extra spline work)
2. Finds the x-point vertex (min-Z for lower x-pt, max-Z for upper, both for double-null)
3. Evaluates ψ_ZZ at the x-point to estimate the flux-surface arm width: `δZ = sqrt(2*(1-psihigh)*psio / |ψ_ZZ|)`
4. Solves β/sinh(β) = n_min_xpt·δZ / (L·n_seg) via 60-iteration bisection, where L = half z-span and n_seg ≈ nz_fine
5. Returns 0 for `:limited` topology or when a uniform grid already suffices; clamps to [0, 8]

This adapts naturally to different tokamak sizes (δZ scales with machine), EFIT grid densities (nz_fine scales with nh×refine), psihigh proximity (1-psihigh factor), and x-point sharpness (ψ_ZZ captures local curvature).

### `benchmarks/equil_inv_grid_sensitivity.jl` (new benchmark)

Convergence study sweeping:
- **Sweep A**: `refine` ∈ {2, 3, 4, 5, 6, 8} at β_z = 2.0
- **Sweep B**: `β_z` ∈ {0, 0.5, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0} at refine = 5

Metrics (ψ > 0.80 edge region):
- max |Δq|, |ΔdV/dψ| vs both efit and high-res inversion (refine=8) as references
- max roundtrip error (ψ,θ)→(R,Z)→ψ
- max |Δr²|, |Δoffset|, |ΔJac| for 2D splines at θ = 0, 0.25, 0.5, 0.75

Outputs: `conv_refine.png`, `conv_bz.png`, `pareto.png`, `surfaces_xpt.png` + 2 CSVs.

### Still to do
- Run the benchmark to see actual numbers and validate that `adaptive_xpt_bz` gives sensible β_z for this DIII-D case
- Wire `adaptive_xpt_bz` into `equilibrium_solver_by_inversion` as the default β_z (replacing hard-coded `2.0`)
- Test on a case with a larger/smaller machine or more/less extreme x-point

logan-nc:
## Dynamic Cartesian grid sizing for `efit_by_inversion`

This update adds physics-based adaptive grid sizing to the `efit_by_inversion` solver, replacing the previous fixed `refine` integer parameter. The goal is to make the grid automatically appropriate for whatever tokamak size, shape, and EFIT grid spacing it encounters, rather than requiring manual tuning.

### How it works

The grid is now derived from two physics constraints rather than a fixed count:

1. **LCFS curvature constraint** — samples the bilinear interpolation error formula along the last closed flux surface: `dR_req = √(8Δψ/|ψ_RR|)`, where `Δψ` is one radial flux grid cell width. This ensures the Cartesian grid can resolve ψ to within one flux surface spacing everywhere on the LCFS, including near the x-point where curvature is highest.

2. **Axis constraint** — requires the global grid cell to be ≤ 20% of the innermost flux surface semi-axis `a_low`, ensuring adequate resolution near the magnetic axis regardless of plasma size.

From these constraints, `h_min` is derived, and a sinh-stretched grid is generated with β chosen from the ratio `h_max/h_min` (capped at β=3 to prevent over-concentration). The `resolution_factor` parameter then scales the resulting grid count uniformly — default `resolution_factor=4.0` was chosen to match the accuracy of the previous `refine=5` default.

### Benchmark results (DIII-D-like H-mode example)

Comparing across all psihigh values from 0.980 → 1.0 (where psihigh=1.0 triggers automatic last-closed-surface detection):

| Method | Last OK psihigh | Roundtrip error | q mono violations | Runtime |
|---|---|---|---|---|
| `efit` | 1.0 | ~3e-5 | 0 | ~0.08s |
| `efit_arclength` | 1.0 | ~5e-5 | 0 | ~0.09s |
| `efit_by_inversion` (new) | **1.0** | ~5e-5 | **0** | ~0.3s |

The new adaptive grid (`~492×784` for DIII-D) is comparable in size and accuracy to the old `refine=5` grid (`~600×900`), while scaling automatically with equilibrium geometry. The stability eigenvalue `et[1]` from `efit_by_inversion` now tracks `efit` closely across the full psihigh range.

The `resolution_factor` kwarg is exposed for users who want to trade accuracy for speed (e.g. `resolution_factor=1.0` for fast scans), and the legacy `refine` kwarg is preserved for backward compatibility.

logan-nc:
## Benchmark update — post-review fixes

Two fixes since the last benchmark run:
1. **Bug fix** (`c2d23c93`): Z zoom grid was using `β_r` parameters instead of `β_z` — copy-paste error. Affects accuracy when `β_r ≠ β_z` (i.e., when the legacy `β_z` override is specified, or in equilibria where the curvature ratio differs between directions).
2. **Benchmark fix** (`eb981bbf`): Sweep C `resolution_factor` range updated to `[0.5, 1.0, 2.0, 3.0, 4.0, 6.0]` with default marker at `rf=4.0` (was incorrectly showing `rf=1.0` as default after the default changed to 4.0 in `365fda33`).

---

## Method comparison (default psihigh=0.997)

| Method | Runtime | Roundtrip error | q mono violations (edge) |
|--------|---------|-----------------|--------------------------|
| `efit` | 0.30 s | 2.6e-05 | 0 |
| `efit_arclength` | 0.19 s | 5.2e-05 | 0 |
| `efit_by_inversion` | 0.78 s | 3.4e-05 | 0 |

---

## psihigh scan — et[1] stability eigenvalue (most important table)

All three methods succeed up to psihigh=1.0 without error. Positive et[1] = stable, negative = unstable/numerically corrupt.

| psihigh | efit | efit_arclength | efit_by_inversion |
|---------|------|----------------|-------------------|
| 0.9800 | +2.20 | **-236** ⚠️ | +2.20 ✓ |
| 0.9850 | +2.16 | +2.41 ✓ | +2.16 ✓ |
| 0.9900 | +1.24 | +1.41 ✓ | +1.27 ✓ |
| 0.9930 | +1.21 | **-77** ⚠️ | +1.20 ✓ |
| 0.9950 | +1.18 | +1.25 ✓ | +1.16 ✓ |
| 0.9960 | +1.17 | **-4750** ⚠️ | +1.09 ✓ |
| 0.9970 | +1.15 | **-264** ⚠️ | +1.16 ✓ |
| 0.9980 | +1.12 | **-305** ⚠️ | +1.08 ✓ |
| 0.9990 | +0.73 | +0.65 ✓ | -1875 ⚠️ |
| 0.9995 | -66 | -12.9 | -2759 |
| 0.9999 | -578k | -1989k | -62k |
| 1.0000 | -8.7M | -24.9M | -641k |

`efit_arclength` produces large spurious negative eigenvalues at 7/12 psihigh values — numerical artifacts from accumulated 1/Bp integration error near the x-point. `efit_by_inversion` tracks `efit` closely through psihigh=0.998, making it the most reliable method for stability analysis over a wide psihigh range.

---

## Grid sensitivity — Sweep C: resolution_factor (adaptive grid, corrected range)

| rf | Runtime | Roundtrip | max\|Δq\| vs efit |
|----|---------|-----------|-------------------|
| 0.5 | 0.04 s | 1.68e-03 | 3.3e-02 |
| 1.0 | 0.04 s | 6.24e-04 | 1.2e-02 |
| 2.0 | 0.06 s | 1.72e-04 | 4.1e-03 |
| 3.0 | 0.11 s | 7.89e-05 | 2.7e-03 |
| **4.0** | **0.19 s** | **4.29e-05** | **1.2e-03** ← default |
| 6.0 | 0.43 s | 2.14e-05 | 8.8e-04 |

The default rf=4.0 gives roundtrip ~4e-05 (comparable to `efit`'s 2.6e-05) at 0.19 s for the adaptive grid alone (full solver ~0.78 s including InverseEquilibrium).

---

## psilow near-axis scan

`efit_by_inversion` works well for psilow ≤ 0.01 (typical use). For psilow ≥ 0.05 the near-axis roundtrip error and q0 extrapolation degrade — this is expected given the InverseEquilibrium solver's handling of large innermost surfaces and is consistent with pre-existing behavior.

| psilow | efit q0 | arclength q0 | inversion q0 |
|--------|---------|--------------|--------------|
| 0.10 | 1.197 | 1.194 | 0.558 ⚠️ |
| 0.01 | 1.209 | 1.209 | 1.266 |
| 0.001 | 1.210 | 1.210 | 1.209 ✓ |
| 0.0001 | 1.210 | 1.210 | 1.209 ✓ |

Recommend documenting that `efit_by_inversion` is intended for psilow ≤ 0.01.

jhalpern30:
@claude summarize the entire above conversation in a brief comment to give me context for reviewing this PR - make the comment as minimal as possible while allowing me to understand all important details discussed above

benchmarks/equil_inv_grid_sensitivity.jl
jhalpern30, line 1: Do we really need all this data/benchmarks? I am a fan of the idea we had where we include them in the commit and then remove then in a final commit before the merge to make them accessible if needed if someone wants to go digging. Considering there are already two more PRs for equil things, this is going to get nightmarish to maintain - better than keeping notebooks up to date, but I can see a world where these all become broken very quickly.

docs/src/equilibrium.md
```diff
@@ -22,6 +22,74 @@ The module exposes a small public API that covers setup, configuration,
 and common analyses used by other GPEC components (e.g. `ForceFreeStates`, vacuum
 interfaces).
 
+## EFIT solver strategies
```
jhalpern30, line 25: I like this - just double check that these numbers are correct. I think they look ok, but want to make sure this isn't something Claude wrote a while ago and is now out of date

examples/DIIID-like_ideal_example/gpec.toml
```diff
@@ -6,7 +6,7 @@ power_bp = 0                            # Poloidal field power exponent for Jaco
 power_b = 0                             # Toroidal field power exponent for Jacobian
 power_r = 0                             # Major radius power exponent for Jacobian
 grid_type = "ldp"                       # Radial grid packing type
-psilow = 1e-4                           # Lower limit of normalized flux coordinate
+psilow = 1e-3                           # Lower limit of normalized flux coordinate
```
jhalpern30: Why are we changing this?

src/Equilibrium/DirectEquilibriumArcLength.jl
jhalpern30, line 1: Your comments say that you shouldn't run with psilow > 0.01 for the equal arc length method - there should be a warning/erroring out somewhere for this

src/Equilibrium/EquilibriumTypes.jl
```diff
@@ -87,6 +87,7 @@ Bundles all necessary settings originally specified in the equil fortran namelis
         elseif jac_type != "other"
             error("Cannot recognize jac_type = $(jac_type)")
         end
+        psihigh = min(psihigh, 1.0)
```
jhalpern30, line 100: Was there a reason this was added? If we're going to do this, I'd wrap this in a warning to the user. "Psihigh must be less than 1, but you have input a larger value. Clamping to 1". But come to think of it, can't you not run with psihigh = 1 anyway due to the issues this PR is trying to address?

test/runtests_equil.jl
```diff
@@ -21,7 +18,71 @@
...
+
+        # B field must be finite and positive
+        B_nodes = plasma_eq_inversion.eqfun_B.nodal_derivs.partials[1, :, :]
+        @test all(isfinite, B_nodes)
+        @test all(>(0), B_nodes)
+    end
+
+    @testset "EFIT Method Consistency" begin
+        # All three methods solve the same equilibrium — q-profiles should broadly agree.
+        # Tolerance is 5% to allow for method-specific discretisation differences.
```
jhalpern30: I don't see where this 5% tolerance is actually used

.gitignore
```diff
@@ -19,3 +19,4 @@ docs/build/
 anaconda_projects/
 modovmc
 pestotv
+examples/DIIID-like_ideal_example/*.csv
```
jhalpern30: I'd rather this just be *.csv or not there at all, this is a bit specific

jhalpern30 (commented):
Overall looks good, all my comments are mostly administrative. I just glanced over the actual numerics (the new equil methods + updated inverse equilibrium) and didn't see anything off. Should be good to go after addressing these, unless you want more equilibrium specific comments from maybe @d-burg

src/Equilibrium/EquilibriumTypes.jl
logan-nc: The PR now includes a check to find the last closed surface from the original EFIT grid - allowing the user to set `psihigh=1.0` to say "go as high as you can possibly go"

logan-nc:
## Profile-driven middle-region grid spacing (`0e3cfed6`)

**Problem**: `h_mid` was previously set to `(1 - psi_split_edge) × dlog`, which ties pedestal resolution to the *edge* log-accuracy criterion — sized for the logarithmic q divergence near the separatrix, not for pressure curvature. For the DIIID H-mode case with A≈0.4, tau=0.005, this gave h_mid≈0.032 and only ~3 knots in ψ=0.85–0.95 — insufficient to resolve the sharp H-mode pedestal. The symptom was visible spline separation between methods in the mid-radius region.

**Fix**: Added `_estimate_mid_spacing(sq_in, psi_split_core, psi_split_edge, tau)`, which samples all 4 profiles (F·2π, μ₀P, dV/dψ, q) at 300 points over the middle region and uses the cubic-spline error bound E ~ (h²/8)|f''|/f_scale < tau solved for h:

```
h_k = sqrt(8 * tau / d2_norm_k)
h_mid = min over k=1..4
```

where `d2_norm_k = max|f''| / max|f|` is estimated by central differences. This is driven by whichever profile has the sharpest feature — almost always the pressure pedestal for H-mode plasmas.

Also tightened the split boundary defaults:
- `psi_split_core`: 0.15 → 0.03 (axis ψⁿ behavior ends much earlier; frees knots for pedestal)
- `psi_split_edge`: 0.95 → 0.98 (pushes pedestal foot into the uniform middle region where h_mid controls resolution)

**Result on DIIID H-mode at psihigh=0.9999**: N_mid increases from ~26 to 73 (h_mid=0.0131), total grid 93 knots. The `psi_accuracy` setting from `gpec.toml` controls all three regions consistently via the shared `tau` parameter.

logan-nc:
Ok. I addressed @jhalpern30's review comments and improved the docs. 

I am happy enough with this for now and am going to merge before it mission creeps any more!!

The 3 methods are all solid options to have and I suspect the "best" one (that can cleanly push to highest psihigh) will vary given the specifics of the gfile. For the DIIID-like example, it turns out the original method can be pushed out quite cleanly. For that, the real big feature add IMO is the automated psi grid size and grid packing. This is done in a way that allows the cubic splines to capture the logarithmic q singularity at the edge (and metric singularities in the core) without ringing. The mid radius packing is set to be sufficient to capture any transport barriers or local current drive. This is a huge win. I think a lot of failed attempts to push to high psihigh were due to bad `mpsi`!

I decided to keep the benchmark scripts (not the results!) for now. I think they will eventually be useful for writing the Julia GPEC paper. We can purge them after that if we want. Here are some final benchmark plots:

Good agreements:
image
image
image

The hardest thing to get turned out to be metric1:
image

But any further tuning for that is left to future work

PR #190, opened 2026-03-13
Improved numerical methods for the vacuum module

jhalpern30:
Per the title, speeding up the vacuum code and making it use less memory. Primarily targeting 3D, but we'll see if any of these help 2D as well

The main goal of this PR is to project the full collocation matrix into Fourier space, making the vacuum response calculation go from O(N^3) compute and O(N^2) memory to O(P^3) compute and O(NP) memory, where N is the physical collocation resolution and P is the number of modes


jhalpern30:
This PR is getting closed per discussions with @logan-nc - ultimately, the added user inputs required due to the truncation error caused by projecting the matrix before inverting was not worth the computational speedup. It risks introducing errors in the result that require expanding the mode space in the vacuum solve to fix (approximately). This is due to the geometric coupling from highly shaped equilibria to other mode numbers, so information is lost at the matrix boundaries when projecting before inverting, which does not occur when performing the current method of inversion in physical space. This PR can be revisited if we really need to prioritize improvements in computational efficiency of this module and are ok with this sacrifice. This implementation is most useful in the limit of grid_points >> num_modes. 

Note: this implementation is working and has been benchmarked. The only missing piece was a robust way to avoid this truncation error without user input and losing numerical benefits.

I am adding all of the non-spectral updates I had in here in #299 to avoid losing them

### PR Summary: Fused Galerkin Vacuum Kernel
Overview
This PR replaces the vacuum module's two-step "evaluate dense kernel → project" approach with a fused Galerkin method that performs Fourier projection simultaneously with kernel evaluation. The result is a mathematically equivalent but more memory-efficient and computationally cleaner implementation for both 2D and 3D vacuum solves.

What Changed
1. Fused Galerkin Kernel Assembly (Kernel2D.jl, Kernel3D.jl)
The core algorithmic change. Previously, each kernel call assembled a full dense M×M point-space matrix (where M = mtheta in 2D or mtheta×nzeta in 3D) and then separately projected it via a Fourier transform. The new path eliminates the dense matrix entirely:

For each observer point j, the kernel row is evaluated and immediately projected into a length-P mode-space vector via rank-1 accumulation into the P×P projected block:
Kc += conj(Z[j,:])' ⊗ (K[j,:] · Z)
Peak memory drops from O(M²) to O(MP + P²), where P = num_modes (typically P ≪ M).
FLOP count is asymptotically unchanged; the dominant cost is still kernel evaluation scaling as O(M² · n_gauss).
2. Complex Exponential Basis (FourierTransforms.jl)
The dual real (cos_mn_basis, sin_mn_basis) pair returned by compute_fourier_coefficients was replaced with a single complex exp_mn_basis:

Z[i, l] = exp(i·(m·θ_i − n·ν_i))
This halves the number of matrix multiplications in the projection step and simplifies the Gram matrix: because Z is a normalized complex DFT matrix, Zᴴ Z = M·I (the Gram is just a scalar, M = num_points). That means there is no need to store or invert a dense Gram matrix at all.

FourierTransform objects used elsewhere still hold real cos/sin parts for backward compatibility (reconstructed from the complex basis on construction).

3. Mode-Space Solve Instead of Point-Space Solve (Vacuum.jl)
The BIE linear system is now assembled and solved entirely in P×P (or 2P×2P with wall) complex mode space:

Before: Build M×M grad_green → LU factor → solve → Fourier-transform the columns of the result.
After: Build P×P (or 2P×2P) projected K_ext / G_ext → LU factor → ldiv! in mode space → wv is read off directly.
The interior kernel K_int is obtained analytically from K_ext via the discrete Green's identity:

K_int = 2·M·I − K_ext
avoiding a second kernel evaluation pass.

grri and grre are reconstructed into their legacy real M×2P layout at the end of the solve for backward compatibility with downstream code.

PR #193, opened 2026-03-20
ForceFreeStates: physics-driven auto-computation of mlow/mhigh and mthvac

logan-nc:
## Summary

- Replaces fixed \`delta_mlow\`/\`delta_mhigh\` padding with \`_estimate_m_bandwidth\`: computes the Fourier spectrum of r² (rzphi_rsquared) at \`psilim\` to find the highest harmonic exceeding \`m_accuracy × DC\`. Uses r² rather than the Jacobian because in Hamada coordinates the Jacobian is constant by definition — r² correctly captures the angular variation of g^{θθ} that drives inter-mode coupling in the EL equations.
- Replaces manual \`mthvac\` with \`_estimate_arc_nonuniformity\`: computes \`mean(ds)/min(ds)\` on the SFL boundary to detect coordinate clustering (e.g. X-point in Hamada), then sets \`mthvac = round_up(max(4*mpert, max(96,6*mhigh)*arc_factor), 32)\`
- Adds \`mlow\`/\`mhigh\` as explicit expert overrides (nonzero bypasses auto) and \`m_accuracy\` as the single tuning knob for mode range
- Removes the \`ctrl.delta_mhigh *= 2\` legacy hack
- Deprecates \`delta_mlow\`/\`delta_mhigh\` with a \`@warn\` (fields retained for backward compat with old configs)
- Also fixes pre-existing bugs: remove phantom PlotlyJS dependency (broke \`Pkg.instantiate\`), fix \`sing_lim!\` BoundsError when \`mpsi=0\` (auto-computed)

## Test results

| Test | Status |
|---|---|
| Unit tests (557 total: Utilities, Vacuum, Equilibrium, ODE, FFS) | ✅ Pass |
| Solovev example | ✅ Pass — bandwidth=7, arc_factor=1.64, mthvac=160, et[1]=-0.459 (consistent) |
| DIIID example | ⚠️ See note below |
| Override/deprecation logic (4 targeted tests) | ✅ Pass |

## ⚠️ DIIID benchmark note — pre-existing issue in parent branch

The DIIID \`et[1]\` values differ from \`develop\` even when mlow/mhigh are set identically to develop's values. Isolation testing confirmed this is **not caused by the auto-m changes**: setting \`mlow=-12, mhigh=21\` (identical to develop) on this branch still produces a different \`et[1]\`.

The source is pre-existing equilibrium differences inherited from \`direct_equilibrium_improvements\`: the auto-mpsi feature produces only 3 edge grid points at \`psilim=0.989\` (vs develop's 128 uniform points), causing under-resolved equilibrium near the separatrix. This is a separate issue to be addressed in \`direct_equilibrium_improvements\` before merging to develop.

The auto-m and auto-mthvac features are self-consistent: DIIID gives bandwidth=4, arc_factor=4.01 (correctly reflecting Hamada coordinate clustering near the X-point), and Solovev validates cleanly against its own baseline.

## Test plan

- [x] Run \`examples/Solovev_ideal_example\` — passes, et[1] consistent
- [x] Run unit tests — 557/557 pass
- [x] Test explicit \`mlow\`/\`mhigh\` override — correctly bypasses auto
- [x] Test \`delta_mlow != 0\` deprecation warning — fires correctly
- [ ] DIIID validation blocked on \`direct_equilibrium_improvements\` equilibrium fix
- [ ] Re-benchmark after parent branch equilibrium issue resolved


matt-pharr:
@logan-nc is this PR going anywhere?

logan-nc:
@matt-pharr I believe this was put on the back burner since until things "settled down" and we became more ok with Julia diverging from fortran. If implemented then it will become harder to directly compare to fortran (still possible - you would just run fortran, read the m range, then set julia mlow and mhigh based on that).

If you and @jhalpern30 think it's time - I can revisit this to bring it up to date and clean it up a bit. 

Rereading the description, one thing that would be a good add is to do the ForcingTerms calculation earlier in the workflow and include that knowledge when setting m ranges - a classic issue is that the plasma physics (DCON) sets some optimized m-range but the simple boxy 3D coils have some broad -m spread and the m-range has to be expanded to fully capture the forcing terms (otherwise weird aliasing can happen). That's probably the number one failure mode that makes people have to re-run cases with more delta_m*, so we should automate the logic.

It would be good to test on some more cases too. I feel like the AIs always over optimize local conditions... perhaps the shaping scan? That is a script in benchmarks now, right? A q95 scan would be good too.

matt-pharr:
@logan-nc the scan tests will be much easier after I finish the tokamaker integration but that is still a ways out, will likely not want to finish that until we make our big refactor push

jhalpern30:
I think this is useful at some point but probably a distraction on the path to GPEC 2.0 and can wait

logan-nc:
> @logan-nc the scan tests will be much easier after I finish the tokamaker integration but that is still a ways out, will likely not want to finish that until we make our big refactor push

@matt-pharr Don't we have a aspect ratio scan in a benchmark script from @d-burg? Is that on develop now? I guess it uses TJ, so can do q95 scans and aspect ratio but not shaping? Solovev can do some shaping I suppose.

Issue #170, opened 2026-02-25
Unit tests for psiedge < psilim logic


jhalpern30:
Currently, we don't have anything testing the edge truncation logic; we should add some unit tests for this since we could accidentally break this easily and not know. This could include 

- A full run test with solovev and psiedge set to less than 1, confirming the code runs through
- Unit tests for the `findmax_dW_edge!` functionality (`free_compute_wv_spline` and `free_compute_total`, which might be difficult since they currently accept the entire structs including the equilibria.

PR #194, opened 2026-03-20
ForceFreeStates: backport 4 edge dW scan bug fixes to develop

logan-nc:
## Summary

Backports two bug fixes and three structural improvements from `edge_spline_extension` to `develop`. These affect all plasma configurations (not just diverted) and can produce incorrect edge stability scan results. Both bugs cause the edge scan normalization to use the wrong psi value.

- **Bug 1** (`free_compute_total`): dV/dψ normalization used `intr.psilim` (fixed for the whole scan) instead of `odet.psifac` (the current scan psi). Each scan step must normalize by the local dV/dψ at its own boundary.

- **Bug 2** (`free_compute_wv_spline`): `VacuumInput` was using `intr.psilim` as the plasma boundary for every spline grid point instead of the actual scan psi `psi_array[i]`. This pinned the vacuum response to the final boundary, producing `norm_wv ≈ 26840` instead of the correct `≈ 5730`.

## Structural improvements (from jhalpern30 and claude-bot reviews)

- **`EdgeScanState` struct**: Extracted 14 edge-scan fields from `OdeState` into a dedicated struct. `OdeState.edge_scan::Union{EdgeScanState, Nothing}` is `nothing` when no edge scan is configured.
- **Vacuum API**: `free_compute_wv_spline` was regressed to a manual per-n loop; fixed to use the multi-n `compute_vacuum_response` API matching `free_run!` (PR #163 behavior).
- **Singfac**: Applied analytically in `free_compute_total` (not pre-baked into spline) using the same vectorized form as `free_run!`, avoiding interpolation of a zero-crossing function near rational surfaces.
- **Bounds fix**: `EdgeScanState.dW_edge` is now allocated with the exact edge step count (post-integration `findall`) rather than `numsteps_init`, which could be exceeded after `resize_storage!`. The peak index is correctly mapped from compact EdgeScan index back to ODE step index for truncation.
- **Variable naming**: Internal variables in `free_compute_total` renamed to match HDF5 output names (`total_eigenvalue`, `plasma_energy`, `vacuum_energy`, `vacuum_eigenvalue`).
- **evonly normalization**: `vacuum_eigenvalue` now divides by the same complex `norm` as the other three energy outputs (was `real(norm)`, now `real(.../ norm)` for type correctness).

## Test plan

- [x] Run DIIID-like example (`examples/DIIID-like_ideal_example`) and confirm `et[1] = +1.200e+00` (stable, matches develop reference)
- [x] No `BoundsError` in `findmax_dW_edge!`
- [x] Edge scan HDF5 output (`edge_scan/total_eigenvalue`, etc.) populated correctly


logan-nc:
@claude review this PR

logan-nc:
## Summary of fixes in this PR

This PR backports edge dW stability scan fixes from `edge_spline_extension` to `develop`. Several bugs were found during testing beyond the original 4 planned fixes:

### Original 4 bug fixes

1. **`free_compute_wv_spline` plasma boundary** — VacuumInput was called with `intr.psilim` for all grid points instead of the actual `psi_array[i]`, so every vacuum matrix in the spline was computed at the same wrong boundary.

2. **`free_compute_total` dV/dψ normalization** — kinetic norm used `dVdpsi_spline(intr.psilim)` instead of `dVdpsi_spline(odet.psifac)`, giving wrong normalization at every scan step except the last.

3. **`transform_u!` / `evaluate_stability_criterion!` ordering** — `transform_u!` was called after `evaluate_stability_criterion!`. The criterion needs transformed (true) solution vectors; using raw pre-transform vectors caused spurious zero crossings in the core where the cumulative Gaussian reduction matrix is ill-conditioned.

4. **`SingularException` guard in `check_for_zero_crossings!`** — `U₂` can be degenerate near rational surfaces in the above-psihigh integration region, causing an unhandled exception. Now propagates the previous `crit_store` value on catch.

### Additional bugs found during testing

5. **`free_compute_wv_spline` API fix** — `intr.mtheta` → `ctrl.mthvac` (field was renamed in develop vacuum API).

6. **Singfac pre-applied in spline** — singfac `(m − n·q)` was baked into the spline grid values. Interpolating a function that passes through zero at every rational surface distorts the peaks between surfaces. Now stores raw wv and applies singfac analytically in `free_compute_total` using the exact local `q(psifac)` — matching the `edge_spline_extension` design exactly.

7. **wv spline grid off-by-one** — grid indexed as `i/npsi` (spans `[qedge+Δq, qlim+Δq]`) instead of `(i-1)/npsi` (spans `[qedge, qlim]`). The last grid point overshot `psilim` into invalid extrapolated equilibrium geometry.

8. **Sparse ODE saving between rational surfaces** — the ODE integrator takes only a few large steps in the smooth region between rational surfaces, so with `save_interval=10` the midpoint peaks (e.g. q=5.5, 6.5) were never sampled. Peak appeared shifted toward rational surfaces. Fixed by always saving when `psifac >= psiedge`.

9. **`transform_u!` bounds error with post-peak fixsteps** — after `trim_storage!` cuts `u_store` to the peak dW step, `fixstep` entries from Gaussian reductions that occurred after the peak can reference indices beyond the trimmed array. Fixed by capping `kfix = min(fixstep[ifix], odet.step)` and breaking early when `jfix > odet.step`.

### New outputs and analysis

- All edge scan energy components written to `edge_scan/` HDF5 group: `psi`, `q`, `total_energy`, `plasma_energy`, `vacuum_energy`, `vacuum_eigenvalue` — key names match `edge_spline_extension` exactly for a trivial rebase.
- `vacuum_eigenvalue` is the minimum eigenvalue of the singfac-scaled vacuum matrix, normalized by the same kinetic norm as the other energies for direct comparison.
- `Analysis.ForceFreeStates.plot_edge_stability_scan`: 4-panel plot vs q with stability boundary (hline at 0) and qlim marker. Verified to show symmetric peaks at rational surface midpoints with clean zeros at each rational surface.

### Performance

Per-call heap allocations in `free_compute_total` (called at every stored ODE step in the edge region) reduced from ~8 allocations to LAPACK-unavoidable only, via 5 pre-allocated buffers added to `OdeState`: `_wv_scratch`, `_singfac_buf`, `_eindex_buf`, `_evals_real_buf`, `_tmp_vec`.

logan-nc:
Here is the bug-fixed result showing proper oscillation of the vacuum energy matrix eigenvalue in rational windows:

image

logan-nc:
## Benchmark & Correctness Update (post-develop merge)

### Result comparison

| Configuration | et[1] | Notes |
|---|---|---|
| Pre-merge benchmark (develop, old psi_accuracy=0.005) | 1.7654 | Prior benchmark result |
| Current develop (psi_accuracy=0.001, qlim via dmlim) | 1.200 | EQUIL PR #187 tightened radial grid |
| **This branch, same settings** | **1.200** | ✅ Identical to develop — merge is clean |
| This branch, edge scan (psihigh=0.9993, psiedge=0.95) | 0.992 | Different truncation point (q=6.095) |

The 1.7654 → 1.200 change is **not a regression from this PR** — it comes entirely from the EQUIL improvements (psi_accuracy 0.005→0.001, mpsi bug fix) merged in PR #187. Both branches agree exactly with the same settings.

### Post-merge bugs fixed

The develop merge re-introduced three bugs in `free_compute_wv_spline` (commit `bb84917c`):

1. **Stale psi** (`intr.psilim` → `psi_array[i]`): all spline nodes were built at the same boundary location instead of sweeping from psiedge to psilim
2. **Old VacuumInput signature**: updated to post-develop API (`nzeta`, `npert=1`, `nlow=n`)
3. **Double singfac**: develop re-applied singfac in the spline-building loop, while `free_compute_total` (from this PR) still applies it analytically at each scan step → singfac² instead of singfac. Fix: store raw wv in spline; singfac belongs only in `free_compute_total`

### Edge scan plot (psihigh=0.9993, psiedge=0.95)

edge_stability_scan

Truncation at peak et: ψ = 0.999, q = 6.095. All four energy panels (total, plasma, vacuum, min vacuum eigenvalue) show physically reasonable behavior with the mode becoming marginally stable near the edge.

logan-nc:
## Benchmark & Correctness Update (post-develop merge)

### Result comparison

| Configuration | et[1] | Notes |
|---|---|---|
| Pre-merge benchmark (develop, old psi_accuracy=0.005) | 1.7654 | Prior benchmark result |
| Current develop (psi_accuracy=0.001, qlim via dmlim) | 1.200 | EQUIL PR #187 tightened radial grid |
| **This branch, same settings** | **1.200** | ✅ Identical to develop — merge is clean |
| This branch, edge scan (psihigh=0.9993, psiedge=0.95) | 0.992 | Different truncation point (q=6.095) |

The 1.7654 → 1.200 change is **not a regression from this PR** — it comes entirely from the EQUIL improvements in PR #187 (psi_accuracy 0.005→0.001, mpsi bug fix). Both branches agree exactly with the same settings.

### Post-merge bugs fixed (commit `bb84917c`)

The develop merge re-introduced three bugs in `free_compute_wv_spline`:

1. **Stale psi**: `intr.psilim` → `psi_array[i]` — all spline nodes were being built at the same fixed boundary instead of sweeping from psiedge to psilim
2. **Old VacuumInput signature**: updated to post-develop API (added `nzeta`, changed to `npert=1, nlow=n` form)
3. **Double singfac**: develop had re-applied singfac in the spline-building loop, while `free_compute_total` (from this PR) still applies it analytically at each step → wv was scaled by singfac² instead of singfac. Fix: store raw wv in spline; singfac belongs only in `free_compute_total`

### Edge scan (psihigh=0.9993, psiedge=0.95, no wall)

Truncation at peak et: ψ = 0.999, q = 6.095. All four energy panels (total, plasma, vacuum, min vacuum eigenvalue) are physically reasonable — the mode approaches marginal stability near the edge with clearly resolved energy components.

Plot saved at `/tmp/edge_stability_scan.png` (run `FFS.plot_edge_stability_scan("gpec.h5"; ylims=(-2,3))` to regenerate).

src/ForceFreeStates/EulerLagrange.jl
```diff
@@ -77,15 +77,16 @@ function eulerlagrange_integration(ctrl::ForceFreeStatesControl, equil::Equilibr
         trim_storage!(odet)
     end
 
+    # Undo Gaussian reduction to get true solution vectors. Must precede evaluate_stability_criterion!,
```
jhalpern30: This doesn't fully make sense to me, but I'm assuming you've benchmarked it so its ok. 

In the Fortran, "fixups" were performed in `idcon` in the loading stage of running IPEC, the main DCON analysis never actually used them. So DCON has worked for decades without pre-transforming the solution matrices before performing the eigenvalue analysis for fixed boundary stability, making me concerned on why we need this here. And if we're calling this a bug (or at least claude is), then is this a bug that we introduced? Or is this not really a bug-fix and more just provides some additional stability to the method?

Note this is assuming you agree with my interpretation of what the Fortran workflow was above.

src/ForceFreeStates/Free.jl
```diff
@@ -137,18 +137,15 @@ function free_compute_wv_spline(ctrl::ForceFreeStatesControl, equil::Equilibrium
             psii, Roots.Newton()
         )
 
-        # Compute vacuum response matrix at this psi (2D single-n, 2D multi-n block-diagonal, or 3D)
-        vac_inputs = Vacuum.VacuumInput(equil, psii, ctrl.mthvac, ctrl.nzvac, intr.mpert, intr.mlow, intr.npert, intr.nlow; force_wv_symmetry=ctrl.force_wv_symmetry)
-        wv, _, _, _, _ = Vacuum.compute_vacuum_response(vac_inputs, intr.wall_settings)
+        for ipert_n in 1:intr.npert
```
jhalpern30: This wasn't merged right and regressed back to the old vacuum code - the new code from #163 doesn't loop through n (it does this under the hood in the vacuum module)

src/ForceFreeStates/Free.jl
```diff
@@ -127,8 +127,8 @@ function free_compute_wv_spline(ctrl::ForceFreeStatesControl, equil::Equilibrium
     wv_array = zeros(ComplexF64, npsi + 1, intr.numpert_total, intr.numpert_total)
 
     for i in 1:(npsi+1)
-        # Space points evenly in q
-        qi = qedge + (intr.qlim - qedge) * (i / npsi)
+        # Space points evenly in q over [qedge, qlim] (i=1 → qedge, i=npsi+1 → qlim)
```
jhalpern30, line 130: From a quick glance, if you remove the singfac scaling you actually don't even need qi - its only used to calculate m - nq at each interpolation point

src/ForceFreeStates/ForceFreeStatesStructs.jl
```diff
@@ -375,6 +375,12 @@ and a small set of temporary matrices and factors used to compute singular-layer
     with shape `(numpert_total, numpert_total, 2, msing)`.
   - `dW_edge::Vector{ComplexF64}` - dW values computed in the psiedge < psilim region for each stored step (length `numsteps_init`).
   - `wvmat::CubicSeriesInterpolant{Float64,ComplexF64}` - Complex-valued precomputed wv matrices used by `free_test`/vacuum routines.
+  - `psi_edge_scan::Vector{Float64}` - ψ values at each edge scan step (psifac ≥ psiedge). Populated by `findmax_dW_edge!`.
```
jhalpern30: I am not a fan of adding a ton of extra entries to this struct, its already a rather large one. We have a few issues open on moving away from these large structs that contain a ton of parameters and where you don't know what is modified in each function. 

All of this analysis is performed in a self-contained section of the code in `findmax_dW_edge!`. If you want all of this data in a struct, why not just a new `EdgeScan` struct that stores all of this? It can be initialized there, and then filled, and optionally returned at the end of the function? By my brief count, there are now 14 parameters in the OdeState struct relating solely to the edge dW, which I think warrants a new struct.

src/ForceFreeStates/Free.jl
```diff
@@ -172,42 +169,92 @@ wv matrix spline to `free_compute_wv_spline` and pass it in `odet.wvmat` (a comp
 """
 function free_compute_total(equil::Equilibrium.PlasmaEquilibrium, ffit::FourFitVars, intr::ForceFreeStatesInternal, odet::OdeState)
 
-    # Allocations
-    wp = zeros(ComplexF64, intr.numpert_total, intr.numpert_total)
-    tot_eigvals = zeros(ComplexF64, intr.numpert_total)
-    wt = zeros(ComplexF64, intr.numpert_total, intr.numpert_total)
+    Npert = intr.numpert_total
```
jhalpern30, line 167: I am 90% sure all of this buf logic is directly addressed by Min-Gu's adaptive array pooling. You can just wrap the function with `@with_pool pool` and use something like 
`wp = zeros!(pool, ComplexF64, Npert, Npert`
and so on for all allocations in this function. They don't need to be buffers within the struct. Looking at the below code, that would reduce a ton of the extra added lines by just doing the pool allocations up here for whatever you need

jhalpern30: Note you can see `free_run!` above for an example of how to do this

src/ForceFreeStates/Free.jl
```diff
@@ -172,42 +169,92 @@ wv matrix spline to `free_compute_wv_spline` and pass it in `odet.wvmat` (a comp
...
 
-    # Compute vacuum matrix from series interpolant (use separate hint for wv grid)
-    # FastInterpolations now natively supports complex values
+    # Retrieve raw vacuum matrix from spline (singfac NOT pre-applied; see free_compute_wv_spline).
+    # Apply singfac = (m - n*q(psifac)) analytically so the correct local q is used at each scan step,
+    # rather than an interpolation of singfac-pre-applied values which is inaccurate near rational surfaces
+    # (singfac passes through zero there, and interpolating a zero-crossing function distorts the peaks).
     odet.wvmat(vec(odet._wv_out), odet.psifac; hint=odet.wv_hint)
-    wv = odet._wv_out
+    wv = odet._wv_scratch
```
jhalpern30: Why not use the simplified logic that was deleted from `free_compute_wv_spline`?
```julia
singfac = vec((intr.mlow:intr.mhigh) .- qi .* (intr.nlow:intr.nhigh)')
@inbounds for ipert in 1:intr.numpert_total
    @views wv[ipert, :] .*= singfac[ipert]
    @views wv[:, ipert] .*= singfac[ipert]
end
```

src/ForceFreeStates/ForceFreeStatesStructs.jl
```diff
@@ -418,6 +424,29 @@ and a small set of temporary matrices and factors used to compute singular-layer
     wvmat::CubicSeriesInterpolant{Float64,ComplexF64} = _empty_series_interp_complex(numpert_total^2)
     _wv_out::Matrix{ComplexF64} = Matrix{ComplexF64}(undef, numpert_total, numpert_total)
 
+    # Pre-allocated buffers for free_compute_total to avoid per-call heap allocations.
```
jhalpern30: See comments in `free_compute_total` - I don't think these are needed

src/ForceFreeStates/FixedBoundaryStability.jl
```diff
@@ -65,7 +65,16 @@ can do it post-integration rather than during and don't directly handle file out
     psi = odet.psi_store[istep]
     u .= odet.u_store[:, :, :, istep]
     dVdpsi = profiles.dVdpsi_spline(psi; hint=odet.spline_hint)
-    crit_val, nonherm = compute_smallest_eigenvalue(u)
+    crit_val, nonherm = try
```
jhalpern30: I have never seen the inversion fail before with a SingularException that wasn't just a bug in the code I introduced so I am not sure why this is needed - I am assuming this is something you now see when pushing really far out in the edge when rationals get much closer together? Ok to leave here, but just wanted to make sure this was intentionally added

jhalpern30 (changes requested):
I think the most major things that need to be modified are consolidating the now 10+ entries that relate to the edge scan into its own struct since its crowding `OdeState` and using adaptive pooling on `free_compute_total!` to remove all the buffers and really simplify the code.

I am already asking for a lot of changes so don't want to pile on - but if you're inspired, I think we're lacking on unit testing in this area, which probably led to all of this breaking in the first place. I had raised an issue on this #170, and with this nice cleanup/improvements to the code, now might be a good time to make sure it doesn't break again

src/ForceFreeStates/EulerLagrange.jl
logan-nc: It's entirely possible this was an iterative issue with me looking at the results and telling Claude they were not physical, prompting claude to bend over backwards to please me. I will try removing it now that the other, real bugs are fixed.

src/ForceFreeStates/Free.jl
logan-nc: Thanks. Yeah, IDK how I ended up doing this backport on a super stale develop branch - I must have switched to a local version instead of the remote one 🤦 totally ruins the whole point of a clean backport bug fix - I was very sad when I realized the develop-to-this merge was huge 😭

src/ForceFreeStates/ForceFreeStatesStructs.jl
logan-nc: Agreed

logan-nc:
## Response to jhalpern30 review

Four commits addressing the review comments:

**Revert transform_u! ordering + SingularException guard** (`6c256cba`)
Tested reverting to the Fortran-equivalent order (evaluate stability criterion on raw u_store, then transform). Results are identical — et[1] and edge scan plot unchanged. The ordering swap and guard were overcorrections driven by the other real bugs, not a fundamental requirement. Both reverted.

**Fix free_compute_wv_spline vacuum API** (`1c332465`)
Replaced the manual per-n loop with the single multi-n `compute_vacuum_response` call from develop (PR #163 API), matching `free_run!`'s pattern. Two targeted bug fixes applied on top: correct per-point psi boundary (`psi_array[i]`), and raw wv stored without singfac pre-application.

**Simplify singfac in free_compute_total** (`c69eb158`)
Uses the same vectorized form as `free_run!`: `singfac = vec((mlow:mhigh) .- q .* (nlow:nhigh)')` with row/column scaling loop. Removed `_singfac_buf` from OdeState.

**EdgeScanState struct** (`29ebaf49`)
Extracted 14 edge-scan fields from OdeState into a new `@kwdef mutable struct EdgeScanState`. OdeState now holds `edge_scan::Union{EdgeScanState, Nothing} = nothing`, initialized by `findmax_dW_edge!` only when psiedge < psilim. Field names cleaned up — no redundant `_edge_scan` suffix: `psi`, `q`, `total_eigenvalue`, `plasma_energy`, `vacuum_energy`, `vacuum_eigenvalue`.

logan-nc:
Here is the new edge scan energies: Still looking physically correct.
image

logan-nc:
@claude review

jhalpern30:
@logan-nc ping me when you want me to give this another look

src/ForceFreeStates/ForceFreeStatesStructs.jl
```diff
@@ -413,10 +457,8 @@ and a small set of temporary matrices and factors used to compute singular-layer
     ca_r::Array{ComplexF64,4} = Array{ComplexF64}(undef, numpert_total, numpert_total, 2, msing)
     ca_l::Array{ComplexF64,4} = Array{ComplexF64}(undef, numpert_total, numpert_total, 2, msing)
 
-    # Used for to find peak dW in the edge
-    dW_edge::Vector{ComplexF64} = Array{ComplexF64}(undef, numsteps_init)
-    wvmat::CubicSeriesInterpolant{Float64,ComplexF64} = _empty_series_interp_complex(numpert_total^2)
-    _wv_out::Matrix{ComplexF64} = Matrix{ComplexF64}(undef, numpert_total, numpert_total)
+    # Edge dW scan state and results (nothing when psiedge >= psilim, i.e. no edge scan)
+    edge_scan::Union{EdgeScanState, Nothing} = nothing
```
logan-nc: Is this type stable? If not, is there a better way since we really know it is indeed going to be a `EdgeScanState`?

src/ForceFreeStates/EulerLagrange.jl
```diff
@@ -522,23 +538,51 @@ for clarity. We create the wv matrix spline once prior to the loop.
...
+            # U₁ is singular at this step; skip it (dW_edge stays -Inf)
         end
     end
 
-    # Return the index that maximizes dW_edge to identify truncation point
-    return findmax(real.(odet.dW_edge))[2]
+    # Build per-step scan arrays; failed steps have -Inf values → NaN for HDF5 output
+    es.psi              = odet.psi_store[edge_start:odet.step]
+    es.q                = odet.q_store[edge_start:odet.step]
+    es.total_eigenvalue  = ComplexF64[isfinite(real(v)) ? v : complex(NaN) for v in es.dW_edge]
```
logan-nc: Why bother using `-Inf` and vonverting to `NaN`? Is it because we are looking for a `max` of `dW_edge`? If so, is there not an equivilent to `nanmax` in python that would make the overall code more simple since we wouldn't need these conversions?

src/ForceFreeStates/EulerLagrange.jl
```diff
@@ -522,23 +538,51 @@ for clarity. We create the wv matrix spline once prior to the loop.
...
+    for j in 1:N_edge
+        istep = edge_start + j - 1
         odet.psifac = odet.psi_store[istep]
-        if odet.psifac >= ctrl.psiedge
-            odet.u .= odet.u_store[:, :, :, istep]
-            odet.dW_edge[istep] = free_compute_total(equil, ffit, intr, odet)
+        odet.u .= odet.u_store[:, :, :, istep]
+        try
+            result = free_compute_total(equil, ffit, intr, odet)
+            es.dW_edge[j]            = result.total_eigenvalue
```
logan-nc: This naming convention is inconsistent. `dW_edge` should be `total_eigenvalue_arr` for consistency (see related comment below where this is actually defined - causing confusion). Lets drop the `_arr` on all these names since everything in this EdgeScanState struct is implicitly a scan (i.e. an array). Or, if a designation is needed, lets use *_scan instead of *_arr

src/ForceFreeStates/Free.jl
logan-nc: psi_array[i] is calculated by finding where the `q_spline` equals `qi`. The psi is needed for the proper vacuum response on the right surface

src/ForceFreeStates/Free.jl
```diff
@@ -168,46 +160,90 @@ Compute total complex energy eigenvalue (total1). This is a trimmed down version
...
-    # FastInterpolations now natively supports complex values
-    odet.wvmat(vec(odet._wv_out), odet.psifac; hint=odet.wv_hint)
-    wv = odet._wv_out
+    # Retrieve raw vacuum matrix from spline and apply singfac analytically at the local q.
+    # Singfac is not pre-applied in the spline (see free_compute_wv_spline) to avoid interpolating
+    # a zero-crossing function near rational surfaces, which would distort the peaks.
+    es = odet.edge_scan
+    es.wvmat(vec(es._wv_out), odet.psifac; hint=es.wv_hint)
+    wv = es._wv_scratch
+    copyto!(wv, es._wv_out)
```
logan-nc: This seems overly complicated. How is `eq._wv_scratch` the thing being copied if `es._wv_out` is the one being passed to the spline eval? It is opaque for code readers to follow. Hopefully, this can be cleaned up by using array pools as @jhalpern30 suggested above.

src/ForceFreeStates/EulerLagrange.jl
logan-nc: This particular one is the worst offender, where we save two arrays that are essentially the same thing: dW_edge and total_eigenvalue

src/ForceFreeStates/Free.jl
```diff
@@ -168,46 +160,90 @@ Compute total complex energy eigenvalue (total1). This is a trimmed down version
...
+    mul!(tmp_v, wp, v)
+    plasma_energy = ComplexF64(dot(v, tmp_v)) / norm
+    mul!(tmp_v, wv, v)
+    vacuum_energy = ComplexF64(dot(v, tmp_v)) / norm
+
+    # Smallest eigenvalue of the vacuum matrix alone, normalized by the same kinetic norm as the other energies
+    # so all four outputs are directly comparable. The singfac-scaled wv should be PSD by
+    # construction (congruence of PSD wv_raw), but numerical noise can make eigenvalues slightly
+    # negative. Clamp to zero to enforce the physical constraint.
+    # Hermitianize wv in-place (safe: plasma_energy and vacuum_energy already computed above).
```
logan-nc: Do we need this for loop when we already have nice built-ins like `Hermitian(wv)` below?

logan-nc (commented):
A few minor cleanups are needed

logan-nc:
@jhalpern30 apparently we have it set such that I am blocked from merging until you remove your request for changes

src/ForceFreeStates/EulerLagrange.jl
```diff
@@ -522,23 +538,40 @@ for clarity. We create the wv matrix spline once prior to the loop.
...
+    # Loop with compact index j into EdgeScanState; ODE index is edge_start + j - 1.
+    # Result arrays are pre-initialized to NaN; failed steps (SingularException) stay NaN.
+    for j in 1:N_edge
+        istep = edge_start + j - 1
         odet.psifac = odet.psi_store[istep]
-        if odet.psifac >= ctrl.psiedge
-            odet.u .= odet.u_store[:, :, :, istep]
-            odet.dW_edge[istep] = free_compute_total(equil, ffit, intr, odet)
+        odet.u .= odet.u_store[:, :, :, istep]
+        try
```
jhalpern30: Is this SingularException also vestigial Claude add?

src/ForceFreeStates/ForceFreeStatesStructs.jl
```diff
@@ -413,10 +444,8 @@ and a small set of temporary matrices and factors used to compute singular-layer
     ca_r::Array{ComplexF64,4} = Array{ComplexF64}(undef, numpert_total, numpert_total, 2, msing)
     ca_l::Array{ComplexF64,4} = Array{ComplexF64}(undef, numpert_total, numpert_total, 2, msing)
 
-    # Used for to find peak dW in the edge
-    dW_edge::Vector{ComplexF64} = Array{ComplexF64}(undef, numsteps_init)
-    wvmat::CubicSeriesInterpolant{Float64,ComplexF64} = _empty_series_interp_complex(numpert_total^2)
-    _wv_out::Matrix{ComplexF64} = Matrix{ComplexF64}(undef, numpert_total, numpert_total)
+    # Edge dW scan state and results (disabled sentinel when psiedge >= psilim, i.e. no edge scan)
+    edge_scan::EdgeScanState
```
jhalpern30: Why not just `edge_scan::EdgeScanState = EdgeScanState(numpert_total, 0)` to match the rest of the struct and avoid changing the constructor?

src/ForceFreeStates/Free.jl
jhalpern30: Yeah I'm dumb, agreed. Was a hidden line in the git diff so I missed it

jhalpern30 (approved):
This looks super clean, thanks for adding the comments. Only two more minor things as I skimmed through that you can decide whether or not its worth it to do

PR #196, opened 2026-03-22
Perturbed Equilibrium and Coil Biot-Savart

logan-nc:
## Summary

- **New `\"coil\"` forcing format**: reads 3D coil geometry from ASCII `.dat` files, computes Biot-Savart field on the plasma boundary, projects to the normal component, and Fourier-decomposes into (m,n) `ForcingMode` amplitudes — fully replacing the Fortran `coil.F` / `field_bs_psi` pipeline
- **118 bundled coil geometry files** in `src/ForcingTerms/coil_geometries/` covering D3D, NSTX/NSTX-U, ITER, KSTAR, JET, MAST/MAST-U, AUG, EAST, COMPASS, and RFXMOD
- **Visualization functions** in `src/Analysis/CoilForcing.jl`: 3D wire geometry, R-Z cross-sections with plasma boundary overlay, bn(θ,ζ) contour, and (m,n) mode spectrum
- **Normalization bug fixes**: added R-factor to `project_normal_flux!`, SFL toroidal angle correction (rzphi_nu) to Biot-Savart observation points, file normalization convention (`normal_field_T` / `sfl_flux_Wb`)
- **Validation**: Julia coil Phi_x matches Fortran Phi_x with mean ratio 1.0000 ± 0.0002 across all 34 modes for DIII-D n=1 case
- **Existing `\"ascii\"` and `\"hdf5\"` formats are completely unchanged**

## Key bug fixes (since initial PR)

| Commit | Fix |
|--------|-----|
| `b3234dbd` | `project_normal_flux!`: add missing R factor; add file normalization tag support (`normal_field_T` / `sfl_flux_Wb`) with auto-conversion on load |
| `a0f3afed` | Add SFL toroidal angle correction (`rzphi_nu`) to Biot-Savart observation points — omitting it scrambled Fourier decomposition (±m symmetry error) |
| `b3e002c7` | `plot_coil_geometry_rz`: use outermost grid point (`rzphi_xs[end]`) instead of extrapolating to ψ=1.0 |
| `d222b8b3` | `PerturbedEquilibrium`: fix toroidal phase correction in surface inductance |
| `106a4133` | `PerturbedEquilibrium`: fix Jacobian/area weighting in b_n, Jbgradpsi, and Phi_res |

## Perturbed equilibrium field reconstruction agreement (106a4133)

Three Jacobian/area fixes bring field reconstruction into agreement with Fortran GPEC for the DIII-D ideal example:

| Quantity | Status |
|----------|--------|
| Jbgradpsi | ✅ Matches (< 1% RMS across profile) |
| xi_psi | ✅ Matches (< 1% RMS across profile) |
| b_n | ✅ Matches (< 2% RMS across profile) |
| xi_n | ✅ Matches (< 2% RMS across profile) |
| Phi_res | ✅ Improved from ~97% → ~7-12% error (inner surfaces); residual from P-matrix differences |
| w_isl | ✅ 3-17% error (inner surfaces) |

## Benchmarks

`benchmarks/benchmark_coil_ForcingTerms_against_fortran.jl` — 4-panel validation figure against a Fortran run directory (reads equil.in, dcon.in, coil.in automatically).

`benchmarks/benchmark_against_fortran_run.jl` — full PE comparison against Fortran GPEC outputs. Default mode uses coil geometry (harder test); `--use-file-forcing` copies Fortran Phi_coil directly. Prints Phi_x ratio table then PE resonant surface comparison (psilim, qlim, W_t, q profile, Mercier, Phi_res, island widths, Chirikov).

## New files

| File | Purpose |
|------|---------| 
| `src/ForcingTerms/CoilGeometry.jl` | Structs, `.dat` parser, shift/tilt transforms |
| `src/ForcingTerms/BiotSavart.jl` | Threaded midpoint-rule Biot-Savart with `PrecomputedSegments` + `@simd` |
| `src/ForcingTerms/CoilFourier.jl` | Boundary sampling, normal projection, Fourier decomp, top-level entry point |


logan-nc:
We are close. Total field on the control surface matches. Pretty good resonant metrics. I suspect claude is comparing b_psi to b_n in the profiles:
image

logan-nc:
## Session update — b_n and xi_n physical normal field outputs

**Previous session** fixed the permeability matrix P (L was ~500× too small, P had 10–4000% errors). After those fixes: L diagonal 0% error, P diagonal 0.1–3.5%.

**This session** adds correct physical normal field outputs and fixes a benchmark unit mismatch.

### Root cause identified
`b_psi` in Julia = Fortran's `Jbgradpsi` (J×b^ψ, Jacobian-weighted contravariant ψ field). The benchmark was comparing this to Fortran's `b_n` (physical normal field), which differ by a factor of J(θ)×|∇ψ|(θ) — a theta-dependent geometric quantity.

### Changes
- **`FieldReconstruction.jl`**: New `compute_b_n_xi_n_modes` — converts Jbgradpsi/J×ξ_ψ to physical normal components via IDFT → divide by J(θ)×|∇ψ|(θ) → forward DFT. Matches Fortran `gpout_xbnormal` (gpout.f lines 3353–3358).
- **`PerturbedEquilibriumStructs.jl`**: Added `b_n_modes`, `xi_n_modes` fields to `PerturbedEquilibriumState`.
- **`Response.jl`**: Calls `compute_b_n_xi_n_modes` after field reconstruction; stores results in state.
- **`Utils.jl`**: HDF5 output: `b_psi` renamed to `Jbgradpsi`; `b_n` and `xi_n` added as new fields under `perturbed_equilibrium/response/`.
- **`benchmark_against_fortran_run.jl`**: Row 3 panels now correctly compare Julia `b_n` to Fortran `b_n`.

### Remaining known discrepancy
- `Phi_tot` edge modes (m=13–21, m=−12 to −5): 15–87% errors, likely due to mode truncation (Fortran includes ±8 extra modes beyond main range).
- `Phi_res` (~97% errors) — separate SingularCoupling issue not yet addressed.

logan-nc:
Ok. This PR obviously had a lot of mission creep, but its good to have some "working" perturbed equilibrium calculations. Our current state is pretty good! Our b and xi profiles now align almost perfectly. We still have some small discrepancies in the resonant quantities, but nothing huge like a wrong area normalization would cause:

image

logan-nc:
As for the small difference are currently in there for the Phi_tot, I think these come from our improvements to the robustness of the vacuum calculations. You can see our upper left and lower right of both the L and Λ matrices is a bit smoother than the noisy fortran. I think I like the look of the julia more, actually. 

image


The net result is small differences in P, and thus the plasma response. But as can be seen in the spectra and profiles of the perturbed equilibrium, these are small.

logan-nc:
## Delta' singular coupling — accuracy status

Recent commits on this branch improve the singular coupling (Δ') computation in `SingularCoupling.jl`. Here is the current comparison against the Fortran GPEC DIIID ideal example (`euler_stride=1`, `n=1`):

| q | ψ | |Φ_res| error | |w_isl| error | Notes |
|---|---|---|---|---|
| 2 | 0.594 | 18.6% | 8.9% | |
| 3 | 0.819 | 14.3% | 6.9% | |
| 4 | 0.928 | 6.4% | 3.1% | Previously 40.9% — major improvement |
| 5 | 0.988 | 50.7% | 22.7% | |

### What was fixed

**`compute_singular_coupling_metrics!`**: The derivative `du/dψ` at the bracket points (used to form `bwp1 = 2πi·χ₁·(singfac·xsp1 − n·q₁·xsp)`) now uses the **chord slope from `u_store`** rather than the stored `ud_store`. Near outer surfaces (q=4/5), `ud_store` can be ~10% off due to ODE solver discretization in rapidly-varying regions; the near-cancellation in the `bwp1` formula then amplifies this into a large Δ' error. The chord slope uses only `u_store` values, which are accurately stored, avoiding the issue. Result: q=4 dropped from 40.9% → **6.4%**.

### Benchmark infrastructure

The benchmark (`benchmark_against_fortran_run.jl`) now reads `euler_stride` from `dcon.in` and uses it as `save_interval` in the generated `gpec.toml`, matching the Fortran ODE output density. Defaults to 1 if absent.

### Remaining errors

- **q=2/3**: The chord-slope derivative is less accurate than `ud_store` for smooth inner surfaces. Hermite cubic interpolation (using `ud_store`) gave 12%/7.6% vs the current 18.6%/14.3%, but caused q=4 to blow up to 40.9%. This is a fundamental tradeoff of interpolating from stored steps vs. Fortran's fresh ODE re-integration at each evaluation point.
- **q=5**: Both derivative methods give identical xsp1 estimates (ratio ~0.99), so the 50.7% error is in the value `xsp` rather than its derivative — likely accumulated ODE integration error near the plasma edge (ψ=0.988) compounded by extreme near-cancellation (xsp1 ≈ ±1600 vs ±14 for q=2).

All other outputs (Jbgradpsi, b_n, xi_n, Φ_x spectrum) match Fortran to <1%.

logan-nc:
@claude review

logan-nc:
## Status update — gpeq pipeline complete + R,Z,φ post-processing helper

Pushed `45727e52` extending the field reconstruction pipeline and adding an Analysis post-processing helper.

### Implemented
- **Full Fortran gpeq pipeline in Julia** — `reconstruct_physical_fields` now executes `gpeq_sol → gpeq_contra → gpeq_cova → gpeq_rzphi`, producing contravariant, covariant, Clebsch (PENTRC), regularized, and cylindrical (R,Z,φ) mode-space outputs.
- **`reg_spot` regularization** — added to `PerturbedEquilibriumControl` (default `0.05`) and exposed in the DIIID-like example TOML, plumbed through `Response.jl`/`FieldReconstruction.jl`.
- **Fourier-transform convention cleanup** — `1/N` normalization now lives in the forward transform (matches Fortran `iscdftf`); the complex forward uses `exp(-imθ)` so `forward(inverse(x)) = x`. `ResponseMatrices.jl`, `SingularCoupling.jl`, and `Vacuum.jl` updated accordingly (the vacuum response now multiplies by the explicit `4π²` Chance 2007 factor).
- **`Analysis.PerturbedEquilibriumAnalysis`** (new submodule) —
  - `modes_to_theta(h5_file, variable; mtheta, keep_sfl_phi)` reads the HDF5 modes, metadata, and ν spline self-sufficiently, returns `(npsi, mtheta, npert)` in SFL θ. With `keep_sfl_phi=false` it applies `exp(i·n·ν(ψ,θ))` and helicity conjugation to reproduce Fortran `gpout_xbrzphifun` machine-coordinate output.
  - `theta_to_thetaphi(theta_data, n_vals; nphi)` extends to a `(θ, φ)` grid via toroidal inverse DFT.
- **HDF5 outputs** — added `clebsch_*`, `xi_psi_J`, `xi_theta(_reg)`, `xi_zeta(_reg)`, `xi_cova_*`, `b_theta_reg`, `b_zeta_reg`, `b_cova_*`, `xi_R/Z/phi`, `b_R/Z/phi`, `psi_area`. Removed legacy `Jbgradpsi` key.
- **Benchmarks** — `benchmark_against_fortran_run.jl` no longer FFT-decomposes the Fortran R,Z,φ data. It loads the raw `(ψ, θ)` arrays from `gpec_profile_outputs_n1.nc`, runs Julia output through `modes_to_theta(...; keep_sfl_phi=false)`, and compares θ=0 real/imag profiles directly in a 6×2 grid (bold Fortran, thin Julia, y-limits clipped to the [2%, 98%] percentiles to suppress spikes).
- **EulerLagrange fix** — `save_interval` is now keyed off `odet.total_steps` instead of `odet.step` so saves happen at the intended cadence across resize/segment boundaries.

### Benchmark status (DIIID-like ideal example)
| Component | Status |
|---|---|
| ξ^ψ, ξ^θ, ξ^ζ (SFL) | well-benchmarked |
| b^ψ, b^θ, b^ζ (SFL) | well-benchmarked |
| ξ_Z, ξ_φ, b_Z, b_φ (cylindrical) | excellent agreement (thin Julia traces bold Fortran across ψ) |
| **ξ_R, b_R (cylindrical)** | **BETA — ~20% discrepancies vs Fortran** |

The R-component issue is the only remaining residual — both the T-matrix and the assembly formulas in `_compute_rzphi_modes` mirror `gpeq_rzphi`, but `Re(ξ_R)` differs from Fortran by a roughly constant ~1.2× scale, `Re(b_R)` deviates for ψ > 0.5, and `Im(b_R)` deviates near ψ ≈ 0.25. The other Z, φ projections agree, which strongly suggests a subtle issue specific to the R reconstruction that I haven't yet pinned down.

For now this is flagged with a `@warn` from `reconstruct_physical_fields` (one-shot per session) and a `!!! warning "Beta"` admonition on the `_compute_rzphi_modes` and `modes_to_theta` docstrings, so downstream consumers know not to rely on the R cylindrical projections yet.

### Suggested next steps
- Diagnose the R-component residual (likely worth a focused side-by-side trace of `t11`, `t12`, `t13` against `gpeq_rzphi`).
- Decide whether `xi_R/Z/phi` and `b_R/Z/phi` should ship in the HDF5 output now (gated by the beta warning) or be held back until R agrees.

logan-nc:
Here is the good agreement of modal displacements and b fields again:
image


Here is the (R,z,phi) projection, which has some mysterious ~20% error (especially in R):
image

logan-nc:
## Status Update — develop merged into feature branch

Merged **develop** (`2242f378`, +84 commits) into `feature/coil-forcing-terms`. Resolved 10 text conflicts + 1 add/add, then fixed a couple of latent bugs and API migrations the merge exposed. New commits pushed:

- `76fb48da` — `Merge - MERGE - Resolve develop conflicts for feature/coil-forcing-terms`
- `08b1b1d1` — `Regression Harness - NEW FEATURE - Track per-surface PE resonant flux Phi_res`

### What was in the merge

**Conflict resolutions** (all deliberate per-hunk, not wholesale):
- `src/Analysis/PerturbedEquilibrium.jl` (add/add) split into two modules — `PerturbedEquilibriumModes.jl` keeps the HEAD `modes_to_theta`/`theta_to_thetaphi` data helper, `PerturbedEquilibrium.jl` holds develop's plotting module.
- Develop's plotting functions updated to read HEAD's new HDF5 schema (complex datasets directly, 1D rational-surface vectors, `rational_psi`/`rational_q`/`rational_n` metadata).
- `EulerLagrange.jl` save condition combines HEAD's `total_steps` counter fix (#185) with develop's `in_edge_scan` guard.
- `ForcingTerms.jl` keeps HEAD's `normalization` tag scanning while adopting develop's `readdlm(...; ...)` style.
- `FieldReconstruction.jl`, `PerturbedEquilibriumStructs.jl`, `Response.jl`, `ResponseMatrices.jl`, `Utils.jl` — kept HEAD (develop copies referenced removed fields/functions from the pre-rewrite schema).

**FastInterpolations v0.2 → v0.4 migration** (develop bumped the dep; 10 callsites needed fixing):
- `CoilFourier.jl` (3 sites) — removed `LinearBinary` import + `search=LinearBinary()` kwargs, `deriv=1` → `deriv=DerivOp(1)`.
- `FieldReconstruction.jl` (8 sites) — `deriv=Val((0,1))` → `deriv=DerivOp(0, 1)` and similar.

**Collateral fixes in `ResponseMatrices.jl`**:
- Deleted dead `unpack_realimag_to_complex` function (never called, used an interleaved layout inconsistent with the actual `grri`/`grre` block layout at `Vacuum.jl:108-111`).
- Fixed `calc_surface_inductance` LU-fallback path (lines 441-444) to use the correct block layout `grri[k, i] + 1im * grri[k, i + mpert]` instead of the interleaved `grri[k, 2i-1] + 1im * grri[k, 2i]`. Latent bug, only reachable on LU failure.

**Housekeeping**:
- Renamed `PerturbedEquilibriumAnalysis` → `PerturbedEquilibriumModes` in `benchmarks/benchmark_against_fortran_run.jl`.
- Added `CoilForcing` + `PerturbedEquilibriumModes` autodocs sections to `docs/src/analysis.md`.

### Verification

- **Precompile + load**: clean.
- **Test suite**: 649 tests passing.
- **Examples**: all 3 `examples/*/analyze_example.jl` scripts produce `pe_summary.png` without errors.
- **Regression harness** (`regress --cases diiid_n1 --refs 2242f378,local`):
  - 21 core stability quantities unchanged (`et[1]`, `ep[1]`, `ev[1]`, energies all, `q0`, `q95`, β, singular ψ/q, m/npert, q/pressure checksums).
  - 5 explained differences:
    - `ODE steps (saved)`: expected from the #185 `save_interval` counter fix.
    - `delta_prime`, `resonant_flux`, `island_half_width`, `chirikov_parameter`: expected from the 2D `[numpert_total × msing]` → 1D rational-indexed schema change introduced upstream in this branch.

### Regression harness extension (`08b1b1d1`)

Added `resonant_flux_all` entry to `regression-harness/cases/diiid_n1.toml` — tracks the full `Vector{ComplexF64}` at `perturbed_equilibrium/singular_coupling/resonant_flux` via `all_complex` extraction, alongside the existing `||resonant flux||` norm scalar. This mirrors the `delta_prime` all-complex tracking already in place, so any per-rational-surface change in Φ_res is now caught element-wise rather than hidden inside the aggregate norm.

Verified: `regress --cases diiid_n1 --refs 76fb48da,local --force` shows the new `resonant flux Phi_res [4 elem]` row matching `0.0e+00` against the merge commit, 27/27 quantities OK.

Ready for re-review.

matt-pharr:
@logan-nc would you like eyes on this? I want this ready badly. Once this is in we have full gpec and 1 medium PR away from resistive gpec.

logan-nc:
Yes please, @matt-pharr! I just didn't have a moment to read the full diff through myself or assign it to someone yesterday. The intended scope is completed. I reached a little for the (r,z,phi) components and burned myself because they don't agree as well as the others... I don't see why they shouldn't, so this is a potential bug but my local claude reviews couldn't see anything wrong with the code in a careful comparison to fortran.

One open question I have is this: At some point in the development of this branch it was determined that the julia EL ode solver's saved dξ/dψ was riddled with huge spikes and no good, so we switched to a finite difference of ξ while fortran uses the EL solution directly for `xsp1mn`... now that the correct paths have been fully fleshed out for `xi_n`, clebsch, `b_*` etc. I wonder if this could be revisited for smoother results. I wonder if this is why `b_phi` is so oscillatory, but the size of the little steps in xi_R might mean this is an issue with some equilibrium grid based term and I haven't investigated enough to know the true source of these things.

All that said, everything other than the (r,z,phi) matches great. The Phi_res is close (I wish it was a little closer), and the (R,z,phi) components have appropriate warnings guarding them since they don't match as well as the rest... so I am ok with you merging after a human read through.

logan-nc:
@claude review

logan-nc:
Update: I found the source of why the ud_store derivatives from ForceFreeStates where dropped in favor of finite differencing. Here is a comparison of the finite differences, segmented splines (so they don't go through singular surfaces and get gibbs rining behavior) and the ud:
image


You can see "random" spikes in the `ud_store` derivatives... it turns out these are the steps directly after gaussian fixups, which do not properly modify ud after modifying u. This is a bug... I am thinking through whether there is an easy solution or if I'll just go with the spline derivatives here and open an issue for the `ud_store` bug to be solved in a separate PR.

logan-nc:
ud fix wasn't too bad in the end, it looks good now:
image

logan-nc:
## Status update: R,Z,φ field reconstruction fixes

### Changes in `caa068e1`

**1. Metric interpolation fix (FieldReconstruction.jl)**
Replaced `_find_closest_index` (nearest-neighbor snap) with `_build_metric_interp` (cubic spline interpolation) for Fourier coefficient evaluation in both `compute_contra_displacements` (Jacobian coefficients, quantities 7–8) and `compute_cova_components` (metric tensor coefficients, quantities 1–6). This matches Fortran's `cspline_eval(metric%cs, psi, 0)` which interpolates smoothly between grid points.

**2. Bug fixes**
- Fixed missing `nn` argument in `_compute_rzphi_modes` call sites (would have caused a method error)
- Fixed `equil.bt_sign` → `equil.params.bt_sign` (field lives on `EquilibriumParameters`, not `PlasmaEquilibrium`)
- Hoisted `R_fun`/`Z_fun`/`phi_fun` buffers out of the per-ψ loop (eliminates per-step allocation)
- Removed stale BETA warning from `reconstruct_physical_fields`

**3. Cleanup**
Removed `_find_closest_index` helper (no longer used). Benchmark's rzphi comparison now uses `modes_to_theta` consistently.

### Benchmark status

R,Z,φ profiles are now **smooth** (no oscillations or step artifacts):
- Previous oscillations in b_φ were caused by mode-space round-trip truncation losing high harmonics generated by geometry products — fixed by the metric interpolation improvement
- Previous step artifacts in ξ_R were caused by nearest-neighbor metric coefficient snapping — fixed by spline interpolation

A **systematic ~7–14% amplitude offset** remains in mode-space xi_ψ compared to Fortran. This is an upstream issue (exists before any R,Z,φ conversion) and is consistent with the ~2–5% P-matrix diagonal differences. This is a separate issue from the artifacts fixed here. Users needing higher-fidelity theta-space reconstruction can increase `mpert` or use `Analysis.PerturbedEquilibriumModes.modes_to_theta` post-hoc.

logan-nc:
Ok, @matt-pharr I am completely done with this branch. The results all look ok now, even for the (R,z,φ) components:
image


In the process of fixing the (R,z,φ) components, I definitely found some silly things claude had done though. I'd appreciate your read through - definitely don't be shy about calling out any funky choices. This whole branch was vibe coded between/after other work so I am sure there are opportunities to clean it up.

matt-pharr:
@logan-nc before I dive in here, what is the current status of this:

image

Claude said it was due to incorrect P matrix calculation? Is that accurate?

logan-nc:
that is indeed Claude's claim. I don't really buy it, since all the magnetic coordinate displacements above match more closely. Some higher mode magnetic displacements start to deviate at large psi, and I can believe that is from small differences in the P matrix. I don't see why that would result in this discrepancy in the core when converting to radial displacement though. My Claude was assigned to triple check it and came up empty handed, but watch out for anything suspect when you read through. Feel free to assign another ai to try and hunt down the reason.

logan-nc:
@matt-pharr  I was able to get slightly better agreement using more m:
image

logan-nc:
## Re: residual Re(ξ_R) core-region offset

@matt-pharr — re-audited the ξ_R pipeline end-to-end and ran a convergence test rather than just relitigating the formula.

### Formula audit — no simple mistake

Verified against `gpout.f:5180-5252` (`gpout_xbrzphifun`) and `gpeq.f`:

- The t-matrix elements in `_compute_rzphi_modes` (`FieldReconstruction.jl:852-864`) match Fortran element-by-element. `v*` coefficients correctly use **minor radius** `rfac = √(r²)` for the η-direction arc length; the **major radius** `R = R0 + rfac·cos(η)` appears **only** in `t33` (the φ-component). No minor/major swap, no missing Jacobian, no basis-vector mixup.
- Inputs (`xwp_modes`, `xmt_modes`, `xvz_modes`) map slot-for-slot to Fortran's `xwp_mn`, `xmt_mn`, `xvz_mn`.
- The SFL-ζ → machine-φ transform (ν-phase rotation + helicity conjugation) is applied via `Analysis.PerturbedEquilibriumModes.modes_to_theta(..., keep_sfl_phi=false)`, matching Fortran's `EXP(i·nn·dphi)` + `CONJG` (when helicity>0) at `gpout.f:5236,5245-5252`. The benchmark loader uses the same flag, so we're in the same machine-φ frame.

### Convergence test — points at Fourier-truncation aliasing

Julia's `_compute_rzphi_modes` does something Fortran's `gpout_xbrzphifun` does not: it multiplies `t(θ)·ξ(θ)` pointwise in θ, then forward-DFTs back to mode space truncated at `mpert`, before the benchmark loader re-expands to θ. Fortran constructs ξ_R(θ) directly from the modes in real space — no roundtrip. `t11`/`t21` mix `v(i,1)`/`v(i,2)` in different cos/sin combinations, so aliasing loss at the `mpert` boundary is asymmetric between ξ_R and ξ_Z — consistent with ξ_R showing the larger offset.

Two ablations on the DIII-D benchmark (Fortran reference unchanged, mpert=34, mtheta=256):

| Julia run | Re(ξ_R) RMS err | % of \|ξ_R_Fortran\| |
|---|---|---|
| baseline (mpert=34, internal mtheta=512) | 2.77e-4 m | **7.31 %** |
| internal mtheta → 4096, mpert unchanged | 2.77e-4 m | 7.31 % (unchanged) |
| `delta_mlow=delta_mhigh=16` (mpert=50), mtheta=512 | 1.48e-4 m | **3.91 %** |

Refining the internal θ-grid inside `_compute_rzphi_modes` alone has ~zero effect — internal aliasing is not the cause. Expanding the mode-space basis (bigger `mpert`) on the Julia side cuts the RMS error nearly in half, with the θ=0 profile tracking Fortran much more closely in the core. This is consistent with mpert-truncation aliasing in the Julia roundtrip.

Convergence plot: `benchmarks/fortran_vs_julia_DIIID_ideal_example/xi_R_aliasing_convergence.png`.

**Caveat — matched-mpert test still owed.** Step 2 only raised mpert on the Julia side; Fortran was held fixed at mpert=34. So this proves Julia's pipeline is mpert-sensitive (aliasing is involved) but does *not* prove Julia would match Fortran exactly at the same expanded mpert. The clean follow-up is to also re-run Fortran at `delta_mlow=delta_mhigh=16, mtheta=512` and recompare like-for-like. Happy to do that next if folks want it before merging.

### Verdict

The formula is correct. The residual ξ_R offset is mpert-sensitive on the Julia side and consistent with truncation aliasing introduced by the mode-space roundtrip in `_compute_rzphi_modes`. Two follow-ups available:

1. **Raise the default `delta_mlow`/`delta_mhigh`** in the DIII-D ideal example (8 → 16) so the reference benchmark shows tighter agreement out of the box.
2. **Document the aliasing trade-off** more prominently in `_compute_rzphi_modes`' docstring (currently noted at `FieldReconstruction.jl:791-793`).

Preference?

logan-nc:
The key is that fortran just calculates xi_R in theta space. In julia, I don't want to store a mix of m and theta space outputs, so I convert everything back to m. But the roundtrip of theta → m → theta-for-plotting Fourier transforms end up with some aliasing. Good to know, but fine as far as the PR feature validation is concerned.

I think this is good to go once you review it

logan-nc:
@claude review this PR. Keep it short.

logan-nc:
One more human look and then merge @matt-pharr

examples/DIIID-like_ideal_example/forcing.dat
```diff
@@ -1,4 +1,5 @@
 # Forcing data for perturbed equilibrium calculations
+# normalization: normal_field_T
 # Format: n  m  real(amplitude)  imag(amplitude)
 # Single mode test case: n=1, m=6, amplitude=1e-4
 1  6  1e-4  0.0
```
matt-pharr: Is this necessary anymore or can we just have it set up to use C coils at 1kA n=1 current?

examples/DIIID-like_ideal_example/gpec.toml
```diff
@@ -65,3 +65,4 @@ compute_response = true                 # Compute plasma response to forcing
 compute_singular_coupling = true        # Compute singular layer coupling metrics
 verbose = true                          # Enable verbose logging
 write_outputs_to_HDF5 = true            # Write perturbed equilibrium outputs to HDF5
+reg_spot = 0.05                         # Regularization width for singular surfaces (0 = disabled)
```
matt-pharr, line 72: Should this really be on by default?

src/Equilibrium/AnalyticEquilibrium.jl
```diff
@@ -301,5 +301,5 @@ function sol_run(equil_inputs::EquilibriumConfig, sol_inputs::SolovevConfig)
     # Print out equilibrium info
     @info "Generating Solovev equilibrium: mr=$mr, mz=$mz, ma=$ma, e=$(@sprintf("%.3f", e)), a=$(@sprintf("%.3f", a)), r0=$(@sprintf("%.3f", r0)), q0=$(@sprintf("%.3f", q0))"
 
-    return DirectRunInput(equil_inputs, sq_in, psi_in, psi_in_xs, psi_in_ys, rmin, rmax, zmin, zmax, psio)
+    return DirectRunInput(equil_inputs, sq_in, psi_in, psi_in_xs, psi_in_ys, rmin, rmax, zmin, zmax, psio, 1)
```
matt-pharr, line 306: Why does this have bt_sign but not ip_sign?

docs/src/forcing_terms.md
```diff
@@ -1,13 +1,111 @@
...
+The coil Biot-Savart pipeline produces mode amplitudes directly equal to Fortran's `Phi_x`:
+
+```
+Julia output ≡ Fortran Phi_x
+```
+
+The formula is `Phi_x(θ_norm, ζ_norm) = 2π × R × (B_R ∂Z/∂θ_norm − B_Z ∂R/∂θ_norm)`, where
+the `2π` comes from the toroidal Jacobian ∂r/∂ζ_norm = 2π·R·ê_φ.
+
+Users who provide forcing files may use the more intuitive **2π-angle Fourier amplitudes**
```
matt-pharr, line 30: I am approving this for now but I think that we should do a more formal docs pass as we write the julia conversion paper

src/Utilities/FourierTransforms.jl
```diff
@@ -260,23 +260,23 @@ modes = ft(data)       # Matrix{ComplexF64} of size (40, 10)
...
 """
 function (ft::FourierTransform)(data::AbstractVecOrMat{<:Real})
-    # Forward transform using matrix multiplication
-    # real_part = data^T * cslth  (or cslth^T * data for vectors)
-    # imag_part = data^T * snlth  (or snlth^T * data for vectors)
-    # Return as complex
+    # Forward transform with 1/N normalization and exp(+imθ) convention for real data.
+    # Note: the real forward uses exp(+imθ) (not exp(-imθ)) because ResponseMatrices and
+    # SingularCoupling are self-consistent in this convention. The complex forward uses
+    # exp(-imθ) to enable correct inverse→forward round-trips for _compute_rzphi_modes.
```
matt-pharr: I don't like this. This seems like it will confuse the heck out of us in the future. Is there a reason we didn't fix this now? @logan-nc

src/PerturbedEquilibrium/Utils.jl
```diff
@@ -85,61 +92,112 @@ perturbed_equilibrium/
...
+
+        # Metadata [n_rational]
+        !isempty(state.rational_psi)       && (coupling_group["rational_psi"]       = state.rational_psi)
+        !isempty(state.rational_q)         && (coupling_group["rational_q"]         = state.rational_q)
+        !isempty(state.rational_m_res)     && (coupling_group["rational_m_res"]     = state.rational_m_res)
+        !isempty(state.rational_n)         && (coupling_group["rational_n"]         = state.rational_n)
+
+        # Energies
         energy_group = haskey(pe_group, "energies") ? pe_group["energies"] : create_group(pe_group, "energies")
         energy_group["plasma_energy"] = state.plasma_energy
```
matt-pharr: These are still empty are they not? Looking at hdf5s it has zeros. They should be NaNs as should all placeholder outputs until we fill them in

src/PerturbedEquilibrium/FieldReconstruction.jl
```diff
@@ -3,319 +3,895 @@ Field reconstruction from eigenmode response.
...
+        ffit.amats(view(amat, :), psi_norm; hint=hint)
+        ffit.bmats(view(bmat, :), psi_norm; hint=hint)
+        ffit.cmats(view(cmat_buf, :), psi_norm; hint=hint)
+
+        # xms = -(A\B)*xmp1 - (A\C)*xsp
+        xsp_vec = view(xi_psi_modes, ipsi, :)
+        mul!(xms_vec, bmat, xmp1_vec)                     # xms = B*xmp1
+        mul!(xms_vec, cmat_buf, xsp_vec, 1.0+0.0im, 1.0+0.0im)  # xms += C*xsp
+        # Solve A*result = xms for result, then negate
+        amat_fact = cholesky(Hermitian(amat, :L))
```
matt-pharr, line 380: cholesky = positive definite right? But this only enforces hermitian? Could we get an error here if hermitian(amat) had a negative eigenvalue?

matt-pharr (changes requested):
Reviewed and looks good but we need to go through and do a huge docstring sweep in the near future once we get over the backlog of PRs. Will take a lot of AI-free human hours for me to be confident in things like units broadly. Please look at the fourier transform comment especially and decide if that needs to be done on this PR. Otherwise I am excited to get this merged!

examples/DIIID-like_ideal_example/forcing.dat
logan-nc: Can you expand on what you mean? Possible layers here:
- The manual mode table interface? Yes, it's absolutely necessary. This replaces the cosmn(2,1)=1e-4 type namelist interface of fortran gpec. It is a great, intuitive catch all for interfacing with any other code (example: TokaMaker). We just need a table of Fourier modes and we've off.
- The normalization? Also, yes. It is unreasonable to expect all other codes (or people just trying to put in intuitive numbers) to get the half-area weighting done correctly. We need to be able to accept the more common b_psi or b_n and such (in addition to accepting the btilde_psi, if someone has it).
- Using this table interface for the example? Not necessary. I'd agree to switching to a coil forcing term. Is this what you mean @matt-pharr? Do you want to do it or should I?

examples/DIIID-like_ideal_example/gpec.toml
logan-nc: Yes, this is the default in fortran and we should maintain that for continuity. It is absolutely necessary for KineticForces (see [[Berkery PoP 2018](http://aip.scitation.org/doi/10.1063/1.4873894)]). The difference between Julia and Fortran is that here we do not use a separate `reg_flag` but more efficiently and intuitively just do not regularize when `reg_spot=0`.

src/Equilibrium/AnalyticEquilibrium.jl
logan-nc: IDK off the top of my head... this warrants a closer look

docs/src/forcing_terms.md
logan-nc: Agreed! We should wait to do it after closing the various variable-rename and struct-argument-breakup issues are resolved. With that, we'll basically be pushing towards a version 0.1.0 release

src/Utilities/FourierTransforms.jl
logan-nc: Hmmmm I don't like this either. I remember the complex round trip being a bug that it took claude a long time to find in this branches development. My guess is that when found, claude just fixed what it needed to and left the real one so as not to accidentally break other things. This warrants a closer look to see (a) if the real round trip works out (b) if it should be changed for consistency than what would it impact and (c) is it this way because of some funky fortran legacy? If (c) then we need to make sure we document the new, more consistent Julia convention in the annotations.

src/PerturbedEquilibrium/Utils.jl
logan-nc: We should just calculate them! Now that we've benchmarked the fundamental PerturbedEquilibrium matrices it should be a trivial add of 1 line each. I'll do this

logan-nc:
## Review comments addressed

Pushed 7 commits (`8ad0d4b0`…`50db9cd5`) addressing the open review threads.

### `Utils.jl` — empty placeholder energies → **computed**
`PerturbedEquilibriumState` energies were zeros. They are now computed and written, ported from Fortran `gpout` surface-inductance energy block: `vacuum_energy`, `surface_energy`, `plasma_energy`, `toroidal_torque` (verified against the Fortran `fortran-physics-reviewer`). Note: Fortran has **no** separate "total energy" — its netcdf emits only `energy_vacuum/surface/plasma`, and "total energy" is just the long_name of the plasma quantity. So the redundant `total_energy` field/dataset was dropped rather than kept as a duplicate. (`8ad0d4b0`, `50db9cd5`)

### `FourierTransforms.jl` — sign-convention split → **unified**
The real-input forward used `exp(+imθ)` while the complex-input forward used `exp(-imθ)`. Both are now unified to `exp(-imθ)` (Fortran `iscdftf`); `inverse(ft, ft(x))` round-trips for real and complex input alike. The split was *not* arbitrary — tracing `gpvacuum_flxsurf` showed Fortran indexes the Green's function at the reversed θ point (VACUUM θ runs opposite to DCON θ), and the old `exp(+imθ)` real-forward silently absorbed that reversal. The reversal is now **explicit** at the one consumer (`compute_surface_inductance_from_greens`) via a `_reverse_theta` helper. The Fortran benchmark is unchanged by the refactor (resonant-surface agreement holds at the prior baseline). (`bc321504`)

> Side note: a stricter port that reverses only the *un-phased* Green's function (phase applied forward, per `gpvacuum.f`) was tried — it sharply improved the q=5 surface (50.6%→4.3%) but regressed q=2 (14.6%→97%). Mixed, so not merged; it's a real lead for whoever digs into the residual resonant-coupling discrepancy.

### `FieldReconstruction.jl` — cholesky on Hermitian `amat`
`amat` is positive-definite by construction (the Newcomb kinetic-energy quadratic form), so `cholesky` is safe — same factorization already used unconditionally in the ODE setup. Documented in-place; no code change needed. (`d1ef5c95`)

### `AnalyticEquilibrium.jl` — `bt_sign` vs `ip_sign`
No `ip_sign` field is needed: `sign(crnt)` is recovered from `params.crnt` downstream for all equilibrium types. The Solovev `bt_sign=+1` literal is now annotated. (`d1ef5c95`)

### DIII-D example → C-coil forcing
The DIIID-like example now drives the DIII-D C-coil through the Biot-Savart pipeline; the manual mode-table interface is retained as a worked example on the Solovev case (`n=1, m=2`). Each forcing workflow now has one canonical example. (`d1ef5c95`)

### Also included
- **`Benchmark`** (`c64417e6`) — `benchmark_against_fortran_run.jl` updated for the current `ForceFreeStatesControl` API (dead `dmlim` keys removed) and now emits `qhigh` to match the Fortran `sas_flag` truncation.
- **`EQUIL` bug fix** (`063f3895`) — `equilibrium_gse!` snaps the repeated periodic θ endpoint exactly equal to the start; `flux_fsx`/`source` accumulate ~1 ulp of round-off that tripped `PeriodicBC`'s strict endpoint check on some equilibria.

Deferred (per prior thread): the `reg_spot` default is intentional, and the formal docs pass is for the 0.1.0 release prep.

logan-nc:
@matt-pharr please see if you are happy with these attempts to address your comments and update the review status accordingly.

matt-pharr (approved):

PR #197, opened 2026-03-22
EQUIL/VAC - Equal-arc vacuum boundary integration

logan-nc:
## Summary

- **Problem**: With \`jac_type = \"hamada\"\` and \`psihigh → 1.0\`, quadrature points cluster near the X-point (where Bp → 0), leaving the midplane and inboard side undersampled. This degrades the vacuum Green's function boundary integral accuracy.
- **Fix**: Perform the vacuum boundary integral on an equal-arc (uniform physical spacing) surface, then transform the Green's function back to jac_type mode coefficients via a DFT-style mode transformation matrix T (G_jac = T · G_eq · T†).
- **Backward compatible**: off by default (\`use_equal_arc_vacuum = false\`); all existing tests pass.
- **Known limitation**: \`use_equal_arc_vacuum = true\` is incompatible with \`PerturbedEquilibrium\` runs — \`grri\`/\`grre\` are not transformed and a hard \`error()\` guards against this. Fix is planned.

## Changes

- \`PlasmaEquilibrium{P,I2D,R}\` gains \`raw_profile::R\` field — stores \`DirectRunInput\` for EFIT/CHEASE/analytic paths, \`nothing\` for inverse-method equilibria
- \`_arclength_fieldline_sol\` helper — runs arc-length ODE and returns \`(sol, bfield)\` so \`sol.t\` (true arc-length) is available without chord-length reconstruction
- \`equal_arc_fieldline_output(equil, psifac; mtheta_eq)\` — interpolates to uniform arc-length grid; \`nu_eq\` formula corrected to \`F*(col4 − θ_jac*col4_total)\` matching equilibrium_solver
- \`jac_to_eq_mode_transform(theta_jac_eq, ...)\` — mode transformation matrix mapping equal-arc → jac_type Fourier coefficients
- \`ForceFreeStatesControl\` gains \`use_equal_arc_vacuum::Bool = false\` and \`vac_eq_arc_theta_factor::Int = 4\`
- \`_free_run_equal_arc_vacuum!\` — computes wv on equal-arc surface and transforms back block-by-block via \`mul!\` (avoids building dense T_full); closes loop before \`VacuumInput\` to satisfy inclusive \`PeriodicBC\`; caps mode space expansion at 8× with \`@warn\`
- \`free_compute_wv_spline\` — documented that it always uses the standard (jac-type) path regardless of \`use_equal_arc_vacuum\`
- PE incompatibility guard — \`error()\` at config load time if \`use_equal_arc_vacuum = true\` with a \`[PerturbedEquilibrium]\` section present (skipped if either \`force_termination = true\`)

## Test plan

- [x] \`runtests_fullruns.jl\` — all 2 tests pass (bit-identical to previous, \`use_equal_arc_vacuum = false\` default)
- [x] \`norm(T - I) < 1e-10\` for \`jac_type = \"equal_arc\"\` — **PASS** (norm = 6.68e-15)
- [x] \`nu_eq\` matches \`rzphi_nu\` within ~0.1% — **PASS** (0.008% core, 0.12% edge)
- [x] psihigh convergence scan — equal-arc vacuum tracks standard path for psihigh ≤ 0.997; at psihigh = 0.999, X-point clustering cap fires (max_rate = 9.9 > cap 8.0) and equal-arc diverges — further work needed for Hamada near separatrix


logan-nc:
@claude review this

logan-nc:
@jhalpern30 have a think about the big picture ideas here before bothering with line by line reviews. 

My initial development here was meant to satisfy a nagging quandary I've had about the usefulness of changing jacobians in the far edge. I know some of the IPP codes called that out as part of their secret sauce, and it definitely doesn't seem good to have your whole grid condense to the x-point if pushing out towards psihigh→1.0 in magnetic coordinates like Hamada. In the end, I didn't love that approach. Varying the jacobian made for some odd swooping theta contours (arcing sharply to the xpoint, then turning around to get back up to equal_arc) in the far edge. It gets super swoopy if you push further and further out, and the answers were different from pure equal_arc runs by up to 5%. Some attempts to converge with scans of m or mtheta failed - making me worried about equal_arc in general. 

Long story short, I pivoted to just doing the vacuum greens function integrals in equal arc since those are in real space rather than mode space and the grid needs to actually cover the surface - otherwise I am pretty sure the finite integration does funky things. This is a pretty straightforward idea - check the minimum arc length of the user's jacobian at the edge, then make an equal arc grid with that all the way around to do the greens function integrations on... but as the Claude review above points out, it can quickly make for huge grids 🤔 ... So I welcome any ideas as to how best to do this. If your new numerical approach makes this moot, then just let me know! I *think* you still had to integrate in real space so this still applies. Correct?

logan-nc:
Here is a figure showing how the splines get modified in the Jacobian blending approach (using `psihigh=0.999`):

image

image

image

The profile splines look unchanged, as they should.

jhalpern30:
@logan-nc my comments on this, most of which we discussed during our 1on1 a week or so ago. 

I think this is potentially useful, but probably not in this form. There is likely some desirable measure of anisotropy in the grid that is beneficial to resolve sharp features in the boundary integral such as x-points, so I don't think equal arc grid is the best choice for convergence either. Although I do agree that the Hamada coordinate limit as psi -> 1 is also not good. I think likely it should be some middleground. Intuitively, something that spaces based on curvature or something of that sort. 

So at some point, I think switching the vacuum grids to a more intelligent choice will be beneficial for convergence. i.e. form the boundary integral operator in some favorable coordinate system that well-resolves the operator, specifically in regions where small wavelength perturbations are prevalent such as the x-point. 

Before merging anything like this into the code, I propose the following must be done:
1. Benchmark the convergence in mtheta of some of the sfl coordinate systems, specifically hamada and maybe pest at various psihigh. I'd like to see how badly the convergence is actually affected. And if there is some nice middleground in the anisotropy of the hamada coordinates where we see the best convergence.
2. Use this to implement a new coordinate system or determine that a new one is not needed. I am not convinced its either sfl or equal arc - I think it might be some curvature based choice, or we should even just stack near x-point.
3. Benchmark this new convergence, hopefully now independent of the jacobian/psihigh, demonstrating that it is actually an improvement

jhalpern30:
@logan-nc considering that this was the method that Michail said ELITE's vacuum code was using and he switched to SFL and got better results, I think we can close this if that works with you.

logan-nc:
Closing this PR for now to keep the active PR list clean. The discussion and commits are preserved here for historical reference, but I’m deleting the branch to keep the branch list manageable. If this becomes relevant again, we can reopen or start a fresh branch/PR from this context.

We can resume later, if needed, by (a) reopening the PR and clicking `Restore branch` if GitHub offers it or (b) manually using:
```
git fetch origin
git checkout -b edge_jacobian 5277bc5
git push -u origin edge_jacobian
```

PR #202, opened 2026-03-24
IMAS - Add IMAS integration: read/write equilibrium via IMASdd

### LINKED ISSUE #138: Adding IMAS capabilities to GPEC

calin1989:
## Summary

- Add \`read_imas()\` to load equilibrium from an IMAS data dictionary with COCOS 11/2 conversion (\`imas_cocos\` config field in \`EquilibriumConfig\`)
- Add \`write_imas()\` to store stability results in \`mhd_linear\` IDS — one \`toroidal_mode\` entry per n_tor, storing the least-stable δW for that n-block
- Add \`dd\` keyword argument to \`main()\` for IMAS-driven workflows
- Add \`n_tor_idx\` field to \`VacuumData\` and populate it via \`argmax(eigenvector)\` in \`Free.jl\` to correctly identify which n-block each sorted eigenvalue belongs to

## Response to review comments (jhalpern30)

**1. EFIT dependency**
Removed EFIT from \`Project.toml\` entirely. \`using EFIT\` now lives only in the benchmark example script, with a note requiring the user to install it separately. The g-file path in the benchmark uses GPEC's own \`eq_type="efit"\` reader.

**2. \`sq_in\` spline (Series wrapper, PR #207)**
Already using the correct API: \`cubic_interp(sq_xs, Series(sq_fs_nodes); extrap=ExtendExtrap())\`. No \`bc=CubicFit()\`. Written fresh on top of develop after PR #207 was merged.

**3. \`psi_in\` spline (search=LinearBinary(), PR #207)**
Already removed. Line is \`cubic_interp((psi_in_xs, psi_in_ys), psi_proc; extrap=ExtendExtrap())\`. \`LinearBinary\` is now the default and not specified.

**4. \`jpec_geqdsk_to_imas!\` export**
Not present anywhere in the source. The function was dropped entirely — conversion is handled by \`EFIT.geqdsk2imas!\` in the benchmark script where it belongs.

**5. Unit tests (cost and scope)**
\`test/runtests_imas.jl\` uses mock data only — no GPEC ODE runs. Runs in milliseconds. Tests only \`read_imas\` and \`write_imas\`. The full IMAS vs g-file benchmark is in \`examples/\`, not the test suite.

**6. \`write_imas\` multi-n correctness**
Fixed. Now writes one \`toroidal_mode\` entry per n_tor with \`energy_perturbed = minimum(real(et))\` for eigenvalues in that n-block. \`n_tor_idx[i]\` (set via \`argmax\` on each eigenvector in \`Free.jl\`) identifies the n-block regardless of how LAPACK orders the columns. Tests 5 and 6 in \`runtests_imas.jl\` verify: (a) interleaved n=1/n=2 eigenvalues are routed correctly, and (b) a combined n=1+2 run produces the same per-n δW as running n=1 and n=2 individually.

## Tests

\`test/runtests_imas.jl\` — mock-data unit tests only (no ODE runs):
- **Tests 1–3** (\`read_imas\`): COCOS 11→2 conversion, COCOS 2 passthrough, spline callability
- **Tests 4–6** (\`write_imas\`): metadata, single-n, multi-n routing, multi-n consistency with individual runs

## Benchmark

\`examples/DIIID-like_ideal_example_IMAS/run_imas_example.jl\` — run manually, requires user-installed EFIT.jl:
- Converts the DIII-D-like g-file to IMAS via \`EFIT.geqdsk2imas!\`
- Runs GPEC from IMAS dd and from g-file with identical \`ForceFreeStates\`/\`Wall\` settings
- Prints δW(n=1) for both paths and relative difference


logan-nc:
@claude review

logan-nc:
haha. This PR broke the claude bot!

logan-nc:
@calin1989 I think the claude review (after it helucinated for a while - I've never seen that before!) brings up some good points. Try to work through them this week. Don't worry about item (3) - I think we'll be using IMAS enough that its fine to have as a hard requirement

calin1989:
Hi Nik! Yes, I've read it as well and it seems reasonable. I'll work on
those points this week. For now, I am focusing on FUSE for the meeting
tmrw.

Project.toml
```diff
@@ -34,10 +35,12 @@ Contour = "0.6.3"
 DelimitedFiles = "1.9.1"
 DiffEqCallbacks = "4.9.0"
 Documenter = "1.14.1"
+EFIT = ">=1.2.5, <2"
```
jhalpern30: I support adding IMASdd as a dependency, but I don't think we need EFIT. The only way it is used in this PR is for `readg` and `geqdsk2imas`. We already have our own g file reader, and the conversion to imas is only needed for your testing purposes. I propose you add one benchmarking script in this PR in examples/DIIID-like_ideal_example_IMAS that someone can run and verify GPEC running from both imas and geqdsk produce the same results. In that script, you can add a using EFIT statement, requiring the user to have it loaded in on their own end without making it a dependency of our project

src/Equilibrium/ReadEquilibrium.jl
```diff
@@ -381,3 +381,153 @@ function read_chease_ascii(config::EquilibriumConfig)
...
+    # abs(f_1d): F(ψ) can be negative depending on toroidal field direction convention;
+    #            take absolute value since GPEC uses magnitude F = R·|Bt|
+    sq_fs_nodes = hcat(
+        abs.(f_1d),
+        max.(p_1d .* mu0, 0.0),
+        q_1d,
+        sqrt.(psi_norm_grid)
+    )
+    sq_xs = collect(psi_norm_grid)
+    sq_in = cubic_interp(sq_xs, sq_fs_nodes; bc=CubicFit(), extrap=ExtendExtrap())
```
jhalpern30: This will probably need a small update based on Min-Gu's recent PR #207 - likely just a wrap with Series(...) but double check

src/Equilibrium/ReadEquilibrium.jl
```diff
@@ -381,3 +381,153 @@ function read_chease_ascii(config::EquilibriumConfig)
...
+        psi_proc .*= -1.0
+    end
+
+    rmin, rmax = extrema(r_grid)
+    zmin, zmax = extrema(z_grid)
+
+    psi_in_xs = collect(r_grid)
+    psi_in_ys = collect(z_grid)
+
+    psi_in = cubic_interp((psi_in_xs, psi_in_ys), psi_proc; search=LinearBinary(),
```
jhalpern30: Same thing here - checkout #207. If I remember right, there are the defaults and so this line can be simplified

src/Equilibrium/Equilibrium.jl
```diff
@@ -20,7 +21,7 @@ include("InverseEquilibrium.jl")
 include("AnalyticEquilibrium.jl")
 
 # --- Expose types and functions to the user ---
-export setup_equilibrium, EquilibriumConfig, PlasmaEquilibrium, EquilibriumParameters, ProfileSplines
+export setup_equilibrium, EquilibriumConfig, PlasmaEquilibrium, EquilibriumParameters, ProfileSplines, jpec_geqdsk_to_imas!
```
jhalpern30: 1. Rename this gpec instead of jpec
2. Is this actually needed? I don't see this used in the Equilibrium module, only the test suite. If its only needed as a verification, it can stay local to your computer, you can add it to a benchmarking script, or can go in the Analysis module

test/runtests.jl
```diff
@@ -26,4 +26,16 @@ else
     include("./runtests_eulerlagrange.jl")
     include("./runtests_sing.jl")
     include("./runtests_fullruns.jl")
+
+    # IMAS integration tests (require EFIT from [extras])
+    try
```
jhalpern30: These are definitely not all needed. Unit tests should only test new functionalities localized to the code, and in small chunks. (Remember these are run every time someone pushes their commits to a PR on github, so we don't want to be adding unnecessary, expensive calculations). Based on your added code, I'd recommend having a unit test testing `read_imas` and `write_imas`, but that's probably it. You can do a script in either a new examples folder for DIIID imas or in benchmarks documenting that you get the same results for a geqdsk run and from an IMAS converted from a geqdsk

src/GeneralizedPerturbedEquilibrium.jl
```diff
@@ -458,6 +460,44 @@ function write_outputs_to_HDF5(
...
+    mhd.ideal_flag = 1  # ideal MHD
+    mhd.code.name = "GPEC"
+    mhd.time = [t]
+
+    resize!(mhd.time_slice, 1)
+    ts = mhd.time_slice[1]
+    ts.time = t
+
+    resize!(ts.toroidal_mode, intr.numpert_total)
+    for i in 1:intr.numpert_total
```
jhalpern30: Have you benchmarked this with multiple toroidal modes? I added a feature to be able to run ForceFreeStates with multiple toroidal modes at once and I want to make sure IMAS is handling that correctly here. You should check that when you run with multi-n (let's say n=1, 2), it is writing the correct data. For example, it should be writing the least stable mode for each mode number - I think since we reorder them by stability and not mode number in `Free.jl`, you don't necessarily know whether the first mode is n = 1 or 2. You should make sure a n = 1 + 2 run writes two energy_perturbed values to the IMAS and that each n_tor's energy perturbed matches that when you run with n = 1 and n = 2 individually. (This assumes I am understanding https://gafusion.github.io/omas/schema/schema_mhd%20linear.html correctly based on a quick read)

jhalpern30 (changes requested):
@calin1989 nice work here! Here are some comments to address before we merge this in. The overview is
1. Consolidate your benchmarks into a single script that isn't run in the unit test folder, add simple unit tests. Remove any code/dependcies that isn't explicitly needed for a GPEC run from the source code
2. Either fix or prove me wrong on how you are writing energy_perturbed, especially for multiple toroidal mode cases
3. Update the splines based on Min-Gu's recent PR #207 
These fixes should actually reduce the line additions drastically here - your final PR will probably look like a new IMAS example showing the workflow, the updates read/write imas routines in the source code, and a few unit tests. Please have a look over these before our meeting tomorrow and prepare any questions on what isn't clear (but definitely don't feel like you have to address them before then)

calin1989:
Hi Jake, thank you for the feedback, I will solve the issues asap.

calin1989:
Hi Jake everything is now on. Please let me know if something is unclear, or not right. I'll be resuming the FUSE actor tests now

calin1989:
@mgyoo86 So I have a problem with a test runtests lts - ubuntu-latestExpected — Waiting for status to be reported. I am currently using Julia 11 version to support IMAS 8. The lts are for Julia 10. Do you have any advice on how I can make this test to pass. Thank you!!  ( All the other tests are running good).

mgyoo86:
@calin1989 
This doesn't look like a Julia/CI issue, but a GitHub repo settings issue. It seems that `runtests lts` is pinned as a **required check** in the branch protection settings for develop. 

I guess @jhalpern30 or anyone with admin access should be able to update the setting, and this PR should pass afterward.

jhalpern30:
> @calin1989 This doesn't look like a Julia/CI issue, but a GitHub repo settings issue. It seems that `runtests lts` is pinned as a **required check** in the branch protection settings for develop.
> 
> I guess @jhalpern30 or anyone with admin access should be able to update the setting, and this PR should pass afterward.

Thanks @mgyoo86!

@matt-pharr I believe you set up the Github test workflow. Any thoughts on the best way to structure this fix?

logan-nc:
@mgyoo86 I believe git blame says you made the test. What is the point of having it if it isn't enforced or checked in any way? I would want to understand the original motivation of the test before clobbering it just to pass a pr. Is it's very purpose to prevent people from implementing changes that break long term support? 

Note, the project is still so new that I would agree with an argument that there is no such thing as long term support at this point in the development stage... but I want to know why this was added before we delete it.

mgyoo86:
@logan-nc

Yes, I added the `lts` matrix entry since the julia LTS is generally a good reference for backward compatibility.
Up to now, GPEC has been able to support it since all dependencies (and GPEC itself) worked on it.
So the lts CI test was a good guard against silently breaking LTS support.

However, from this PR, it is no longer the case, because the new dependency `IMASdd` (its latest version) only supports julia `>= 1.11`.
So dropping LTS is kind of unavoidable if GPEC wants to use `IMASdd` as a dependency, and I think that's a fair reason to do so on GPEC side.

Ideally, general-purpose libraries (like FastInterpolations.jl) should try to keep LTS support as long as they can, since many downstream users rely on them.
However, this is much harder for end-product packages like GPEC or FUSE.
They rely on a long chain of upstream dependencies and inherit whatever minimum Julia version any one of those requires. 
Also, most of the effort usaully focuses on implementing the physics or new features with better speed and efficiency, which sometimes need the latest language features.
This is why IMASdd has moved on to support julia 1.11+, instead of maintaining the LTS compatibility.

jhalpern30 (approved):
@calin1989 looks great! I think this is ready to merge once we resolve the Github test issue. I just committed small changes in some of the long comment blocks for brevity (and got hit by the formatter on some random, unrelated things in ForceFreeStatesStructs, don't worry about those).

logan-nc:
Thanks for the clarification @mgyoo86! Your framing it in terms of general vs end-product packages was very helpful. I think I get it now, and agree we are ok ditching lts. I've removed it from the test.yaml and from the github branch protection rules. 

Great work @calin1989 - since @jhalpern30 has worked through everything with you, I've signed off and this will automatically merge once the tests all pass (all being rerun now since I merged the lts test removal commits).

PR #206, opened 2026-03-27
FFS: Frobenius axis initialization via Glasser 2016 Eq. 51 (free magnetic axis)

logan-nc:
## Summary

This PR replaces the original "fixed axis" initialization (`[U₁, U₂] = [0, I]`) with a physically correct Frobenius eigenvalue initialization at the magnetic axis, based on Glasser (2016) §VI Eq. 51. The new default (`fixed_axis=false`) correctly initializes the full space of N regular Frobenius solutions for all modes, including the degenerate m=0 case.

A `fixed_axis=true` option is retained in `ForceFreeStatesControl` (default: `false`) for comparison with the original Glasser initialization.

---

## Investigation Background

A parametric benchmarking study compared `fixed_axis=true` (`[0,I]`) against `[I,I]` initialization across multiple configurations:

**Test cases:** DIIID-like (n=1, n=2, ψ_low = 1e-4 and 1e-1), Solovev (n=1).

**Finding:** For n=2 and Solovev, `[I,I]` and `[0,I]` gave nearly identical results. For DIIID n=1, `[I,I]` produced a large spurious negative eigenvalue — a false instability.

**Isolation:** Selectively zeroing individual m-diagonal entries of U₁ revealed:

| Zeroed m values | et[1] result |
|---|---|
| none (pure `[I,I]`) | large negative (spuriously unstable) |
| m=0 only | matches `[0,I]` exactly |
| m=−1 and m=+1 only | still spuriously unstable |

Zeroing m=0 alone was sufficient; m=±1 had no effect.

**Intermediate workaround** (commit f6b7595): U₁=I with U₁(m=0,n=1)=0 — suppressed the spurious instability but incorrectly excluded the physical constant-displacement regular solution for m=0. This workaround is superseded by the Frobenius initialization in this PR.

---

## Physical Explanation: Frobenius Analysis at the Axis

From Glasser (2016) §VI, the two linearly independent solutions near ψ→0 behave as:

| m | Regular solution | Irregular solution |
|---|---|---|
| m ≠ 0 | ψ^(\|m\|/2) → 0 | ψ^(−\|m\|/2) → ∞ |
| m = 0 | ψ^0 = **constant** | ln(ψ) → −∞ |

For **m ≠ 0**: initializing U₁(m)=0 at small ψ_low is correct in the limit ψ_low→0 (the regular solution vanishes at the axis). This is what `fixed_axis=true` does.

For **m = 0** (degenerate, both Frobenius exponents = 0): the regular solution is ξ^ψ = const (a rigid displacement of the axis). The `[0,I]` initialization sets this to zero — the "fixed axis" boundary condition. The Frobenius initialization allows this constant displacement, the correct free-axis boundary condition.

---

## Implementation: Frobenius Eigenvalue Init

**`compute_axis_init`** (new function in `EulerLagrange.jl`):

For each mode j, forms the 2×2 diagonal block of the Frobenius matrix A₀ = ψ_low · M(ψ_low) using the stability matrix splines (F̄, K̄, G evaluated at ψ_low via `ffit`). Solves the 2×2 eigenvalue problem:

```
lim_{ψ→0} [ψ M(ψ) − a I] v = 0    [Glasser 2016 Eq. 51]
```

Selects the eigenvector with Re(a) ≥ 0 (regular Frobenius solution). For the degenerate m=0 case (both a≈0), the regular eigenvector is identified by dominant |U₁| component (constant displacement, not logarithmic). Returns U₁_init and U₂_init normalized so U₂_init = I.

**`initialize_el_at_axis!`** branches on `ctrl.fixed_axis`:
- `fixed_axis=true`: original U₁=0, U₂=I (constrains axis displacement to zero)
- `fixed_axis=false` (default): Frobenius init via `compute_axis_init`

---

## Results

DIIID-like n=1, ψ_low=1e-4:

| Initialization | et[1] |
|---|---|
| fixed_axis=true (original U₁=0) | 1.19960 + 6.18e-5i |
| fixed_axis=false (Frobenius) | 1.19961 + 1.18e-4i |

Real part agrees to ~1.4×10⁻⁵ relative. Both stable. Comparison figure: `benchmarks/compare_fixed_axis_DIIID-like_ideal_example_psilow1e-4_n1.png`.

---

## Test Plan

- [x] Run `examples/DIIID-like_ideal_example` (n=1): et[1] = 1.1996132 + 1.18e-4i ✓
- [x] Run `examples/Solovev_ideal_example`: et[1] = -0.46277 + 1.93e-3i ✓ (differs from fixed_axis result as expected — m=0 Frobenius solution now included)
- [x] Full test suite: 557 tests pass (Utilities 11, Vacuum 252, Equilibrium 213, ODE 79, FFS full runs 2)
- [x] Comparison figure generated via `benchmarks/run_fixed_axis_comparison.jl`


logan-nc:
Here are some examples:
zeroing m=0 DIII-D example:
image

Not zeroing m=0:
image

DIII-D n=2 no zeroing:
image

logan-nc:
Interestingly, Solovev, with q0 well above 1 did not need any zeroing
image



But I just made n=1, m=0 zero out for all cases to be safe.... should we do something more intelegent?

logan-nc:
Latest results compared to old way:
image

logan-nc:
Haven't check'ed Claude's significant run at this today at all though... suspiciously similar result to using 0s 🤔

matt-pharr:
@logan-nc my adventures in cylindrical land actually taught me about this because I implemented frobenius axis initialization for the scalar newcomb equation there. My conclusions were that it is not that important for overall structure of the solution, as so long as you implement xi = const at some epsilon away from the axis, everything should be fine. 

However, here you have poloidal coupling which muddies the understanding. 

Regardless, I would love it if we could implement analytic curves for the frobenius asymptotics in between the diff-eq solution and the singularity -- would definitely clarify some things. Something like this: 

image

matt-pharr:
Here is the behavior of an $m=1$ solution at the axis:

image

logan-nc:
@matt-pharr would thinking about this problem be a helpful warm up exercise for @viaweber98? If so, it would be great to get two birds with one stone and have the review of this PR be an action item for next week.

If not, who is going to be my reviewer @jhalpern30 or @matt-pharr? It's not a long one - and I'd like to get it merged.

logan-nc:
@claude review this PR. Keep it short and to the point

logan-nc:
## Review feedback addressed

Rebased onto current `develop` (was branched at `4396444d`, before PR #178 introduced the Riccati / parallel paths) and applied the three review items.

### Critical bug — all `initialize_el_at_axis!` call sites updated

5-arg signature now used at every call site, not just the serial standard path:

- `src/ForceFreeStates/Riccati.jl:1229` — `riccati_eulerlagrange_integration` (`ffit` in scope)
- `src/ForceFreeStates/Riccati.jl:1598` — `_initialize_parallel_odet` (signature extended; caller at `parallel_eulerlagrange_integration:1559` passes `ffit`)
- `test/runtests_riccati.jl:165` — `ffit` in scope from `make_matrix(equil, intr_tmp, metric)`
- `test/runtests_parallel_integration.jl:131,194,361` — three additional sites that the original review didn't catch; `metric`+`ffit` constructed locally before each call. The `ode_itime_cost` test had been hand-forcing `mpert=8`; replaced with the standard `mlow/mhigh` derivation so `make_matrix` sees consistent sizes.
- `src/ForceFreeStates/Riccati.jl:79` — stale docstring bullet rewritten. Riccati `S(ψ₀)` is now `0` when `ctrl.fixed_axis=true` and `U₁_Frobenius · I⁻¹` (nonzero) when `false`; S-evolution is well-defined either way. Per-path inheritance of `ctrl.fixed_axis` was chosen so Frobenius init is actually exercised in the production parallel path used by the regression cases.

### Benchmark driver removed

Deleted `benchmarks/run_fixed_axis_comparison.jl` (carried the "Remove before merging" note and `include`d a non-existent `compare_fixed_axis.jl`).

### Full test suite — 911 tests pass

```
Utilities                  11/11
Vacuum                    252/252
Equilibrium               261/261
ODE                        92/92
Riccati Integration        14/14
Parallel FM Integration   108/108
Sing                       75/75
TJ-analytic                16/16
Full ForceFreeStates runs  12/12
CoilGeometry / BiotSavart  19/19
CoilFourier                24/24
D3D I-coil integration      8/8
IMAS integration           21/21
```

### Regression harness

`regress --cases solovev_n1,diiid_n1 --refs develop,local`

**diiid_n1** — `Re(et[1])`: `1.402864 → 1.402872` (+8.2e-6, ~6e-6 relative). Δ' BVP diagonal matches to 9e-13. Perturbed-equilibrium diagnostics shift coherently: `||Φ_res||` -2.27%, island widths +3.24%, Chirikov +5.82%, PE plasma energy -0.18%, PE surface energy -0.26%.

**solovev_n1** — `Re(et[1])`: `9.985e-02 → 9.952e-02` (-3.4e-4, ~0.34% relative). The m=0 Frobenius constant-displacement mode now contributes by default (matches PR description's "differs from fixed_axis result as expected"). ODE step counts drop slightly (633→621 total).

q-profile, pressure-profile, q0, q95, β, singular-surface locations are bitwise identical on both cases.

### Out of scope

@matt-pharr's request for analytic Frobenius asymptote overlays in the plotting tools is tracked separately — that's an `analysis` module extension, not part of this physics PR.

jhalpern30:
> @matt-pharr would thinking about this problem be a helpful warm up exercise for @viaweber98? If so, it would be great to get two birds with one stone and have the review of this PR be an action item for next week.
> 
> If not, who is going to be my reviewer @jhalpern30 or @matt-pharr? It's not a long one - and I'd like to get it merged.

My thoughts on this: the code itself looks clean and I can sign off on that, but I am still skeptical of the physics implementation and the plots above don't convince me 100% that it is working - but I may be wrong and my skepticism of Claude's ability to implement new physics is showing. I thought the last time we talked about this one, we discussed repeating similar scans to my paper to verify it. I would expect something along the lines of the asymptotics Matt plotted for the cylindrical case to verify that the axis displacements are correct, not just nonzero and not massively altering the rest of the solution. Or a longer term goal of recreating Fig. 8 of my paper, showing that a rigid shift of the coil set produces the same Delta', which I believe this change should allow.

I see a few paths to getting this PR merged:

- @matt-pharr is more familiar with the actual physics going on here, so if they think this is a sound implementation that's ok with me and we can merge
- merge the PR, but leave a dedicated physics review for later. Change the fixed_axis to default to true such that in the meantime, the old results stay and we don't accidentally introduce any bugs
- Leave the PR open until we thoroughly verify all of the physics

Thoughts?

logan-nc:
fully agree. Don't consider this my personal thoughts on the physics. My part was just saying that I thought we should be able to have a free axis. I started with an identity initialization and claude did this from there... definitely needs a human to think it through a bit

src/ForceFreeStates/EulerLagrange.jl
```diff
@@ -170,7 +170,7 @@ function eulerlagrange_integration(ctrl::ForceFreeStatesControl, equil::Equilibr
     # Initialization
     odet = OdeState(intr.numpert_total, ctrl.numsteps_init, ctrl.numunorms_init, intr.msing)
     if ctrl.sing_start <= 0
-        initialize_el_at_axis!(odet, ctrl, equil.profiles, intr)
+        initialize_el_at_axis!(odet, ctrl, ffit, equil.profiles, intr)
```
viaweber98, line 173: I think this is a good change, since the code passes in ffit, which is needed for Frobenius to properly run with matrices.

src/ForceFreeStates/EulerLagrange.jl
```diff
@@ -251,7 +251,94 @@ function eulerlagrange_integration(ctrl::ForceFreeStatesControl, equil::Equilibr
 end
 
 """
-    initialize_el_at_axis!(odet::OdeState, ctrl::ForceFreeStatesControl, profiles::Equilibrium.ProfileSplines, intr::ForceFreeStatesInternal)
+    compute_axis_init(ffit, profiles, intr, psi_low) -> (U1_init, U2_init)
```
viaweber98, line 254: This is a major change, and a quite a bit of it I dind't fully understand (based on my understanding as a beginner). I got stuck in a few places. I understand that this change aims to find the eigenvectors/eigenvalues for each little matrix, and it picks the "regular" solution and normalizes things so the result lines up with the expected value (I heavily referenced the "The direct criterion..." paper for understanding). I have two questions: a) for the m=0 case, the comments explain that when the two eigenvalues are nearly equal, it picks the eigenvector w/ the bigger first component instead (from what I understand). I think I follow why, but a comment saying something the eigenvalue ranking could help my readability. b) at line 329 "if abs(v2)....else" the comment says it handles the "purely displacement solution" case. As a beginner I didn't necessarily follow why v2 is approx. 0 means that. It may have been explained in the papers, but currently I can't locate the reference to explain this and might benefit from a comment explaining why. These are all comment/readability suggestions, as I'm not yet qualified to judge whether the math is right (just whether I could follow it).

src/ForceFreeStates/ForceFreeStatesStructs.jl
```diff
@@ -288,6 +288,7 @@ A mutable struct containing control parameters for stability analysis, set by th
     use_parallel::Bool = true    # Default on: unlocks singular/delta_prime_matrix (STRIDE BVP Δ' matrix) used by SLAYER/GGJ downstream.
     populate_dense_xi::Bool = false  # When use_parallel=true, set to true ONLY if a PerturbedEquilibrium pipeline will consume dense ξ. Default false avoids the ~1× parallel-BVP serial-EL re-run for non-PE runs (Δ'/vacuum/ideal-stability only). See ForceFreeStatesControl docstring for the full trade-off (et[1] convention differs by ~0.12% on DIIID between populate=true vs false).
     extended_precision_bvp::Bool = true   # Promote Δ' BVP to Complex{Double64}; default on (Float64 drifts the imaginary Δ' by 2–5× on DIIID-class cases).
+    fixed_axis::Bool = false
```
viaweber98, line 291: I believe this is in regards to the magnetic axis in the Frobenius method. It picks the well-behaved  starting soln. when "Bool = false". My only note is that it was not updated in the initial "Fields" section. As a beginner, that's where I would go for an initial explanation. (however I realize it may have been left blank for the purposes of this exercise)

src/ForceFreeStates/Riccati.jl
```diff
@@ -1223,8 +1226,7 @@ function riccati_eulerlagrange_integration(
     # Initialization — same as eulerlagrange_integration
     odet = OdeState(intr.numpert_total, ctrl.numsteps_init, ctrl.numunorms_init, intr.msing)
     if ctrl.sing_start <= 0
-        initialize_el_at_axis!(odet, ctrl, equil.profiles, intr)
-        # axis init sets u[:,:,1]=0, u[:,:,2]=I → S=0 at axis ✓
+        initialize_el_at_axis!(odet, ctrl, ffit, equil.profiles, intr)
```
viaweber98, line 1229: I understand that ffit was added for correct intial eigenvectors.

src/GeneralizedPerturbedEquilibrium.jl
```diff
@@ -76,7 +76,6 @@ function main(args::Vector{String}=String[]; dd::Union{IMASdd.dd,Nothing}=nothin
     # Read input data and set up data structures
     intr = ForceFreeStatesInternal(; dir_path=path)
     inputs = TOML.parsefile(joinpath(intr.dir_path, "gpec.toml"))
```
viaweber98, line 78: The only noticeable change I see was the removal of whitespace. For a check of my understanding, I believe this line needs settings input by the user, then converts them to the right name and feeds them in as named arguments to create the ctrl object.

test/runtests_parallel_integration.jl
```diff
@@ -124,8 +124,11 @@ using TOML
         intr.mband = intr.mpert - 1
         intr.numpert_total = intr.mpert * intr.npert
 
+        metric = GeneralizedPerturbedEquilibrium.ForceFreeStates.make_metric(equil; mband=intr.mband, fft_flag=ctrl.fft_flag)
+        ffit = GeneralizedPerturbedEquilibrium.ForceFreeStates.make_matrix(equil, intr, metric)
+
         odet = GeneralizedPerturbedEquilibrium.ForceFreeStates.OdeState(intr.numpert_total, ctrl.numsteps_init, ctrl.numunorms_init, intr.msing)
-        GeneralizedPerturbedEquilibrium.ForceFreeStates.initialize_el_at_axis!(odet, ctrl, equil.profiles, intr)
+        GeneralizedPerturbedEquilibrium.ForceFreeStates.initialize_el_at_axis!(odet, ctrl, ffit, equil.profiles, intr)
```
viaweber98, line 131: I believe this was changed so the "ffit" was included, so the test preps the matrices before intializing them.

test/runtests_parallel_integration.jl
```diff
@@ -184,8 +187,11 @@ using TOML
...
```
viaweber98, line 194: Same thoughts as from lines 126-134

test/runtests_parallel_integration.jl
```diff
@@ -341,11 +347,18 @@ using TOML
...
+        intr.numpert_total = intr.mpert * intr.npert
+
+        metric = GeneralizedPerturbedEquilibrium.ForceFreeStates.make_metric(equil; mband=intr.mband, fft_flag=ctrl.fft_flag)
+        ffit = GeneralizedPerturbedEquilibrium.ForceFreeStates.make_matrix(equil, intr, metric)
 
         # Use the first chunk from chunk_el_integration_bounds: guaranteed rational-free interior
-        odet_tmp = GeneralizedPerturbedEquilibrium.ForceFreeStates.OdeState(8, 10, 5, intr.msing)
-        GeneralizedPerturbedEquilibrium.ForceFreeStates.initialize_el_at_axis!(odet_tmp, ctrl, equil.profiles, intr)
+        odet_tmp = GeneralizedPerturbedEquilibrium.ForceFreeStates.OdeState(intr.numpert_total, 10, 5, intr.msing)
+        GeneralizedPerturbedEquilibrium.ForceFreeStates.initialize_el_at_axis!(odet_tmp, ctrl, ffit, equil.profiles, intr)
```
viaweber98, line 361: I believe this change is so it derives the real mode count from the equilibrium. It also adds ffit into "initialize_el_at_axis!" for the same reasoning stated in previous comments.

test/runtests_riccati.jl
```diff
@@ -162,7 +162,7 @@ end
 
         # Use an initialized OdeState just for spline_hint and chunk bounds
         odet_tmp = FFS.OdeState(N, ctrl.numsteps_init, ctrl.numunorms_init, intr_ric.msing)
-        FFS.initialize_el_at_axis!(odet_tmp, ctrl, equil.profiles, intr_ric)
+        FFS.initialize_el_at_axis!(odet_tmp, ctrl, ffit, equil.profiles, intr_ric)
```
viaweber98, line 165: The change added in ffit to match the new Frobenius magnetic axis intialization. It needs it to compute the correct starting values.

viaweber98 (commented):
I left comments for my review of this pull request. Some comments were for clarification of material (for my own understnading), and some were questions I had regarding the changes made.

viaweber98 (approved):

matt-pharr:
@logan-nc have you tested whether this affects any of the core GPEC results e.g. dominant modes/etc?

logan-nc:
🙌

src/ForceFreeStates/ForceFreeStatesStructs.jl
logan-nc: @matt-pharr it looks like this review comment got missed? The new struct field `fixed_axis` is still missing from the struct doc string.

PR #236, opened 2026-05-06
Pr 196

aoifeturner:
## Summary

Only `PerturbedEquilibrium/FieldReconstruction.jl` was changed. This file rebuilds the perturbed fields (ξ, B) from the ODE solution and response matrices, covering eigenmode stacking, perturbed b pieces, Clebsch/contra/cova, b_n/xi_n, area-weighted b^ψ, and R/Z/φ reconstruction.

The parallelism comes from `@threads` over the per-flux-surface (`ipsi`) loops, since each surface is independent. Anywhere the serial code shared a matrix, vector, or spline buffer between iterations, each thread now gets its own copy to keep things race-free.

Equilibrium, ForceFreeStates, and vacuum are untouched. To see a speedup, run with `julia -t N` where N > 1, on a case that goes through PE and `compute_response` like the DIIID-like example.

## Benchmarks

Tested on the DIIID-like ideal example across 1–10 threads (5 trials each, averaged):

| Threads | Avg runtime (s) |
|---|---|
| 1 (og) | 12.41 |
| 2 | 11.38 |
| 3 | 10.99 |
| 4 | 10.91 |
| 5 | 10.94 |
| 6 | 10.23 |
| 7 | 10.38 |
| 8 | 11.22 |
| 9 | 10.67 |
| 10 | 10.70 |

Best speedup at 6 threads: ~17% faster than single-threaded. Note that the threaded section (field reconstruction inside Perturbed Equilibrium) is a small fraction of total runtime, so this represents close to the theoretical maximum gain from this change alone.

## Testing

- Physics output is unchanged vs. single-threaded runs
- Tested with `julia -t 2` through `julia -t 10`, then also did og run on julia -t 1
- All existing tests pass

image


logan-nc:
Superseded by #263.

This PR was branched from a then-current PerturbedEquilibrium branch that has since been squash/rebase-merged into `develop` under different commit SHAs. As a result GitHub computed the diff against a stale base and showed ~161 files / +103k — but **all of that is already in `develop`**. The only real, unique contribution here was a single commit threading the per-surface loops in `FieldReconstruction.jl` (+89/−44 on one file).

#263 re-applies that work cleanly on top of current `develop` as a single-file change, with the threading hardened to the project convention (`@threads :static` + `maxthreadid()`-sized per-thread workspaces). It's verified to produce output identical to `develop` and bit-identical across thread counts. Original authorship is preserved via `Co-authored-by`.

Closing this one in favor of the clean re-application; thanks @aoifeturner for the contribution.

PR #252, opened 2026-06-04
ForceFreeStates rewrite — historical record (superseded by #178)

logan-nc:
> **Status:** Historical record only. This PR will be **closed without merging**. The team has chosen to land the incremental, opt-in dispatcher approach in **#178** (`perf/riccati`) instead. The follow-on issues below capture the pieces of this branch that remain worth porting on top of #178.

## Summary

`feature/forcefreestates-rewrite` is a clean reimplementation of the ForceFreeStates ideal-MHD stability module that grew out of `perf/riccati` (#178). On top of #178's chunked-Riccati + parallel-FM foundation it adds: concurrent dual-route integration, an eigenvalue cross-check diagnostic between the two paths, BLAS pinning to avoid nested-parallelism oversubscription, a role-based file split, and a unified `integration_method` / `integration_threads` API in place of the per-feature flags.

## Provenance — branches this work built on

- `feature/unified-ustore-reconstruction` — baseline being consolidated (the three messy integration paths this rewrite collapsed)
- `perf/riccati` — head of #178: chunked-Riccati, parallel FM, Δ' output
- `perf/riccati-full-geqdsk-scans` — full-geqdsk scan prototyping
- `test/parallelize_equilibrium_solver` — adjacent parallelism experiments

## Key differences from #178

- **Concurrent dual-route execution.** Route A (legacy EL, ~1 thread, produces the trusted `u_store` consumed by PerturbedEquilibrium) and Route B (chunked-Riccati on N−1 threads, produces Δ' and a diagnostic eigenmode) run **in parallel** via per-chunk `Threads.@spawn`. Wall time pegs to `max(EL_serial, FM_parallel) ≈ EL_serial` for N ≥ 2 threads. #178 does not run the two paths concurrently.
- **Eigenvalue cross-check diagnostic.** `et[1]` from both routes is compared every run, with **>1% warn / >10% error** thresholds; Route B's `odet_riccati` is written to HDF5 as a diagnostic. #178 has no such cross-check.
- **BLAS pinning during the concurrent section** to avoid the OpenBLAS-per-call-pool × Julia-tasks oversubscription that was destroying scaling on n=4.
- **Unified `integration_method` / `integration_threads` API.** `use_parallel`, `use_riccati`, and `parallel_threads` were removed and collapsed into a single dispatcher field on `PerturbedEquilibriumControl`. #178 keeps the per-feature flags.
- **Role-based file split.** `DeltaPrime.jl` (Δ' assembly, 817 lines) and `Chunking.jl` (cost-based load-balanced partitioning, 202 lines) were extracted from the larger `Riccati.jl` / `Integration.jl`. #178 keeps the combined files.
- **Out-of-scope additions on this branch** (mentioned for completeness, not part of the comparison): TJ-like analytic equilibrium and LAR scan support were ported into `src/Equilibrium/`. Independent of the stability rewrite; the team can pull these in separately if desired.

## Performance results

[benchmarking figure: to be attached by author]

- **DIII-D-like, warmed wall time:** 1 thread **42.7 s** → 4 threads **22.2 s** (~**1.9× speedup**, pegged to legacy-EL time as designed).
- **BLAS pinning recovered the expected 4-thread scaling.** Before: chunked n=4 ≈ 116 s with ±23 s variance (BLAS thrashed by nested parallelism). After: **53 s with ±6 s variance**.
- **Thread-scaling artifacts** at `benchmarks/ffs_thread_scaling_results/threads_{1,2,3,4}.toml` (BenchmarkTools output, 3 samples each, n=1 and n=4 cases).
- **Regression status.** Route A is byte-identical to develop on `diiid_n1` for resonant flux, Δ', island half-widths, and `et[1]`. Solovev cases drift ≤ 4.5%, correctly caught by the Route-A / Route-B cross-check.

## Lessons learned (not turning into issues)

An earlier experiment on this branch tried to construct `u_store` directly from the chunked-Riccati fundamental-matrix histories, eliminating the need to also run the legacy EL sweep. It was **reverted** when perturbed-equilibrium regressions revealed 10⁴–10⁶× corruption in resonant flux, Δ', and island half-widths. Root cause: per-chunk Riccati renormalization makes the basis **piecewise**, which breaks PerturbedEquilibrium's assumption of a uniform coefficient vector across the integration domain. Kept here so the next person doesn't reattempt the same shortcut without a basis-stitching step.

## Why this is closing without merging

The team chose the incremental, opt-in `perf/riccati` (#178) approach for lower disruption to the rest of the project. The valuable pieces of this branch are being captured as separate follow-on work below.

## Follow-on issues (after #178 lands)

- #253 — Concurrent dual-route integration with BLAS pinning
- #254 — Eigenvalue cross-check diagnostic between integration paths
- #255 — File split (DeltaPrime/Chunking) and unified `integration_method` API


logan-nc:
@d-burg @matt-pharr FYI
To eat up my final week of Claude Max I had asked it to re-write the Riccati implementation from scratch trying to make it aas clean a code architecture as possible. Since you've been doing a great job on #178, I've decided the best approach is for you two to merge that (within a week of now please!) and then to implement the key lessons learned from this exercise afterwards as separate, smaller, more manageable PRs. I've had claude make the issues for us, so it should be easy to tack each on one at a time once you've merged.

This branch was able to get the following performance:
image

It's actually better than I had remembered @d-burg! It means Riccatti (and thus Δ') costs us essentially no wall time as long as we have 2 threads for n=1 and ~5 threads in n=4. Sweet! I'd be curious to see a similar plot for your current state of #178 so we know how big a priority these add-ons are after that merges.

logan-nc:
@d-burg I assigned you, with the task being to close this PR. Closing it will be interpreted as acknowledgement that you've skimmed it and are aware of the issues listed above

d-burg:
Lessons noted.

PR #263, opened 2026-06-04
PerturbedEquilibrium - PERFORMANCE - Thread per-surface field reconstruction

logan-nc:
## Summary

Parallelizes the per-flux-surface (`ipsi`) loops in `PerturbedEquilibrium/FieldReconstruction.jl` with `Threads.@threads`. Each surface is independent, so eigenmode stacking, the perturbed-b pieces, Clebsch/contra/cova components, `b_n`/`xi_n`, area-weighted `b^ψ`, and R/Z/φ reconstruction all parallelize cleanly across ψ.

**This supersedes #236**, which was branched from a stale base (a now-merged PerturbedEquilibrium branch) and therefore showed a 161-file / +103k diff that was entirely already in `develop`. The real contribution was a single commit touching one file; this PR re-applies that work on top of current `develop` as a clean, single-file change, and hardens the threading to the project's established convention.

## Threading hardening vs. #236

The original change used `nthreads()`-sized workspaces and the default `:dynamic` scheduler while indexing buffers by `threadid()`. This PR brings it to the GPEC convention used in `Riccati.jl`, `Vacuum/Kernel3D.jl`, and `Equilibrium/DirectEquilibriumByInversion.jl`:

- **`Threads.@threads :static`** so `threadid()` is a stable per-iteration index
- **per-thread workspaces sized by `Threads.maxthreadid()`** (not `nthreads()`), since `threadid()` can exceed `nthreads()` under the interactive threadpool — the original could index out of bounds
- the immutable `FourierTransform` functor is shared read-only (verified no internal mutable state)

## Verification

- **Regression harness** (`regress --cases diiid_n1 --refs develop,local`): all **30 quantities identical** to develop (0.0e+00 diff) — `et`/`ep`/`ev`, resonant flux, island half-widths, Chirikov, Δ′, PE energies, torque.
- **Threaded determinism**: full `gpec.h5` output (**239 numeric datasets**) is **bit-identical across 1, 2, and 6 threads** — proves no data races.
- Builds and loads cleanly; field-reconstruction section scales ~17% at 6 threads on the DIII-D-like example (original benchmark).

### DIFF
--- a/src/PerturbedEquilibrium/FieldReconstruction.jl
+++ b/src/PerturbedEquilibrium/FieldReconstruction.jl
@@ -126,7 +126,8 @@ function reconstruct_physical_fields(
     mthsurf = length(equil.rzphi_ys) - 1
     thetas_for_avg = [(k - 1) / mthsurf for k in 1:mthsurf]
     Jb_psi_modes = similar(b_psi_modes)
-    for ipsi in 1:npsi
+    # Surfaces are independent; :static keeps threadid() stable for any per-thread state.
+    Threads.@threads :static for ipsi in 1:npsi
         psi = psi_grid[ipsi]
         hint2d = (Ref(1), Ref(1))
         area = 0.0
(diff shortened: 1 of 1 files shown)


logan-nc:
image

Threading parity verified directly on the reconstruction outputs (not just the scalar regression): xi_psi, clebsch_psi/psi1/alpha, b_n, xi_n and all other mode fields are bit-identical (max abs diff = 0.0) between -t 1 and -t 4. See overlay of the m=2 and m=5 components vs ψ above

PR #278, opened 2026-06-11
ForcingTerms: HDF5 + analytic coil interfaces, with gpec.h5 coil snapshot (issue #229)

### LINKED ISSUE #229: Updated coil file format needed
We should update the ascii file geometry descriptions to a modern format.
Probably h5.

Also add analytic window pane coil options.
And PF hoops.

logan-nc:
Closes #229 (builds on `feature/rerun-from-h5`; retarget to `develop` after that merges).

## Summary

Modernizes coil/external-field input: keeps the legacy ASCII `.dat` format, adds a modern HDF5 format and analytic coil generators, and makes a coil run replayable from `gpec.h5` alone. Every source reduces to a `CoilSet`, so the existing currents + `apply_transforms` (shift/tilt) path is shared and unchanged.

### Coil input sources (TOML `[[ForcingTerms.coil_set]]`, selected by `source`)
- **`.dat`** — unchanged (legacy preserved).
- **`pf_hoop`** — horizontal circle (`radius`, `height`); tilt breaks axisymmetry for n≠0.
- **`window_pane`** — axis-aligned picture-frame array; cross-section as two opposite `[R,Z]` corners (`rz_corners`), `ncoil_gen`, `gap_fraction`, `phi0`.
- **`helical`** — helical array on a circular cross-section torus (`R0`, `a`, `m_hel`, `n_hel`, `n_coils`, ...).
- **`file` + `.h5`** — modern HDF5 (extension-dispatched; `name` selects one set from a multi-set file).

### HDF5 coil format
Group-based, **one subgroup per coil set**, content-tolerant (unrelated datasets ignored). Functions: `read_coil_h5` / `write_coil_h5` / `save_coils_to_h5` / `load_coils_from_h5_group!`, plus `convert_coil_dat_to_h5` / `convert_coil_h5_to_dat`.

### Rerun-from-h5 integration
A coil run stores the geometry actually used (post shift/tilt, with currents) under `input/raw_inputs/coils/`. `main_from_h5` gains `--coil-source {forcing-modes|coils}`:
- `forcing-modes` (default) — recompute from the stored TOML coil config (needs original geometry files).
- `coils` — recompute from the geometry stored in `gpec.h5`; **no original `.dat`/`.h5` required**, and the equilibrium may be overridden on replay.

## Verification
- **Unit tests** (`runtests_coils.jl`): generators, PF-hoop analytic B_z (Jackson 5.38), h5 round-trip / multi-set / junk-tolerance, `dat↔h5`, dispatch — all pass.
- **Rerun test** (`runtests_rerun_from_h5.jl`): decisive case **deletes the `.dat` between source run and replay** — default replay fails, `--coil-source coils` reproduces the resonant flux. Also fixed a pre-existing stale-key bug in the existing #171 test (`vacuum/et` → `FreeBoundaryStability/eigenmode_energies`); suite now 17/17 + 8/8.
- **Regression harness** `diiid_n1` (the `.dat` coil case), base vs HEAD: **30/30 quantities unchanged, every diff `0.0e+00`** — refactor is behavior-preserving.
- **Docs**: build clean (no `missing_docs`/broken links); `forcing_terms.md` and `coil_geometries/README.md` document the sources, HDF5 schema, and rerun snapshot.

## Repo-maintenance note
The 118 bundled `.dat` files (3.9 MB) stay git-tracked and canonical (diff-able, review-able). HDF5 is for user-supplied coils, the rerun snapshot, and optional export — converters provided. No monolithic binary blob.


logan-nc:
## New analytic & file-based coil building paths

This PR rounds out the `ForcingTerms` coil interface so external perturbation coils can be
specified three ways — **from a geometry file**, **from analytic generators**, or **placed by
physical standoff off the plasma surface** — all feeding the same downstream forcing pipeline
(`compute_coil_forcing_modes!` → `b_n_x` on the control surface).

### What's new

| Source (`source = ...`) | Builder | How you specify it |
|---|---|---|
| `pf_hoop` | `make_pf_hoop` | `radius`, `height` (+ optional `tiltx`/`n_tilt` for an n≠0 shape) |
| `window_pane` (corner mode) | `make_window_pane` | two opposite `rz_corners`, `gap_fraction` |
| `window_pane` (**standoff mode**, new) | `make_window_pane_standoff` | `standoff`, `poloidal_angle`, `poloidal_length`, `poloidal_tilt` |
| `helical` | `make_helical` | `R0`, `a`, `m_hel`/`n_hel`, poloidal extent |
| `file` / `.h5` / `.dat` | `read_coil_h5` / `read_coil_dat` | path to a coil geometry file |

The headline addition is **standoff placement** for window-pane arrays: instead of hand-computing
`[R, Z]` corners, you give a physical standoff distance and a poloidal angle, and the frame is
built from the equilibrium's outward surface normal/tangent at that point. `load_coil_sets` now
takes the `equil` so this works directly from the main forcing pipeline.

```toml
# Standoff mode — no manual corners
[[ForcingTerms.coil_set]]
source = "window_pane"
ncoil_gen = 2
standoff = 0.25         # m outward along the surface normal
poloidal_angle = 0.0    # deg (0 = outboard midplane)
poloidal_length = 1.2   # m span along the surface tangent
poloidal_tilt = 0.0     # deg (0 = tangential)
gap_fraction = 0.15
currents = [1e4, -1e4]
```

### Visual verification

3D plasma surface coloured by the applied `b_n_x`, with the generating coils overlaid. Blue coil
→ blue surface, red coil → red surface in every case.

**PF hoop (tilted → n=1)**

image

**Window-pane — corner mode**

image

**Window-pane — standoff mode (now matches corner winding)**

image

**Helical array (m/n pitch)**

image

### Testing

- `test/runtests_coils.jl` — analytic builders, standoff geometry + winding, HDF5 round-trip,
  source dispatch (all green).
- Coil builders aren't exercised by the regression-harness cases, so no tracked quantity moves.
- Figures regenerated with `julia --project=. benchmarks/coil_3d_demo.jl` (demo is a throwaway,
  not committed).

logan-nc:
@ebursch @matt-pharr this is cool because
1. It replaces the need to use OMFIT for big scans of potential coil arrays like Nils and I published for SPARC/ARC
2. It allows a user to re-use coils from someone else's gpec run given the single h5 output. We no longer have keep track of the g-file and coil dat files used for old runs! For example, @matt-pharr could just hand off one h5 of an ITER run and boom, @ebursch would have all the PF coil windings to do with what he likes.

Please prioritize a review here in soon so we get it into the parent `feature/rerun-from-h5` before #217 merges. Even a skim is helpful.

ebursch (commented):
Only had time for a skim, but this looks good to me.

src/ForcingTerms/CoilGeometry.jl
```diff
@@ -196,6 +287,449 @@ function read_coil_dat(filepath::String)
...
+Build a single horizontal circular loop (PF hoop) of `radius` [m] centered on the
+machine axis at vertical position `height` [m], discretized into `nsec` points.
+
+The loop is closed (point `nsec` coincides with point 1). A bare hoop is axisymmetric
+and produces only an n=0 field; tilting or shifting it (via [`apply_transforms`](@ref))
+breaks axisymmetry to produce n≠0 forcing. Currents are zeroed; `load_coil_sets` fills them.
+"""
+function make_pf_hoop(; radius::Real, height::Real, nsec::Int=361, name::String="pf_hoop")
+    radius > 0 || error("make_pf_hoop: radius must be > 0 (got $radius)")
+    nsec >= 4 || error("make_pf_hoop: nsec must be >= 4 (got $nsec)")
```
NilsLeuthold: So to my understanding nsec is just the resolution of the coil. Should probably be significantly higher than 4 points for such a poloidal loop, as 4 points would basically give you a square coil acound your tokamak. Maybe nsec >= 128 would be more reasonable, or at least >=64.

src/ForcingTerms/CoilGeometry.jl
```diff
@@ -196,6 +287,449 @@ function read_coil_dat(filepath::String)
...
+    Rp, Zp = point(theta_sfl + h)
+    Rm, Zm = point(theta_sfl - h)
+    dR = (Rp - Rm) / (2h)
+    dZ = (Zp - Zm) / (2h)
+
+    nR, nZ = dZ, -dR
+    nrm = hypot(nR, nZ)
+    nR /= nrm
+    nZ /= nrm
+    if nR * (R - equil.ro) + nZ * (Z - equil.zo) < 0  # force outward orientation
```
NilsLeuthold, line 356: Looks like you added a check for the normal vector pointing outwards, that's good!

NilsLeuthold (changes requested):
Looks great. Just one minor suggestion for change, see comment below on line 302. On another note, when I start to use julia GPEC I'll have a look over the script again and make variable more human-readable. Also then I'll put in a check to see if the coil touches the separatrix (can happen right now e.g. when coil is poloidally-tilted 90deg and the standoff distance is shorter than half the coil length).

logan-nc:
Thanks @NilsLeuthold! Addressed your review in b78248c3:

- **`make_pf_hoop` `nsec` floor** raised from `>=4` to `>=361` (the default) — a full poloidal loop now gets at least ~1° resolution instead of a near-square.
- **Separatrix-touch guard** added to `make_window_pane_standoff`: errors when a tilted coil would cross the boundary (`standoff - ½·poloidal_length·|sin(tilt)| ≤ 0`), exactly the 90°-tilt / short-standoff case you flagged.

Added unit tests for both (sub-floor `nsec` throw; standoff-intersection throw + a valid case). The readable-variable-name pass is left to you as you mentioned. Re-requesting review.

PR #279, opened 2026-06-13
ForceFreeStates: alias-free spline-quadrature metric Fourier fit (fspline_fit_1 port), wired to fft_flag

matt-pharr:
## Summary

Wires up the (previously dead) `fft_flag` control and adds an **alias-free spline-quadrature** path for the metric tensor's poloidal Fourier coefficients — a faithful port of Fortran `equil/fspline.f::fspline_fit_1`. With `fft_flag=false` (the existing `ForceFreeStatesControl` default, matching Fortran's `.FALSE.`), each metric Fourier coefficient is the **analytic integral of a periodic cubic spline** against `exp(-imθ)` instead of a bare FFT.

## Why

The metric components (`g^ij · J` on shaped geometry) are **not band-limited**, so a bare DFT over the `mtheta` θ-samples aliases high-θ-harmonic content into the low-`m` coefficients that set Δ′. That makes Δ′ / energies converge only as `mtheta` grows. The Fortran code avoids this by default (`fspline_fit_1`, alias-free, O(Δθ⁴)). Julia's `make_metric` accepted `fft_flag` but never used it — it always FFT'd.

## What changed (3 files)

- **`src/Utilities/FourierCoefficients.jl`** — constructor split into `_ffit_fft` / `_ffit_spline` + `_periodic_nodal_derivs`; faithful `fspline_fit_1` port (periodic-spline analytic integration, with the small-`m·Δθ` Taylor fallback). Same `c_m = (1/2π) ∫ f e^{-imθ} dθ` convention as the FFT path, so it is a drop-in.
- **`src/ForceFreeStates/Fourfit.jl`** — `make_metric` now passes `fft_flag` through; corrected docstring (the old one had the meaning backwards).
- **`test/runtests_utilities.jl`** — new spline-path testset.

## Validation

- **Unit test:** band-limited coefficients recovered exactly; for `f = 1/(a − cosθ)` the spline error is ~1e-4 at N=16 and ~1e-7 at N=64, vs the FFT's ~1e-1 / ~1e-2 (aliased).
- **End-to-end (DIII-D-like):** the total-energy eigenvalue `et[1]` changes **0.2%** over `mtheta` 64→256 with the spline, vs **11%** with the FFT. Crucially, the FFT converges *up onto* the spline value as `mtheta` grows — i.e. **the spline is the `mtheta`-converged answer and matches the Fortran result; it converges onto Fortran, it does not depart from it.**

## Behavioral change (regression-visible)

The effective default flips from FFT to spline (the flag default was already `false` but ignored), so finite-`mtheta` Δ′/energy values shift slightly — they become `mtheta`-robust and aligned with Fortran. The regression harness should be re-baselined as part of review.

## Notes for review

- When this merges onward into the galerkin branch, a **second `FourierCoefficients` call** (the kinetic `fmodb` path, not present on `develop`) will also need `fft_flag` wired.
- `_periodic_nodal_derivs` uses per-column periodic-spline fits — correct, and a candidate for `Series` batching later.

### DIFF
--- a/src/ForceFreeStates/Fourfit.jl
+++ b/src/ForceFreeStates/Fourfit.jl
@@ -44,7 +44,10 @@ The metric coefficients stored in `metric.fs` include:
 ### Arguments
 
   - `mband::Int`: Number of Fourier modes to retain in the metric representation.
-  - `fft_flag::Bool`: If `true`, enables use of Fourier fitting for storing metric coefficients.
+  - `fft_flag::Bool`: Poloidal Fourier method for the metric coefficients. `false` (default in
+    `ForceFreeStatesControl`, matching Fortran) uses an alias-free periodic-spline quadrature, so
+    Δ′ is essentially independent of `mtheta`. `true` uses a bare FFT, which aliases the
+    non-band-limited metric and makes Δ′ converge only as `mtheta` grows.
 
 ### Returns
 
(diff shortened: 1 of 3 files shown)


logan-nc:
this is a classic example of something very smart being done in DCON but the interface being confusing, forcing users to be experts in the source code and inviting mistakes.

what if like to see in this pr: 
- benchmarks comparing the fft and this approach. how much does it cost in time? can it be optimized in juli to have lower cost?
- benchmark showing impact on ideal results (PE resonant fields??)
- No user flag. Just have code logic do the right thing under the hood. if the above shows it is fast enough and/or matters then we just switch wholesale. If it only matters for certain paths, then we wire that up under the hood. No toml bloat!

PR #280, opened 2026-06-14
KineticForces - Validate calculated kinetic-DCON vs Fortran (#227); thread nutype, guard torque diagnostics

### LINKED ISSUE #227: KineticForces — replace `ximag` offset with principal-value + residue for `nutype="zero"` pole handling
## Motivation

The energy integrand in `src/KineticForces/EnergyIntegration.jl` has a real
pole wherever the resonance denominator vanishes:

```
denom = i*(l_eff*wb*sqrt(x) + n*(we + wd*x)) - nu
```

When `nutype = "zero"` (collisionless) there is no `nu` to regularize the
integral, so the current code relies on a **contour offset** `ximag > 0`:
integrate along `(1e-15, x + i*ximag)` instead of the real axis. This is
inherited from Fortran PENTRC and has two defects:

1. **Accuracy**: the offset introduces an `O(ximag²)` error in the integrated
   torque. There is no principled way to pick `ximag` — users currently set
   it by hand, and the default `0.0` silently disables pole handling.
2. **Performance**: when `ximag > 0` the entire `[0, xmax]` range is
   integrated with a ComplexF64 kernel (`_energy_integrand_complex`),
   paying full complex `sqrt`/`exp` cost across the whole domain even
   though the pole is typically localized to a small x-window.

A proper principal-value + residue decomposition would be both more
accurate (exact in the `ε → 0` limit) and faster (the PV pieces run in
the fast Float64 kernel landed in PR #224; only the small ε-semicircle
around each pole pays the complex kernel cost).

## Performance context

PR #224 landed a Float64 real-axis energy kernel that's ~2× faster than
the old complex kernel on the `ximag == 0` path. That makes the `ximag == 0`
fast path cheap, but users who currently need pole handling for
`nutype=\"zero\"` are stuck paying the slow complex kernel everywhere.
PV+residue would let them stay in the Float64 kernel for most of the
range, paying the complex kernel only on the ε-arc.

## Proposed implementation

Replace the `ximag`-based contour offset with:

```
∫₀^xmax f(x) dx  =  PV ∫₀^xmax f(x) dx  +  i·π·Σ Res[f, xₚ]
```

where the principal value is computed as `PV = lim_{ε→0} [∫₀^{xₚ-ε} + ∫_{xₚ+ε}^xmax]`
plus a small-ε semicircular arc contribution of `i·π·Res[f, xₚ]` (upper
half-plane convention consistent with Landau causality).

Pole locations `xₚ` are the real roots of the denominator and can be
found analytically for the quadratic-in-sqrt(x) form. Residues follow
from `f(xₚ) / denom'(xₚ)`.

## TOML changes (follow-up PR)

Replace `ximag = 0.0` with `x_residue_radius = 0.1` in `[PENT_CONTROL]`.
Semantics: residue-arc radius in `x = E/T` units. Default `0.1`.

## Regression risk (important)

A prior attempt at this rewrite broke the Solovev kinetic \"calculated\"
fullruns test: `et[1]` dropped from 34.176 → 20.475. Root cause was
never identified; the source edits were lost before bisection
completed. That test has since been disabled in PR #224 (the
`kinetic_source = \"calculated\"` path is out of scope for the current
perturbative FFS→PE→KF work and its baseline was never physics-validated).

**Before shipping the PV+residue rewrite**, the follow-up PR must:

1. Re-enable the `kinetic_source = \"calculated\"` fullruns test.
2. Establish a *physics-based* baseline for that test — either against
   Fortran PENTRC on the same Solovev input, or against an analytical
   NTV limit. Do **not** re-pin to a number captured from the code.
3. Validate that the PV+residue result matches the `ximag → 0` limit of
   the old code on a smooth-test-case regression harness case with
   `nutype = \"zero\"`.

## Cross-references

- PR #224: landed the fast Float64 real-axis kernel and deferred this work
- Plan file (local): `/Users/nlogan/.claude/plans/validated-exploring-clarke.md` (as of the commit landing #224)
- Relevant files:
  - `src/KineticForces/EnergyIntegration.jl` (`_energy_integrand_real`, `_energy_integrand_complex`, `integrate_energy_ode`)
  - `src/KineticForces/KineticForcesStructs.jl` (control struct with `ximag` field)
  - TOML: every `pentrc.toml` with `[PENT_CONTROL]`

### LINKED ISSUE #281: KineticForces — collisionless (nutype="zero") energy integral NaNs on real resonance poles; add nuzero regression sentinel once fixed
## Summary

Two independently-developed code paths derive flux-surface geometry (`|∇ψ|`, Jacobian, `r²`/`η`-offset and their θ-derivatives) from the same `rzphi` bicubic, via separate implementations. They should be consolidated onto a single rzphi-geometry primitive.

## Background

- **`develop`** (`src/PerturbedEquilibrium/FieldReconstruction.jl`): `_build_rzphi_geometry` resamples the smooth `rzphi` primitives (`r²`, `η`-offset, `J`, and their θ-derivatives) onto a coarse ψ grid and builds cubic interpolants for the PE field-reconstruction transform (added in #264, `3dcde8fd`).
- **`feature/power-norm-eigenvalues`** (`src/Equilibrium/FluxSurfaceMetrics.jl`): `flux_surface_metric` is a pointwise evaluator of the same `rzphi` quantities returning `(r, jac, |∇ψ|)`, shared by FFS (`PowerNorm.jl`) and PE (`SingularCoupling.jl`) after the `7dcfd5dc` refactor.

These overlap conceptually: both pull `r²`, the `η`-offset (`deta`), `J`, and the θ-derivatives off `rzphi` and combine them. They are **not** redundant code today (different consumers, different representations — real-space `|∇ψ|` weight + area integral vs. mode-space reconstruction geometry), so this is a maintainability cleanup, **not** a performance fix. The hot per-ODE-step path is already served by coarse ψ-splines on both sides (`sqrtamat_spline` in FFS, `_build_*_interp` in PE).

## Proposal

Make `flux_surface_metric` (in the Equilibrium module) the single source of truth for extracting smooth geometry primitives from `rzphi`, and have both `_build_rzphi_geometry` and the area/weight helpers sample it. This keeps the GPEC `gpout.f` metric conventions in one place and prevents the two paths from drifting.

## Risk / why this is its own issue

This crosses module boundaries (Equilibrium ↔ PerturbedEquilibrium) and carries real FMA-level numerical risk: the `7dcfd5dc` refactor already documented benign 1-ULP shifts (≤ 2.2e-16) in 5 PE singular-coupling quantities purely from moving the metric block across a function boundary. Any consolidation must be validated against the regression harness (`diiid_n1`, `solovev_*`) with the usual 1e-8 noise thresholds, and the ψ-derivative handling needs care — `_build_rzphi_geometry` deliberately differentiates the resample spline rather than resampling the bicubic ψ-derivatives (those are only C¹ and ring near axis/edge), so a shared primitive must preserve that distinction.

## Out of scope for

`feature/power-norm-eigenvalues` (PR #225) — that branch already centralized the FFS/PE `|∇ψ|` duplication; this is the follow-on consolidation with `develop`'s reconstruction geometry.

logan-nc:
## Summary

Closes the regression-validation gaps in #227 (PV+residue energy-integral pole handling)
and folds in the collisionless energy-integral fix (#282, merged into this branch).

**Investigation finding:** the PV+residue rewrite #227 proposed is already on develop
(exact analytic Sokhotski-Plemelj; `ximag` deprecated; the proposed `x_residue_radius`
knob is unnecessary). The open gap was #227's acceptance criterion 2 — the calculated-path
Solovev `et[1]` was pinned to a code-captured value rather than a physics baseline.

## Criterion 2 — physics validation against Fortran kinetic DCON

`benchmarks/benchmark_solovev_kinetic_stability.jl` generates a Fortran kinetic-DCON deck
matched to the Solovev calculated fixture (same Solovev params, kinetic profile, mode band,
`nl`, collision operator, grid, edge truncation), runs Fortran `dcon`, and compares its
`W_t[1]` to Julia's `XiNorm/eigenmode_energies[1]`:

| | Julia | Fortran | Error |
|---|---|---|---|
| Re(et[1]) | 15.888 | 15.619 | **1.7 %** |
| Im(et[1]) | −0.711 | −0.660 | **7.6 %** |

Passes the DIIID kinetic acceptance thresholds (Re ≤5 %, Im ≤20 %): the PV-era value is
correct; the pre-rewrite `ximag` value was not.

## Criterion 3 — collisionless energy integral (now included via #282)

The energy integral (`src/KineticForces/EnergyIntegration.jl`) is a **single real-x-space
path over `[0, X_ENERGY_MAX]` for all collisionalities** (matching Fortran PENTRC
`energy.f90`). Each resonance pole `x_pole = x_res − i·ν/Ω′` is removed analytically by
principal-value + residue (Sokhotski-Plemelj); the collisionless case (ν ≡ 0) is the exact
ν→0⁺ limit, with its on-axis pole resolved by the analytic Laurent regular-part limit
`[h′ − h·Ω″/(2Ω′)]/(iΩ′)`. This resolves the collisionless NaN that earlier blocked the
`nuzero` sentinel, which is now added as both a `runtests_fullruns.jl` end-to-end smoke test
and a `solovev_kinetic_nuzero` regression case (the calculated deck run collisionless via a
new harness `[overrides]` mechanism).

## Bug fixes

1. **Calculated path ignored `nutype`/`f0type`/`nufac`** — `compute_kinetic_matrices_at_psi!`
   never accepted them, so a user setting them in `[KineticForces]` was silently ignored
   (hardcoded `"harmonic"`/`"maxwellian"`). Now threaded from `kf_ctrl`; defaults match the
   prior values, so existing cases are byte-identical.
2. **Standalone NTV torque diagnostics crashed** when `[KineticForces]` was present without
   PE data (the calculated path). Now skipped with an info message.
3. **`nstep_total` was 0 on the kinetic paths** — the Riccati callback and the parallel-FM
   propagator solves never counted solver steps. Now counted (Riccati callback per accepted
   step; parallel-FM chunks via per-proxy accumulation summed after the BVP barrier).

## Verification

- Kinetic unit tests pass (energy-integral output byte-identical for ν>0).
- Regression harness (local): `solovev_kinetic_calculated` et[1] ≈ 1.894−1.525i,
  `solovev_kinetic_nuzero` et[1] ≈ 2.333−2.174i, both matching the pressure-consistent
  `kinetic.dat`. The calculated case's `example_dir` now points at a new
  `examples/Solovev_kinetic_calculated_example` (absent on develop), so the first baseline
  is captured with `--refs local`.

## Review response (this PR)

- Benchmark: renamed local `run`→`rundir` (shadowed `Base.run`); mirrored `nufac` into the
  Fortran deck.
- Trimmed over-long comments to the project convention (`X_ENERGY_MAX`, the resonant-integral
  docstring, the NTV-skip comment).
- Energy-integral hot path: `sizehint!` the pole buffers and replaced the type-unstable
  `breaks...` splat into `quadgk` with an explicit breakpoint-count branch.


logan-nc:
@claude review this pr

PR #290, opened 2026-06-19
DOCS - Add ballooning docs missed by PR #234

logan-nc:
PR #234 merged into develop before the `document update` commit (`e834a69c`) on the ballooning branch, so the ballooning documentation never landed. This restores just that documentation content.

## What's added
- `docs/src/ballooning.md` — the ballooning / Mercier local-stability page
- `docs/make.jl` — registers the page under API Reference (develop's `size_threshold` settings preserved)
- `docs/src/stability.md` — cross-reference to the new page + minor `D_R` wording
- `docs/src/assets/ballooning/*.png` — the two referenced figures
- `.gitignore` — un-ignore `docs/src/assets/**/*.png` so the figures can be committed

## Intentionally excluded from `e834a69c`
The scratch artifacts from that commit are **not** brought over: `analysis/pr234_mercier_core/*` (incl. a committed `.pyc`), the root `Ballooning_en.tex`, and `core_local_stability_spike_report.md`.

Docs-only; no code or numerical change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #291, opened 2026-06-19
Followup/ballooning docs

jmlmir369:
I have converted the document previously uploaded as a PDF into Markdown and uploaded it here. Additionally, I have included the mathematical derivation requested in the previous pull request, which demonstrates that the Mercier criterion and the method using the determinant ($\det()$) are equivalent.


jmlmir369:
I have just noticed that this PR (https://github.com/OpenFUSIONToolkit/GPEC/pull/290) is open. I will close this PR now.

PR #292, opened 2026-06-19
EQUIL - BUG FIX - Integrate equilibrium 0D flux integrals out to the boundary

logan-nc:
## Bug

The 0D flux-surface integrals behind `betat`, `betap*`, `betaj`, `li1/2/3`, and `volume` integrated only up to the last grid point (`psihigh`) and **dropped the edge slice out to the plasma boundary ψ=1** that Fortran `equil_out_global` includes. On the Fortran-matched DIII-D grid (`psihigh=0.993`) this made `volume`/`li2`/`li3` ~1.5–1.8% low and `betat`/`betan` ~1.5% high.

## Fix (minimal)

Replace the four trapezoidal sums with the **native FastInterpolations cubic integral** (already used elsewhere in this file) plus Fortran's linear edge slice to ψ=1 — a small local `fsi` closure, no new module-level machinery:

```julia
dpsi_b = 1.0 - xs[end]
function fsi(y)
    s = cubic_interp(xs, y; extrap=ExtendExtrap())
    return FastInterpolations.integrate(s, xs[1], xs[end]) + (y[end] + deriv1(s)(xs[end]) * dpsi_b / 2) * dpsi_b
end
```

Verified against the Fortran DIII-D KEFIT `dcon.out` (matched hamada/ldp grid, psilow=1e-4, psihigh=0.993, mpsi=128, mtheta=256): `volume` 0.02%, `betat` 0.015%, `li1` 0.09%, `li2/li3` ~0.5%, `betan` 0.34%. All remaining residual is a separate, pre-existing ~0.3% `crnt` edge-geometry difference (every leftover is a crnt/bwall-dependent quantity), out of scope.

## Regression

Tracks the previously-untracked 0D params (`li1/2/3`, `betap1/2/3`, `betaj`, `volume`, `crnt`, `bt0`, `bwall`, `aratio`, `kappa`) in `diiid_n1`. `regress --cases diiid_n1 --refs develop,local`: **10 changed, 36 unchanged** — only the flux-integration params move (≤0.12% on the shipped example, whose `psihigh=0.9995` shrinks the edge slice); all energies, q/pressure checksums, singular surfaces, Δ', and PE outputs are bit-identical.

## Also

Adds a short "Minimal-change discipline" note to `CLAUDE.md` (reuse native ops/existing utilities before writing new ones; size the change to the problem; no throwaway benchmark/agent-memory artifacts for minor fixes).


logan-nc:
@parkjk who found funny li values? I was only able to see ~1% differences with fortran and here I fix them to be <1%. I don't see any way the minor differences would have made "crazy" li unless the psilow or psihigh was 0.5 or something unusual like that. If your group is still or is ever finding odd results - be sure to post an issue so we can fix them!

jmlmir369:
@logan-nc 
I was the one who previously observed the odd li3 values. Upon checking, it seems the issue was due to my using an older version of GPEC. When I pulled the develop branch for ballooning on April 23, the li3 calculation was indeed incorrect, but it has since been resolved in the main branch. Thank you for benchmarking this

PR #296, opened 2026-06-22
EQUIL - BUG FIX - Newton separatrix finder for limited equilibria (#294)

### LINKED ISSUE #294: Julia direct equilibrium setup uses a different method for separatrix finding vs Fortran, resulting in some equilibria working in fortran but not Julia
Fortran uses regular Newton iteration which extrapolates as far as it needs to to find the separatrix. Julia uses a brent method with a maximum extrapolation distance of 50% of a. This makes some equilibria in the aspect ratio scan from tokamaker I used in this paper http://arxiv.org/abs/2606.02901 not work in Julia. 

The vibe-code-looking comment on Julia says this:

    # ψ = 0 at the separatrix (after renormalization), and ψ changes sign between the
    # magnetic axis (ψ > 0) and the region outside the plasma (ψ < 0), so Brent is
    # globally convergent within the bracket (start_r, end_r) and needs no restarts.
    # For limited equilibria whose grid box hugs the LCFS (e.g. some TokaMaker gEQDSKs),
    # ψ can stay positive out to the grid edge with no sign change to bracket. The separatrix
    # (ψ = 0) then sits just beyond the grid; extend the bracket outward via the extrapolating
    # ψ spline until ψ changes sign (capped), so rs is a faithful ψ=0 crossing rather than the
    # arbitrary grid edge (which lands in the degenerate ψ→0, Bp→0 region and breaks tracing).

However, clearly this is not working for all Tokamaker gEQDSKs as claimed. 

@logan-nc any thoughts? I have not dived into the implications of this decision or exactly how this affects the construction of the equilibrium.

logan-nc:
## Summary

Closes #294.

`direct_position!` in the Julia direct-equilibrium solver found the renormalized separatrix (the ψ=0 crossing along the midplane) with a **bracketed** `Roots.Brent()` solve over `(ro, rmin)` / `(ro, rmax)`. Bracketed Brent requires ψ to change sign *inside the grid box*. Limited equilibria whose computational box hugs the LCFS — e.g. some TokaMaker gEQDSKs from the aspect-ratio scan in [arXiv:2606.02901](http://arxiv.org/abs/2606.02901) — keep ψ > 0 all the way to the grid edge, so there is no interior sign change to bracket and equilibrium setup fails in Julia, even though the original Fortran handles them.

As @logan-nc noted on the issue, the range constraint was overdone in a blanket "replace manual Newton with standard root-finders" pass, and *"swapping to newton here is also fine — it's not a performance-deciding iterator."*

## Change

Replace the Brent separatrix finder with an **unbracketed Newton iteration** `dr = -ψ/ψ_r` faithful to Fortran `direct.f direct_position`, including its adaptive initial-guess restart loop (`ird = 0..5`). Newton extrapolates past the grid edge, so it finds the separatrix regardless of where the grid box ends. The implementation mirrors the O-point Newton loop already in the same function.

Single file changed: `src/Equilibrium/DirectEquilibrium.jl` (one block in `direct_position!`).

## Why it's regression-neutral

- `direct_position!` returns `(ro, zo, rs1, rs2)`. `rs1` is **discarded** by all callers.
- `rs2` is used **only as an initial guess**: `direct_fieldline_int` sets `r = ro + sqrt(psifac)*(rs2 - ro)` and then refines `r` with a Newton solve to the exact ψ₀. The converged flux surface is independent of the seed.

So Brent→Newton changes only an intermediate seed.

## Verification

- **Regression harness (local vs develop):** `diiid_n1` (EFIT — exercises this path) **46/46 OK**; `solovev_n1` **21/21 OK**. All diffs at floating-point noise (≤1e-13).
- **TJ-analytic unit tests:** 16/16 pass.
- **TJ-analytic ε scan** (`tj_analytic_direct`) up to the ideal-kink pole: all points solve.
- **Reviews:** fortran-physics-reviewer PASS (initial guesses match exactly for `ird = 0..5`; the unified Julia guess formula is algebraically/IEEE-bit-identical to the separate Fortran inboard/outboard forms); clean-code-reviewer no critical issues.

## ⚠️ Note — needs a test on a real failing case

I **could not reproduce the original failure** with inputs in the repo: the `LAR_epsilon_scan` (`tj_analytic_direct`) builds its own grid box with margin beyond the LCFS, so Brent always brackets — it passes on `develop` even hard against the pole. The bug specifically needs an input whose grid box **hugs the LCFS**, i.e. the TokaMaker EFIT gEQDSKs from the paper, which aren't in the repo.

@matt-pharr — could you run this branch on one of the aspect-ratio-scan gEQDSKs that **failed in Julia but worked in Fortran** and inspired the issue, to confirm it now constructs and traces correctly? If you can share one such gEQDSK, I'll add it as a regression case so this edge-hugging-separatrix path is tracked going forward.


jhalpern30:
I had also encountered this and had a similar fix on the backburner. I just went ahead and tested it and everything looks good

Testing with `eq_eps0.100_k1.000_d0.000.geqdsk` from Matt's scan
Develop
```julia
┌ Info: 
│   Equilibrium
└ ----------------------------------------
[ Info: Forcing hamada coordinate jacobian exponents: power_*
[ Info: Processing EFIT g-file: /Users/jakehalpern/Projects/3DDCON/data/geqdsks_aspect/eq_eps0.100_k1.000_d0.000.geqdsk
[ Info: Parsed from header: nw=129, nh=257
[ Info: Magnetic axis found at R = 10.044, Z = -0.000
ERROR: ArgumentError: The interval [a,b] is not a bracketing interval.
You need f(a) and f(b) to have different signs (f(a) * f(b) < 0).
Consider a different bracket or try fzero(f, c) with an initial guess c.
```

This branch
```julia
┌ Info: 
│   Equilibrium
└ ----------------------------------------
[ Info: Forcing hamada coordinate jacobian exponents: power_*
[ Info: Processing EFIT g-file: /Users/jakehalpern/Projects/3DDCON/data/geqdsks_aspect/eq_eps0.100_k1.000_d0.000.geqdsk
[ Info: Parsed from header: nw=129, nh=257
[ Info: Magnetic axis found at R = 10.044, Z = -0.000
[ Info: inboard separatrix found at R = 9.013
[ Info: outboard separatrix found at R = 10.990
[ Info: Estimated separatrix log slope A = 0.100 from probe integrations at psi = 0.9970, 0.9985
[ Info: Auto-mpsi: N_core=10 + N_mid=63 + N_edge=8 = 81 (A=0.100, h_mid=0.0152, tau=0.001)
[ Info: Equilibrium construction completed in 9.474 s
```

matt-pharr:
@logan-nc Confirmed — this fixes it. I ran the branch against a spread of the TokaMaker aspect-scan gEQDSKs (the ones that motivated #294), loading each through `setup_equilibrium` → `direct_position!` on this branch vs. pre-fix `develop` (`c80d94f0`):

| gEQDSK (ε, κ, δ) | `develop` (Brent) | this PR (Newton) |
|---|---|---|
| eps0.0500 k1.00 d0.00 | **FAIL** | PASS |
| eps0.1136 k1.72 d0.36 | PASS | PASS |
| eps0.1667 k1.90 d0.45 | **FAIL** | PASS |
| eps0.2534 k2.03 d0.52 | PASS | PASS |
| eps0.3788 k2.12 d0.56 | **FAIL** | PASS |
| eps0.5282 k2.16 d0.58 | PASS | PASS |
| eps0.7500 k2.20 d0.60 | PASS | PASS |
| **summary** | **4 / 7** | **7 / 7** |

The failures on `develop` are exactly the bracketing problem you diagnosed:

```
ArgumentError: The interval [a,b] is not a bracketing interval.
You need f(a) and f(b) to have different signs.
```

i.e. ψ > 0 all the way to the grid edge, so bracketed Brent has no interior sign change. The Newton finder extrapolates past the edge and converges on all of them, and the 4 cases that already loaded on `develop` are unchanged — consistent with this only touching the `rs2` seed for the field-line trace.

One thing worth flagging: the failure set is **not monotonic in ε** (0.05 fails, 0.11 passes, 0.17 fails, …). Whether the box hugs the LCFS depends on how TokaMaker wrote each grid, not aspect ratio per se — so the original "limited large-aspect" framing is really "any TokaMaker box that hugs the LCFS."

Happy to add a regression case. `eq_eps0.0500000_k1.000_d0.000.geqdsk` is the cleanest reproducer (smallest, circular, fails on `develop`). I'll attach it / open a follow-up so this edge-hugging-separatrix path gets tracked — let me know if you'd rather it live as a fixture in the repo or as a downloaded artifact in the test setup.

logan-nc:
Thanks for confirming. 

I am ok with a path of least resistance here if it is pushing the newton iterator and just recording a downloadable g-file if anyone wants to trace things back. 

Honestly though, if I was feeling more on my high horse I would say the hard fail is actually my preferred behavior. I don't think anyone should be allows to give us g-files with a separatirx that extends beyond the grid. That's bad form and allowing cubic spline extrapolations beyond the grid has got to be setting ourselves up for trouble. For this reason I do not want such a bad example of inputs in our repo. I would also be perfectly happy with declaring those g-files not acceptable, closing this issue, closing the PR without merging, and kicking the issue to TokaMaker to be better in making g-files. 

I don't think the aspect ratio scan is part of our test set or anything, right? Just something you are doing outside the repo? As I write this, it seems like option 2 is best - it's a user problem, not a GPEC problem.  I leave it up to you though @matt-pharr

logan-nc:
I might ask @jhalpern30 to weigh in on the above too: In your opinion, is this a GPEC or TokaMaker issue?

matt-pharr:
@logan-nc The only thing 'wrong' with the failing equilibria is that the poloidal flux ends at the edge of the grid the same sign as it is at the magnetic axis. 
image
image

I support merging this as is. This is just a problem specific to fixed-boundary equilibria it seems like, and I am not sure what would need to be changed about tokamaker to prevent this, or if that would even be physical.

logan-nc:
👍 go for it. 
Thanks for the clarification - plots are great for the records!

matt-pharr (approved):

PR #298, opened 2026-06-22
ForceFreeStates - FEATURE! - Cap the integration domain where resistive layers overlap

amlakbekaluachule:
## Release note

- **Audience:** users
- **Numerical impact:** none on any shipped deck _(harness @ a3e7c51fa)_: 181 quantities across the 14 standard cases, none moved beyond the known SLAYER γ root-search jitter. Results change in two situations no shipped deck exercises: (1) runs with `set_psilim_via_dmlim = true` and a finite `qhigh`, where `dmlim` could previously step past `qhigh` (`qhigh = 4.0` gave `qlim = 4.2`, now `3.2`); (2) real reconstructions on `grid_type = "auto"` that the old edge q-law gate wrongly rejected, which now get the edge knot-density floor.
- **Migration:** a deck with `dmlim` on and a finite `qhigh` now truncates one rational lower whenever `qhigh` sits on a rational or less than `dmlim/n` above one; raise `qhigh` above `(rational + dmlim)/n` to keep the old domain. `LayerWidths` gains `delta_norm`, `delta_visco` and `delta_dr`, so code constructing it positionally must supply them. `sing_lim!` gains an optional `psilim_cap` keyword and is otherwise unchanged.

GPEC can now locate the point past which adjacent rational surfaces' resistive layers overlap — the outer limit of the matched-asymptotic treatment, after Fitzpatrick, Nucl. Fusion 2025 Sect. 5.9 — and, when `psilim_from_layer_overlap = true`, use it as an upper bound on the stability integration domain `psilim`. The overlap point is always recorded, and GPEC warns whenever `psilim` lands past it. The truncation logic is documented in the new "Integration domain truncation" section of the stability docs.

## What is actually wired

| piece | where |
|---|---|
| opt-in control | `[ForceFreeStates] psilim_from_layer_overlap` (default `false`) |
| the scan | `Tearing.resistive_layer_overlap`, run in `prepare_force_free_states!` whenever kinetic profiles are readable, once per `n` in `nn_low:nn_high`; the innermost overlap point binds |
| the cap | `sing_lim!(...; psilim_cap=...)` — an upper bound on `qlim`, applied before `dmlim`, which can no longer step past it |
| the record | `ForceFreeStates/LayerOverlap/` — always written when the scan ran |
| the warning | fires whenever `psilim` lands past the overlap point, flag on or off |

The scan runs regardless of the flag; only *applying* the cap is gated, so a run always shows where layer physics would have cut the domain. Profiles come from whichever of `[KineticForces] kinetic_file` or `[SLAYER] profile_file` exists.

## Why it stays off by default, and why the DIII-D decks keep their `psihigh`

On the DIII-D-like equilibrium the Eq. (100) overlap point is ψ ≈ 0.9997 (0.99969 at `psihigh = 0.995`, 0.99975 at 0.9999 — it shifts slightly as edge surfaces move from extrapolated to grid-resolved). That is beyond every shipped deck's `psihigh`, so on every shipped deck the cap sits outside the domain and **cannot bind**: turning the flag on changes nothing.

Raising the ideal and riccati decks to `psihigh = 0.9995`, as suggested, does not change that — the cap is still outside the domain (measured: all 47 `diiid_n1` and 18 `diiid_n1_riccati` quantities identical flag on vs off at 0.9995). It does move results substantially, because the domain gains the q = 7 surface:

| `diiid_n1`, 0.995 → 0.9995 | |
|---|---|
| `msing` / `mpert` | 5 → 6 / 35 → 36 |
| `pe_toroidal_torque` | −0.0506 → +0.0492 (sign flip) |
| `et_imag` | +1.23e-4 → −1.71e-6 (sign flip) |
| `ep_real`, `ev_real` | −37 %, −24 % |
| NTV torque, resonant field | −26 %, −21 % |
| `q0`, `q95` | unchanged to 8 figures |

So that change is a full golden re-pin that does not activate the feature, at a runtime cost on the Riccati deck; it is **not** made here. Where the cap does bind is covered instead by a dedicated deck (below).

## Verified end to end

`examples/DIIID-like_truncation_example` runs the DIII-D-like equilibrium at `psihigh = 0.9999` with the cap on — ideal fixed-boundary stability only, about two minutes:

```
Resistive-layer overlap caps the domain: qlim 8.352 -> 8.000 (psi 0.999900 -> 0.999755)
Setting psilim via dmlim: initial qlim = 8.000, dmlim = 0.200
Filtered 1 singular surface(s) outside integration domain: [([8], 8.0)]
qlim = 7.200, psilim = 0.999
```

The new harness case `diiid_truncation_n1` tracks that run, including the `LayerOverlap` outputs (`applied`, `psilim_overlap`, `first_overlap_index`, `width_dr`) that no case tracked before.

**Correction to an earlier version of this description**, which said `dmlim`'s `trunc` "backs off to the rational below" the cap. It did here, but only because `q(psilim_cap)` happened to evaluate to 7.99999…: the cap lands exactly on a rational by construction, and a value rounding the other way gave `qlim = 8.2`, i.e. past the cap. Fixed in `14d60b6b9` (below); the capped domain no longer depends on that rounding.

## The domain-limit convention

The scan cuts at ψ[k−1], the position of the last surface with a well-separated inner region. This is the paper's own `1 − ε_c`, and it is what `runtests_layer_overlap_fitzpatrick.jl` reports in order to reproduce the paper's Ψ = 0.9985 / 0.9952.

Settled against ten equilibria (the shipped deck plus nine IDA DIII-D reconstructions) rather than by preference — the two half-width alternatives both fail:

| convention | result |
|---|---|
| ψ[k−1] − w/2 (the original) | drops the last surface the criterion found clean; half a layer width more conservative than the paper |
| **ψ[k−1]** | always inside the plasma, always inside the first contaminated surface |
| ψ[k−1] + w/2 | **exceeds ψ = 1 on 4 of 10**, and on those same 4 overshoots surface k itself |

Full per-case table in the thread.

## Bugs found while making the feature trustworthy

**1. The edge q-law gate rejected 8 of 9 real DIII-D reconstructions** (`d4ce058ff`, carries `!`). `EDGE_FIT_MIN_R2 = 0.99` was calibrated against the synthetic TokaMaker deck (r²_log 0.9957, 0.9989). Real EFIT/CAKE reconstructions carry noise in the edge q and fit at 0.929–0.977, so all were refused: no surfaces placed past the grid, and the scan reported "no overlap" — a silent fail-open on exactly the case the criterion exists for. No absolute threshold can separate the classes (the limited a10 deck sits at 0.972, inside the diverted band); the `r²_log > r²_linear` comparison does, and is unchanged. Lowered to 0.90 as a weak floor rather than removed. The `!` is because the same gate controls the auto-grid edge density floor.

**2. Reverse shear crashed the layer build** — split out as #431, being pre-existing on develop. #431 has since merged; the cherry-picked copy resolved away in the develop sync merge (`7e9e66508`).

**3. A single failing surface aborted the whole scan** (`6781b639f`). The per-surface guard caught only `ArgumentError`.

**4. The no-surfaces path threw instead of returning an empty scan** (`c02a3c43d`, found in adversarial review). The early return built `LayerOverlapScan` with 13 arguments for a 15-field struct — a `MethodError` the pipeline's blanket catch downgraded to a spurious "scan failed" warning. Fixed, with a regression test forcing the path (`m_max=0`, extrapolation off), and the catch now logs the backtrace so a programming error is distinguishable from a physics refusal.

**5. `dmlim` could step past `qhigh` and the cap** (`14d60b6b9`, carries `!`). `sing_lim!` stepped the `(last rational + dmlim)/n` offset back only while it exceeded `qmax`, so `qhigh = 4.0` gave `qlim = 4.2`, and a cap on a rational could be exceeded depending on float rounding. It now steps back against `min(qmax, qhigh, cap)`: `qhigh = 4.0` and `4.1` both give `3.2`, and a cap on the exact rational and 1000 ulps either side all give the same `qlim`. 8 of the 12 new assertions fail against the old step-back. The existing cap-wiring tests all ran with `set_psilim_via_dmlim = false`, which is why this was not caught.

**6. The scan's catch-all hid code errors** (`8cfbbee7e`). `resistive_layer_overlap` already records expected per-surface failures in its notes and rethrows anything else; the outer catch downgraded those to a warning (bug 4 was found through exactly that). Only the kinetic-file read is guarded now.

## Jacobian independence

@logan-nc asked for confirmation that `delta_s` gives a Jacobian-independent ψ width. Measured under `pest` / `hamada` / `equal_arc`: `r_s` invariant to 1e-16, `da/dψ` to 3e-14, `|δ_s|` and both Δψ widths to ~7e-6 (the equilibrium solve's own noise — the inputs are invariant to 1e-14). Same under `:flux`. Confirmed clean.

The check did find that `:fsa` and `:halfwidth` are coordinate-dependent (up to 7% in `r_s`); those labels live in #403, are not defaults, and are not used here.

## Response to the review-package questions (@logan-nc, 2026-09-15)

| question | outcome |
|---|---|
| convention vs `dmlim` | kept as agreed; `dmlim` now provably stays inside the cap (bug 5) |
| two width channels | Eq. (100) kept as the criterion; the |δ_s| channel is still reported for comparison |
| shipped decks stay a no-op | yes — flag off everywhere; the `psihigh = 0.9995` harmonisation is declined for the reasons above |
| Solovev auto-grid classification | added to `runtests_grid_refinement.jl` (`9f2010cfd`): Solovev is rejected as limited on both passes, including with the absolute r² floor disabled, and the DIII-D-like equilibrium is accepted as a positive control |
| always-on cost | kept; the scan is ~15 layer solves per `n` |

## Notes for reviewers

- **The scan is driven at `rs_method = :flux`**, not the SLAYER default. Eq. (100) is not covariant — δ ∝ `(q/|dq/dr|)^(1/2)` — so it is only valid in the coordinate it was derived in, the toroidal-flux label of the paper's Eq. (30).
- **`delta_s_abs` appears in a third HDF5 group.** The one-name-per-quantity rule says this is right, but @logan-nc asked to be told if it stayed ambiguous.
- **Still open, not gated on this PR:** the two width channels (|δ_s| Riccati vs Eq. 100) disagree by about two surfaces (0.99857 vs 0.99969 on the DIII-D-like deck). Both are reported.
- **Seen in passing, not addressed:** the fixed-boundary path (`vac_flag = false`) reports a non-Hermitian W-inverse at ~40 integration steps on the DIII-D-like equilibrium at any `psihigh`. Pre-existing and invisible on the shipped decks, which all run `vac_flag = true`; the truncation deck now does too, so no case exercises it.

## Regression report

`regress --cases <all 14 standard cases> --refs 6e3270881,a3e7c51fa --force` — develop at the sync merge vs this head; julia 1.11.6, pinned manifest:

```
diiid_n1                    Summary: 47 unchanged
diiid_n1_riccati            Summary: 17 unchanged
diiid_slayer_n1             Summary: 1 changed, 16 unchanged
                            SLAYER γ_Hz [2/1,3/1,4/1]  2.133e-01 (0.02%)  ** CHANGED **
efit_fixedbdy_separatrix    Summary: 5 unchanged
gal_resistive_diiid         Summary: 10 unchanged
gal_resistive_pe            Summary: 8 missing/N/A   (on both refs; see below)
ggj_ray_q500i               Summary: 4 unchanged
ggj_reference               Summary: 4 unchanged
solovev_kinetic_calculated  Summary: 14 unchanged
solovev_kinetic_multiion    Summary: 6 unchanged
solovev_kinetic_ntv         Summary: 6 unchanged
solovev_kinetic_nuzero      Summary: 14 unchanged
solovev_multi_n             Summary: 15 unchanged
solovev_n1                  Summary: 22 unchanged
```

181 quantities compared, 180 unchanged. The one change is the threaded dispersion root search's known reproducibility step: every SLAYER layer input on that case is identical, and the same signature (0.076-0.145 Hz) has been measured on same-source re-runs; this run's 0.213 Hz is above that band, so it is worth re-measuring if it grows.

`gal_resistive_pe` extracts nothing on **either** ref, develop included, so it predates this PR; it is being looked at separately. The case runs, but none of its `SingularCoupling` datasets are found.

Runtimes are not quoted: the machine was under heavy external load during this run (load average up to 38 on 8 cores), which slowed random cases on both refs by 10-15x in both directions.

`diiid_truncation_n1` is new, so it runs at this head only:

```
Ref 1: a3e7c51fa
layer-overlap cap applied              1
layer-overlap psilim                   9.997552e-01
first overlapping surface index        8
diffusive-resistive layer widths (Δψ)  [15 elem]
qlim                                   7.200000e+00
psilim                                 9.989203e-01
mpert                                  37
# singular surfaces                    6
singular q values                      [6 elem]
singular psi locations                 [6 elem]
plasma energy Re(ep[1])               -7.958321e-01
total energy Re(et[1])                 8.676963e-01
```

## Changes from adversarial review

- **The criterion no longer depends on the Riccati solve** (`cd6dce306`, carries `!`). The scan
  called `slayer_layer_thickness`, which always runs the stiff `riccati_del_s` solve, and dropped
  any surface whose `delta_s` came back non-finite -- even though the criterion uses only the
  closed-form `delta_dr`. A deleted surface left `_first_overlap_limit` comparing its neighbours
  across a gap twice the physical one, hiding an overlap and reporting too wide a domain, silently.
  `delta_s` is now best-effort (`NaN` plus a note, its own channel only); a surface that cannot be
  scored on the DR channel is skipped with an explicit "not conclusive" note, matching the `m_max`
  and `max_layer_solves` backstops. Measured unchanged: `psihigh = 0.9997535788` over 14 surfaces,
  no notes, 3.9 s per `n`.
- **The truncation case pins physics again** (`6de1b8f03`). It ran `vac_flag = false`, which writes
  no energies, so only bookkeeping was tracked on the one deck where the cap binds. The deck now
  runs free-boundary and the case pins `ep_real` and `et_real`.
- **`refined_psi_grid`'s `psihigh` keyword and `_truncate_density` are removed** (`9848fa332`) --
  left from the design that re-formed the equilibrium, unreachable and untested since the `qlim`
  cap replaced it. (The Migration note below no longer mentions them.)
- **The flag is inert on the programmatic `solve(prob, alg)` path**, which resolves no kinetic
  file; documented in the control docstring and the stability page (`a3e7c51fa`).
- `LayerOverlapScan` gains a keyword constructor -- the positional form takes 15 arguments in a
  fixed order, which is how the earlier empty-scan return came to pass 13.

## Validation

- [x] Develop merged into this branch at `620e4c531`, no conflicts
- [x] `runtests_layer_overlap_fitzpatrick.jl` at this head: JET reproduction 9/9 (Ψ = 0.9985111 vs paper 0.9985 for n = 1, 0.9949738 vs 0.9952 for n = 4, plus the no-ODE width path asserted equal to the full one), cap wiring 28/28
- [x] `runtests_grid_refinement.jl` at this head: 65/65
- [x] Multi-n end to end on the DIII-D-like equilibrium at `psihigh = 0.9999`: n = 1 overlaps at 0.99975, n = 2 at 0.99935, `nn = 1:2` binds at n = 2
- [x] Docs build locally with the new cross-references resolving
- [ ] Full suite: left to CI on this push

> [!IMPORTANT]
> Do not merge without human review.


amlakbekaluachule:
@logan-nc this PR does two things:

1. Fixes a bug in slayer_layer_thickness (LayerThickness.jl) 
2. Adds set_resistive_width_based_psihigh (new file,SetResistiveWidthPsihigh.jl) 

Let me know if anything looks off.

d-burg:
I synced this branch with `feature/tearing-growthrates` (merge `8a771555`) and retargeted the PR base from `develop` to `feature/tearing-growthrates`. The diff now shows your 2 files instead of 65.

Worth knowing why the retarget was needed: `LayerThickness.jl` doesn't exist on `develop` at all — the whole SLAYER module is tearing-branch work, and `develop` only carries a `Slayer.jl` stub. So the PR couldn't have merged there regardless of how clean it was.

The merge hit two conflicts, both caused by `src/Tearing/InnerLayer/SLAYER/` being relocated to `src/InnerLayer/SLAYER/` on the parent branch (the `InnerLayer` module had to move above `ForceFreeStates` to break a dependency cycle). Nothing to do with your changes — resolved keeping your `δ_FKR` / `δ_visco` constructor arguments. The package precompiles and the SLAYER tests pass on the merged branch.

Two things worth a look before review:

### 1. `SetResistiveWidthPsihigh.jl` is never loaded

`SLAYER.jl` includes `LayerParameters.jl`, `Riccati.jl`, `LayerThickness.jl` and `LayerInputs.jl` (lines 51–54) but not `SetResistiveWidthPsihigh.jl`, so the helper isn't reachable:

```julia
julia> using GeneralizedPerturbedEquilibrium
julia> isdefined(GeneralizedPerturbedEquilibrium.InnerLayer.SLAYER, :set_resistive_width_based_psihigh)
false
```

Adding `include("SetResistiveWidthPsihigh.jl")` alongside the others should wire it in, plus an `export` if it's meant to be public API.

### 2. `δ_FKR` / `δ_visco` aren't in the `LayerWidths` docstring

The `# Fields` list in `LayerThickness.jl` (lines 133–139) still stops at `d_beta`. The repo convention is to document struct fields in the docstring rather than as inline comments, so these want a line each saying what they are and their units.

Minor: I also stripped some trailing whitespace from both files while resolving the merge — the pre-commit hygiene hook rejects it.

logan-nc:
@d-burg whats the status of this? Should it be reviewed now? Or should it still be marked as draft?

amlakbekaluachule:
Fixed both flagged issues — SetResistiveWidthPsihigh.jl is now included/exported, and δ_FKR/δ_visco are documented in the LayerWidths docstring.

logan-nc:
@claude review this

logan-nc:
## Status

Rewritten from the original 172-line contribution, which had zero call sites and compared layer widths in metres against positions in normalized flux. That file is gone.

**Feature** — `[Equilibrium] psihigh_from_layer_overlap`, **opt-in, default off**. Forms a first-pass equilibrium, finds where adjacent rational surfaces' resistive layers overlap, checks the reconstruction can actually deliver the resulting domain, and re-forms. Floored by the rational-q window containing `psiedge` so the dW scan band stays open. The scan runs whenever kinetic profiles are readable regardless of the flag, so `gpec.h5` always records `Equilibrium/psihigh_resistive_layers` beside `psihigh_resolved`, plus per-surface widths under `Equilibrium/ResistiveLayers/`.

Default is off because enabling it crosses a rational surface on a diverted plasma: on `DIIID-like_ideal_example` the domain goes 0.995 → 0.993281, 5 surfaces → 4, `mpert` 35 → 34, moving `et[1]` 76% and NTV torque ~1800%. That is the criterion working, not a defect, but not a change to make on every user's behalf silently.

**Nine defects fixed along the way**, all found by trying to make the feature trustworthy:

| defect | consequence |
|---|---|
| field-line ODE `retcode` never checked | truncated solves consumed as closed flux surfaces — silently wrong equilibria |
| round-trip check sampled only at ψ knots | structurally blind to inter-knot ringing; passed while `et[1]` reached −47707 |
| `surface_da_dpsi` stencil clamped at 1−1e-4 | the real limit on near-boundary accuracy, not `psi_cap` |
| `run_slayer` passed `b0exp` as B_T | Lundquist number off by ~95% — split out as #399 |
| trial formations unguarded | crash instead of back-off, `psihigh` left corrupted |
| round-trip unavailable on the `efit` path | the path every shipped deck uses |
| gpec.h5 replay crash | stored relative profile paths don't resolve in the output dir |
| edge q-law duplicated, applicable to limited plasmas | one shared helper, with a guard that actually discriminates |
| inverse equilibria extrapolated | fabricated 3 non-existent surfaces on the CHEASE fixture |

**Verification** — 122 + 67 + 59 + 32 + 14 tests; docs clean; the metres→flux conversion triple-checked three independent ways (the binding pair is 1.821 mm apart with half-widths summing to 2.338 mm, argued in pure metres); the back-off gate validated against ForceFreeStates energies as ground truth, and it flips one step *before* they degrade.

**Still open**
- Final frozen six-case regression. `diiid_slayer_n1` needs re-pinning — **for the #399 `bt` fix, not for this feature.**
- Full suite against the current tree; the IMAS deck has never been run with the flag on (left off, with a comment explaining why).

logan-nc:
@d-burg — ready for review, but **please take #399 first**.

#399 is the one-line `bt` fix (SLAYER was running at `b0exp` = 1.0 T instead of the physical ~1.95 T, so the Lundquist number was off by ~95%). It is cherry-picked here as 4304caad, so once #399 merges this branch's copy resolves cleanly and the diff shrinks to this PR's own work.

That ordering also separates the regression movement. `diiid_slayer_n1` needs re-pinning, and with #399 landed first it is unambiguous which part is the `bt` correction and which is this feature — measured separately, #399 alone accounts for S +94.5%, D_norm 35%, γ 4.4%.

Worth your attention specifically:

- **The `psiedge` floor governs every DIII-D deck**, so the layer criterion never actually sets the domain in the shipped examples — each asks for ~0.9892 and is held at the q=6 window top. Working as specified, but you may want a deck that exercises the physics path.
- **The `bt` fix changes the layer widths themselves**, so any `Tearing/LayerWidths/` numbers you have from before #399 were computed at half the true field.
- **`delta_s_abs` now appears in two h5 groups.** They agree to ~1e-6 and the difference in scope is documented in `hdf5-conventions.md`; flag it if that still reads as ambiguous.

Everything except the opt-in flag is a bug fix that stands on its own — the field-line ODE `retcode` was never checked, and the round-trip gate was structurally blind to spline ringing (it passed while `et[1]` hit −47707).

logan-nc:
@d-burg, as claude says this has hit a point where it needs your expert review. Workflows are in place to decide the truncation from the resonant layer widths. The issue is that it is reporting big widths and wanting to truncate way in (~0.9892 for DIII-D). I find this suspect, since Fitzpatrick had been finding things like to be very far out past 0.9999. In the process of adding the truncation logic, claude did convince me the units of the layer widths had been wrong in the original overlap-checking-loop (comparing layer widths of m to rational spacing in psin) and a conversion from m to psin was added... so maybe double check that? 

Do you think we need to try and reproduce the exact case from Fitzpatricks paper on this as a sanity check?

d-burg:
the `n_kept < 2` refusal and the `psiedge` floor are all real and do what they claim. Four findings,
one blocking, plus an answer to your question about the widths.

## 1. Blocking: the trial/back-off loop cannot compile on current `develop`

The mechanism mutates the config in place — `eq_config.psihigh = hi` in the back-off loop, and again
on the `psiedge`-floor path. #367 froze `EquilibriumConfig`: `develop` has `@kwdef struct`, this
branch still carries `@kwdef mutable struct`, which is why `EquilibriumTypes.jl` is one of the four
conflicts. This is not a cosmetic rebase — mutation *is* the mechanism here.

The idiom to move to is your own `31df659f`: carry the dict and rebuild per trial,

```julia
cfg = EquilibriumConfig(merge(equil_dict, Dict{String,Any}("psihigh" => hi)), dir)
```

which also removes the "psihigh must be restored, since it was mutated before the attempt" failure
mode the loop currently has to hand-manage in its `catch`.

## 2. The width solver is correct — the criterion is what produces ~0.9892

Taking your question about the widths first: I validated `riccati_del_s` against the textbook and it
reproduces it. GPEC already integrates Fitzpatrick's layer equations verbatim — its `E` and `F` match
(6.29)/(6.30) term-for-term, `dW/dq = W/q - W^2/q + qE/F` is (6.38), and it returns `dels_db`, which
is literally the plotted δ_s/d_β. Driving it at prescribed normalized parameters (`D_norm = 1`,
`P_perp = P_tor = P̂`, `Q_e = Q̂*/(1+1/τ)`, so the code's internal `Q_hat` equals the book's Q̂*)
reproduces **figures 6.2 and 6.3** over Q̂* ∈ [0,4], P̂ ∈ [0,4]. Eight spot checks, all inside the
published contour bands:

| (Q̂*, P̂) | Re GPEC | fig 6.2 | Im GPEC | fig 6.3 |
|---|---|---|---|---|
| (0.25, 0.25) | 0.818 | ~0.8–0.9 | −0.222 | ~−0.15…−0.3 |
| (0.5, 0.5) | 0.943 | ~0.9–1.0 | −0.354 | ~−0.3…−0.45 |
| (2.0, 0.5) | 1.351 | ~1.3–1.5 | −1.417 | ~−1.2…−1.35 |
| (4.0, 0.5) | 2.008 | ~1.9–2.1 | −2.225 | beyond −1.8 |
| (0.5, 4.0) | 1.537 | ~1.5 | −0.119 | ~−0.15 |
| (4.0, 4.0) | 1.350 | ~1.3–1.4 | −1.466 | ~−1.35…−1.5 |

(τ = 1 assumed — the caption does not state it; the agreement is the evidence.)

The widths on the actual DIII-D surfaces are also not anomalous. Table 6.2 of the book quotes
|δ_s| = 4.69 mm (low-field reactor) and 2.95 mm (high-field); ours span 1.8–4.4 mm:

| m | ψ | r_s [m] | \|δ_s\| [m] | δ_visco [m] | S | δ_visco/δ_s |
|---|---|---|---|---|---|---|
| 2 | 0.518 | 0.362 | 3.91e−3 | 2.81e−3 | 1.9e7 | 0.720 |
| 3 | 0.770 | 0.458 | 4.13e−3 | 2.85e−3 | 3.6e7 | 0.689 |
| 4 | 0.893 | 0.504 | 3.85e−3 | 2.62e−3 | 5.5e7 | 0.680 |
| 5 | 0.968 | 0.531 | 4.42e−3 | 2.74e−3 | 3.3e7 | 0.620 |
| 6 | 0.993 | 0.540 | 2.85e−3 | 1.76e−3 | 5.4e7 | 0.615 |
| 7 | 0.999 | 0.542 | 1.82e−3 | 1.20e−3 | 1.4e8 | 0.661 |

**So where does 0.9892 come from?** Between m=6 (ψ=0.99328) and m=7 (ψ=0.99851) the spacing is
Δψ = 0.0052, while dr/dψ ≈ 0.35 m there, so the two layers occupy δψ ≈ 0.0081 and 0.0052. The
half-widths sum to 0.0067 > 0.0052 — overlap — and cutting at ψ(6) − w(6)/2 gives **0.9893**, the
number you are seeing. Nothing is wrong with the width; the surfaces are simply 5 mΨ apart out there.

**Two notes on `delta_visco`.** It is a consistent 0.62–0.72 × δ_s across the whole domain, so it is
the *narrower* width, overlaps later, yields a *larger* psihigh, and therefore `minimum(cands)` never
selects it — the viscous criterion is effectively dead code on this deck. And since it is algebraic
(`δ_norm·P_perp^(1/6)`) rather than from the Riccati solve, it is the weaker of the two anyway. Worth
deciding whether it earns its place. (`delta_FKR` is a non-issue: it was renamed `delta_norm` and is
not used as a criterion — the primary width is already the validated Riccati δ_s.)

## 2b. Why the criterion is nonetheless close to self-fulfilling

This is where I think your suspicion about ~0.9892 is right, and I do not think the metres→ψ_N
conversion is the culprit.

Past the last real surface the edge law is `q ~ -A·ln(1-ψ)`, so surfaces sit at
`ψ_m = 1 - exp(-m/(nA))` and their spacing shrinks **geometrically** as ψ → 1. Meanwhile the layer
width does not shrink: toward the edge `T_e` falls, η rises, `S` drops, and `δ_s ∝ r_s S^(-1/3)`
*grows*. A shrinking gap against a growing width means overlap is guaranteed at some radius for any
equilibrium — it is a property of extrapolating toward a separatrix, not a measurement of this
plasma.

So "where do layers first overlap" is set by where the geometric spacing crosses the width, and both
inputs there are extrapolations: `A` from `edge_q_law` fitted on the outer knots, and the kinetic
profiles continued past the grid. That is consistent with what you are seeing — the criterion always
wants to cut in aggressively, and the `psiedge` floor ends up binding on every shipped deck.

Fitzpatrick's ψ > 0.9999 is not really the same quantity: TJ places its outermost surface of
interest, it does not ask where layers first touch under an extrapolated q. I would expect these to
disagree by construction.

## 3. The viscous criterion fails open

`da_dpsi` (`LayerOverlap.jl:257`) and `delta_s_m` (`:277`) are both finite-checked, each recording a
note and `continue`-ing. `delta_visco` is not — `:290` pushes `lw.delta_visco / abs(da_dpsi)`
unguarded. A NaN there makes every comparison in `_first_overlap_limit` false, so the viscous scan
silently returns "no overlap" with no note, and `recommended = minimum(cands)` quietly loses its
more conservative half. The bias is outward, which is the unsafe direction, and it is the one path in
this file that does not follow its own record-and-skip discipline.

## 4. The scan's cost is bounded only by `m_max = 2000` layer solves

Every surface in `found` runs `build_slayer_inputs` + `slayer_layer_thickness`, and the latter is a
Riccati solve. The extrapolation terminates on `psi_cap`, but how many surfaces that takes depends on
`A`; a slowly-rising edge law reaches the cap late. Worst case is 2000 Riccati solves — and since the
scan runs whenever kinetic profiles are readable, regardless of the flag, that cost lands on runs
that never asked for the feature. Bounding the extrapolated-surface count (or the work) rather than
only the `m` index would make it predictable. Worth quoting the added runtime on a default deck.

## On reproducing the Fitzpatrick case

Done — see §2. It was cheaper than expected because the solver already implements the book's
equations; no new physics was needed, only driving it at prescribed normalized parameters. I would
not chase the width formula any further: it matches the published figures *and* the published
magnitudes.

I have also pushed two fixes to this branch (0c2b3ac0): `delta_visco` now gets the same
finite-check-and-note treatment `delta_s_m` and `da_dpsi` already had (it was the one path that could
fail open), and the layer-solve count is bounded with a note when the bound bites. 122/122 tests pass.

One coverage note on your own observation that the floor governs every DIII-D deck: as it stands the
physics path is never exercised end-to-end by a shipped deck or by `runtests_layer_overlap.jl`, which
tests the pieces (conversion, q = m/n, the cut rule, the limited-plasma refusal) rather than a run
where the criterion actually sets the domain. A deck that reaches the physics path — or a test that
drives `_choose_psihigh` with a synthetic scan — would close that.

Also: this is `dirty` against `develop` (4 conflicts). In `test/runtests_slayer_runner.jl` the
cherry-picked `4304caad` collides with #399 as merged — #399's source-text test was replaced with a
behavioural one in `runtests_slayer_inputs.jl`, so take `develop`'s side and drop the cherry-pick.

⚠️ Not an approval — needs a third-party human review before merge.

logan-nc:
Clean out any approximate visco layer widths from the overlap criteria - those and any other limit approximations can be output in the `Tearing` section of SlabLayer outputs if flagged on. The Equilibrium truncation should use the general model.

The Fitzpatrick model this is based off of is https://iopscience.iop.org/article/10.1088/1741-4326/ae4fdd/meta (also at  https://arxiv.org/pdf/2511.07666 if you hit a paywall), NOT the book (which is a fine benchmark for the layer calculations but doesn't have the truncation logic in it). Figure 10 shows an example for typical JET parameters that finds 0.998 and Fig. 11 for n=4 finds 0.9938. @d-burg can you have your AI add this paper to the resources then cite it in annotations and the docs where appropriate? Might as well also have it double check it's own logic above based on the text... This feature is not much help if it regularly suggests truncating deep into the pedestal 😛  Maybe it gets better (further out) for an ITER case?

d-burg:
@logan-nc ahh sounds good, I shall dig into this

docs/development/references.md
```diff
@@ -100,6 +100,24 @@ The `InnerLayer` inner-layer solvers and the `Tearing` stack that drives them ar
     semi-collisional (Eq. 12), and resistive-inertial (Eq. 13). The `P^(1/6)` factor of Eq. (11)
     is the `delta_visco` scale in `LayerWidths`.
 
+- **Fitzpatrick (2025)**: "Response of a magnetically diverted tokamak plasma to a resonant
+  magnetic perturbation"
+  - Location: `docs/resources/2025-Fitzpatrick-Response_of_a_magnetically_diverted_tokamak_plasma_to_a_resonant_magnetic_perturbation.pdf`
+  - Published: Nucl. Fusion (2025), doi 10.1088/1741-4326/ae4fdd; arXiv:2511.07666
+  - Describes: **the source of the edge-truncation criterion**. q diverges logarithmically at the
```
logan-nc: @d-burg this is way longer than the other one-liner descriptions in this md. One liner goes here. The actual docs webpage should describe the layer widths and truncation logics and cite this paper using cross references.

docs/development/references.md
```diff
@@ -100,6 +100,24 @@ The `InnerLayer` inner-layer solvers and the `Tearing` stack that drives them ar
...
+    separatrix, so rational surfaces accumulate without bound; the paper shows their spacing
+    shrinks *faster* than the resistive layer width, so the layers merge in a narrow region
+    straddling the separatrix. Only surfaces with `0 < Ψ < 1 - ε_c` need be retained, the rest
+    treated as vacuum — which is exactly the empirical truncation existing response codes
```
logan-nc: "which is exactly" is sooooo AI 🙄 
also, no other existing response code actually uses this that I know of

docs/development/references.md
```diff
@@ -100,6 +100,24 @@ The `InnerLayer` inner-layer solvers and the `Tearing` stack that drives them ar
     semi-collisional (Eq. 12), and resistive-inertial (Eq. 13). The `P^(1/6)` factor of Eq. (11)
     is the `delta_visco` scale in `LayerWidths`.
 
+- **Fitzpatrick (2025)**: "Response of a magnetically diverted tokamak plasma to a resonant
```
logan-nc: In-line annotations of the layer widths and/or truncation logic in src should be citing this as per the claude.md instructions

d-burg:
@logan-nc read the paper — agreed it changes the picture, and with its width the truncation lands
where you expected instead of in the pedestal.

Sect. 5.6 shows the strong shear near the separatrix forces every resonant layer into the
diffusive-resistive regime, where the layer equation solves analytically: Eq. (100) is a closed-form
width, no ODE integration involved. The 2023 book's general two-fluid Riccati solve (what the scan
uses today) remains a good benchmark of the layer solver — it reproduces the book's figs 6.2/6.3 —
but the book has no truncation logic, and the paper's ε_c values are calibrated in the paper's own
width convention, so the criterion should use Eq. (100) verbatim. The strongest reason: the paper's
fig 9 cross-validates the overlap point against an independent criterion — |Δ_k|/(−E_kk) dropping
through ~1, where the shielding response actually becomes vacuum-like — and the two coincide; that
validation is attached to Eq. (100)'s width definition. Empirically the two widths differ by a
near-uniform factor ~2 on this deck (Eq. 100 ≈ 0.5 × Riccati across m=2–10): the ODE width is
defined through the dispersion relation (the equivalent resistive-sheet thickness, book Eq. 6.24),
Eq. (100) is the spatial decay scale of the layer solution, so a constant O(1) offset is expected —
and that factor is what separates the two truncation points below.

One trap for the implementer: the paper's S = τ_R/τ_A uses the shear-free τ_A of its Eq. (74), while
GPEC's `lu` carries the shear via `tau_h ∝ 1/(n·sval_r·bt)`. So `lu = (n|s|)·S_paper` and Eq. (100)'s
explicit `(n|s|)^(1/2)` cancels, leaving `δ_DR = r_s^(3/2)·P_perp^(1/4)/(lu^(1/2)·d_β^(1/2))` in GPEC
variables (`tau_r`, `P_perp`, `d_beta` match the paper's Eqs. 73/75/77 directly; recorded in
`references.md`, 08b9c5a8). Missing the cancellation understates the width by `(n|s|)^(1/2)`,
divergently toward the separatrix.

Both truncation points on the DIII-D SLAYER deck, n=1, in-grid plus edge-q-law surfaces:

| width | first overlap | cut ψ | ε_c | last kept rational |
|---|---|---|---|---|
| Riccati two-fluid (current scan) | m=6 / m=7 | 0.98919 | 1.1e−2 | q = 5 |
| Eq. (100) | m=7 / m=8 | 0.99711 | 2.9e−3 | q = 6 |

The paper's references: ε_c = 1.5e−3 (JET n=1, fig 9) and 4.8e−3 (n=4, fig 10); our 2.9e−3 sits in
that band where 1.1e−2 does not. Also worth noting the derived cut ratifies the existing empirical
guideline on this deck — `dmlim = 0.2` gives `qlim ≈ 6.2`, i.e. the same keep-q≤6 truncation — which
matches the paper's framing: it derives the ε_c that GPEC currently sets by guideline, and
adds the machine/profile/n dependence. Since Eq. (100) is algebraic in quantities
`build_slayer_inputs` already computes, the scan needs no Riccati solves at all (that width stays
available as SlabLayer diagnostic output; δ_visco is already out of the criterion per your comment,
0c2b3ac0).

On where the cut should land: the paper's prescription (p. 22) is to solve the response equations
(its Eqs. 47/48) in `0 < Ψ < 1−ε_c` and write `b = ∇V, ∇²V = 0` beyond — the overlap region's
response is vacuum-like. Grad-Shafranov is never truncated; figs 8–10 evaluate equilibrium-derived
quantities across and beyond the separatrix, so the background extends as far as it is constructible
and only the response solve stops. In GPEC that surface is `psilim`, where the Euler-Lagrange
integration ends and the vacuum matching already imposes `∇²V = 0` — and GPEC's empirical guideline
the paper refers to is the `dmlim` mechanism in `sing_lim!`, which sets `psilim`.
Deriving `qlim`/`psilim` from the Eq. (100) overlap point is also a strictly smaller change than
what's on the branch: nothing re-forms, so the trial/back-off loop and the immutable-config collision
(finding 1) disappear. That leaves `EdgeQuality` answering "how far is the equilibrium representable"
(psihigh) and the overlap criterion answering "how far is the response model valid" (psilim).

Was there a reason the feature targeted `psihigh` originally that this misses? If not I'll prototype
the `psilim` version.

Your references.md review comments are addressed on the branch (4201f675): the entry is trimmed to
the one-liner convention, the description and the S-convention note moved to the workflow docs page,
and LayerOverlap.jl's header now cites the paper directly.

d-burg:
@logan-nc prototyped the psilim version. Three things changed since my last comment, one of
which invalidates the numbers in it.

The formula I quoted was right, my implementation of it wasn't. I reintroduced the
`(n|s|)^(-1/2)` on top of GPEC's `lu` — exactly the double-count I described two paragraphs
earlier — which understated every width by `sqrt(n|s|)`, about 6.6x at q=7, and reported no
overlap anywhere. Fixed; `delta_dr = r_s^(3/2)*P_perp^(1/4)/(lu^(1/2)*d_beta^(1/2))`, no shear
factor, as written.

Second, the radial label matters and we had it wrong. Eq. (100) is not covariant — the width
goes as `(q/|dq/dr|)^(1/2)`, so it picks up `J^(1/2)` under a change of radial variable and is
only valid in the coordinate it was derived in. That is the flux label of the paper's Eq. (30),
`dpsi_p/dr = B0*r*g/q`, i.e. the toroidal-flux radius, not the geometric minor radius SLAYER
was using. On this deck that is worth one rational surface:

| radial label | first overlap | cut psi | eps_c | last kept rational |
|---|---|---|---|---|
| midplane minor radius | m=8 / m=9 | 0.99752 | 2.5e-3 | q = 6 |
| Eq. (30) flux label | m=9 / m=10 | 0.99939 | 6.1e-4 | q = 7 |

Worth saying plainly that the flux label is not a circular-plasma construct — it is defined
from flux alone, with no geometry input. On the circular LAR deck it reduces to the geometric
minor radius (r/a = 0.99 to 1.05); on the shaped DIII-D deck it departs from it by 20 to 45%.
It is the geometric radii that fail to generalise. Flux-surface-averaged radius does not fix
this: it is still geometric, so its Jacobian collapses at the separatrix, and in the edge band
it could not score the extrapolated surfaces at all (it is well-behaved at interior surfaces —
this is an edge-specific failure, not a general one).

Third, and the reason I would now trust the criterion: it reproduces the paper. Running the
Sect. 5.8 JET model directly — its Eqs. (32)-(36) q profile solved self-consistently for
rhat95/rhat105, its Eqs. (70)-(78) layer quantities, mtanh profiles off its fig. 8 — gives

| | this implementation | paper |
|---|---|---|
| n = 1 inner boundary | Psi = 0.9985 | Psi = 0.9985 |
| n = 4 inner boundary | Psi = 0.9936 | Psi = 0.9952 |

n = 1 lands on it; n = 4 is 34% high in eps_c, which I would put on the digitised pedestal. The
n-scaling also comes out right on our own deck: eps_c(n=4)/eps_c(n=1) = 2.85 against the paper's
3.20, both sublinear, as `u_c ~ n*|ln u|` predicts. Note our own eps_c = 6.1e-4 sits below the
paper's 1.5e-3 — that is a different machine, and the JET reproduction above is the controlled
comparison.

So, to your original worry: the 0.98 truncations were not real. They came from the ODE width
plus the geometric label. With Eq. (100) in its own coordinate the criterion changes no retained
surface on any shipped DIII-D deck — it is beyond psihigh for the ideal (0.995), riccati (0.995)
and gal_resistive (0.993) decks, and on the SLAYER deck (0.9995) it binds by 1.1e-4 in psi
without crossing a rational. Two of those decks carry no kinetic profiles, so the scan cannot
run on them regardless. The feature is a no-op on everything we ship, which argues for shipping
it opt-in with the recorded overlap point as the actual deliverable.

Architecture, per your sign-off: the overlap point is an upper bound on psilim, nothing
more. It is applied as a cap on `qlim` at the top of `sing_lim!`, so `dmlim`/`qhigh` still select
the final surface from inside it, and a cap beyond psihigh loses the `min` and is inert with no
special case. psihigh is untouched and stays user-set. Consequences: no equilibrium re-formation,
so the trial/back-off loop, `EdgeQuality`, the `refined_psi_grid` retarget and the
immutable-config collision all disappear; roughly 1800 insertions become about 700.

Still open, and I would not gate this PR on it: the two width channels now disagree by about two
surfaces (|delta_s| cuts at 0.98978, Eq. 100 at 0.99939). Both are reported separately. That is
the subject of the ODE-vs-Eq.(100) reduction check.

The radial-label plumbing landed separately as `radial_label` in #403; this branch stacks on it
rather than duplicating it.

Regression harness: not run yet on this branch — I am re-running it. I will post the table separately. Expected
movement is on the limited decks only (Solovev, LAR): the shared `edge_q_law` fit now gates the
grid-refinement edge density floor, which was previously applied unconditionally, so a plasma
with finite edge q is no longer packed as if q diverged. Diverted decks should be untouched.

logan-nc:
@d-burg please hold for me to find time to skim the full diff. Quick thoughts from the comments:

> Architecture, per your sign-off: the overlap point is an upper bound on psilim, nothing
more. It is applied as a cap on qlim at the top of sing_lim!, ... roughly 1800 insertions become about 700.

Great. I like this simplification a lot. Open physics question: Do we want to always cut inside this stochastic boundary (cut off some valid region) or always get *at least to* this boundary (include a little invalid region in the name of making sure we capture all the valid region)? The former is numerically easier (we don't have to go out as far), but perhaps the later in more physical (debatable - lets debate it here a bit)?

> Still open, and I would not gate this PR on it: the two width channels now disagree by about two
surfaces (|delta_s| cuts at 0.98978, Eq. 100 at 0.99939). 

Please explicitly confirm non of the r-vs-rho and other normalization bugs need to be extended to delta_s. Are we tripple sure this one is being implemented correctly for getting a Jacobian independent psi width (note that this is easily testable by running with Hamada and PEST)?

> The feature is a no-op on everything we ship. 

Do we want this to be true? If the feature is the future and how we want users to default use the code, then shouldn't we be shipping examples that use it (i.e. psihigh 0.9995 for all DIII-D examples)? The DIII-D like example inherited the dmlim truncation behavior only to make it easy to compare to fortran, but we've moved past that I think (and the bechmark_with_fortran script can easily account for it by defaulting to a dmlim type comparison). @jhalpern30 and @matt-pharr may want to add their two cents.

logan-nc:
@d-burg and I would like to request @jhalpern30's review. Primarily focus on the top level workflow, which now does not touch the psihigh for equilibrium reconstructions but does sets a psilim that the standard truncation logics (dmlim, max dW, etc.) cannot exceed. Also, please weigh in on this being a no-op on the shipped examples (i.e. should we change psihigh to 0.995 for the DIII-D examples).

Of course, calling out AI nonsense is much appreciated too. 

@d-burg please do post replies to my comments above. You can assume my review becomes an approval if @jhalpern30 or another human looks through the diff and approves.

d-burg:
@logan-nc all three taken, and you were right that the entry never landed — the branch rewrite dropped `references.md` from the diff entirely, so your comments ended up orphaned on a file this PR no longer touched. Fixed now:

- The paper is in `docs/resources/` (arXiv:2511.07666v2, the open-access version of doi 10.1088/1741-4326/ae4fdd).
- `references.md` gets a one-liner matching the surrounding entries, not the essay. There was no InnerLayer/SLAYER section at all — not even the Burgess 2026 paper the module is built on — so I added a short one covering both.
- "which is exactly" is gone, along with the claim about other response codes; I could not support it and it should not have been asserted.

In-source annotations already cite it (`LayerOverlap.jl` header, the `delta_dr` docstring in `LayerThickness.jl`), which was your third point.

One correction while I was in the paper: the figures are **Fig. 9 for n = 1 and Fig. 10 for n = 4**, not 10 and 11. The abstract states ε_c = 1.5e−3 and 4.8e−3 directly, and `1 − ε_c` gives Ψ = 0.9985 and 0.9952 — exactly the two values the test asserts.

d-burg:
## The convention question, settled with data

@logan-nc you asked whether to cut inside the overlap boundary or reach it. I ran it across ten equilibria — the shipped DIII-D-like deck plus nine IDA DIII-D reconstructions — and the answer is neither of the half-width options. All values in ψ, n = 1, Eq. (100) width:

| case | cut_inside | **at_surface** | reach | next_surf |
|---|---|---|---|---|
| shipped D3D-like | 0.997414 | **0.998508** | 0.999602 | 0.999684 |
| 153072_3415 | 0.996756 | **0.998617** | 1.000477 | 0.999633 |
| 153833_3450 | 0.995828 | **0.997955** | 1.000082 | 0.999752 |
| 153841_3000 | 0.994295 | **0.996575** | 0.998854 | 0.999559 |
| 169510_3000 | 0.998939 | **0.999547** | 1.000155 | 0.999933 |
| 201586_4200 | 0.990769 | **0.992546** | 0.994324 | 0.995331 |
| 204441_4400 | 0.996759 | **0.998276** | 0.999794 | 0.999634 |
| 204441_5307 | 0.995140 | **0.997262** | 0.999384 | 0.999825 |
| 179633_3200_L | 0.998810 | **0.999503** | 1.000196 | 0.999902 |
| 189392_4080_L | 0.996998 | **0.998379** | 0.999760 | 0.999774 |

`cut_inside` = ψ[k−1] − w/2 (what the code did), `at_surface` = ψ[k−1], `reach` = ψ[k−1] + w/2, `next_surf` = ψ[k].

**`reach` is not implementable.** It exceeds ψ = 1 on four of ten — a domain limit outside the plasma. On those same four it also exceeds `next_surf`: the half-width of the last clean surface extends past the very surface it is meant to exclude, because by then the layers are wider than their separation.

**`at_surface` is the answer**, and it is "reach the boundary" in the only well-defined sense. It is the paper's own `1 − ε_c`; it is what `runtests_layer_overlap_fitzpatrick.jl` already reports in order to hit the paper's 0.9985 / 0.9952; it is always inside the plasma and always inside the first contaminated surface. It also removes an inconsistency I should have caught earlier — the scan and its own external validation were using different conventions.

Changed in `d10d8b417`. Moves the limit outward by 1.2e−3 to 4.6e−3 in ψ, which is consistently exactly one rational surface.

d-burg:
@logan-nc on your δ_s question from 2026-08-21:

**Confirmed clean, by measurement rather than assertion.** I took your suggestion and ran it under three jacobians (`pest`, `hamada`, `equal_arc`). Everything is evaluated at fixed ψ_N, so the jacobian sets the θ-coordinate and nothing that matters should move. Max relative spread, default `:midplane` label:

| quantity | spread |
|---|---|
| `r_s` | ≤ 1.3e-16 |
| `da/dψ` (the metres→flux Jacobian) | ≤ 3.0e-14 |
| `\|δ_s\|` in metres | ≤ 7.4e-06 |
| `δ_s` as Δψ | ≤ 7.4e-06 |
| `δ_DR` (Eq. 100) as Δψ | ≤ 1.2e-06 |

The ~1e-6 floor on the widths is the equilibrium solve's own coordinate-dependent numerical noise propagating through the Riccati layer solve, not a coordinate artifact — the inputs feeding it are invariant to 1e-14. Same result under the `:flux` label. So `δ_s` and the metres→ψ conversion are Jacobian-independent, and none of the r-vs-ρ normalization issues need extending to it.

**The check did find something adjacent**, though it does not touch the truncation: `:fsa` and `:halfwidth` **are** coordinate-dependent. `:fsa` is a uniform mean over the coordinate's θ, which is a genuine flux-surface average only when the jacobian is already uniform in θ — so `r_s` moves 3.8–6.6 % and `da/dψ` up to 14.5 % on a `jac_type` change alone. Full table on #403, where that code lives. Neither label is the default and neither is used here.

Your third question — whether this should stay a no-op on the shipped decks — I'll answer separately; worth noting for @jhalpern30 / @matt-pharr that the decks are not uniform today: SLAYER and IMAS are already at `psihigh = 0.9995`, ideal and riccati at 0.995, gal_resistive at 0.993.

d-burg:
## Status: moved to draft; two bugs found against real data

`resistive_layer_overlap` has no call site, `sing_lim!`'s `psilim_cap` is never passed, `refined_psi_grid`'s `psihigh` has no caller, and there is no TOML key or HDF5 output. The body's claims — "upper bound on `psilim`", "the overlap point is recorded whether or not it binds", and the per-deck binding numbers — describe a prototype whose wiring was lost in the rewrite. The body will be rewritten to match.

Scanning ten real equilibria to settle the convention question turned up two things worth knowing.

### 1. The edge q-law gate rejected 8 of 9 real DIII-D reconstructions (`d4ce058ff`)

`EDGE_FIT_MIN_R2 = 0.99` was calibrated against the synthetic TokaMaker deck (r²_log = 0.9957, 0.9989 — smooth and analytic). Real EFIT/CAKE reconstructions carry noise in the edge q and fit the log law at r²_log = 0.929–0.977, so every one of them was refused. No surfaces were placed past the equilibrium grid, and the scan reported "no overlap" — a silent fail-open on exactly the case the criterion exists for.

| class | case | r²_log | r²_linear |
|---|---|---|---|
| diverted (real) | five DIII-D reconstructions | 0.929–0.977 | 0.713–0.773 |
| diverted (synthetic) | TokaMaker D3D-like | 0.9957, 0.9989 | — |
| limited | a10 fixed-boundary | 0.972 | — |
| limited | Solovev | 0.760 | 0.9996 |

**No absolute threshold can do this job**: the limited a10 deck sits at 0.972, inside the real diverted band. The old 0.99 only appeared to work because it was tuned against synthetic data that happens to be smooth. The `r²_log > r²_linear` comparison *does* separate the classes and is unchanged.

Why 0.90 rather than removing the absolute test: it is a floor against a fit that describes nothing at all, kept deliberately weak so the relative test does the discriminating. It clears the measured diverted band (0.929–0.977) with margin while still rejecting Solovev-class fits outright. I would have dropped the absolute test entirely except that only one shipped limited deck (Solovev) actually exercises the relative rejection — a10 rejects on knot count before reaching it — which is thin evidence for removing a safety check. If someone has a limited deck with a well-fitting log edge, that is the case that would change my mind.

Carries `!` because the same gate controls the auto-grid edge density floor: an equilibrium previously rejected now receives it and gets a different refined grid. Every shipped deck is unaffected (the DIII-D-like deck passed both thresholds; Solovev and LAR reject on the relative test; a10 on knot count), so no harness case moves — but a user running a real reconstruction on `grid_type = "auto"` will see a different grid.

### 2. Reverse shear crashed the layer build — split out as #431

`slayer_parameters` used the signed r-based shear in five expressions that are magnitudes, so a negative-shear surface produced a negative Lundquist number and `tauk = lu^(1/3)·tau_h` threw a `DomainError` before any layer quantity existed. Shot 153072_3415 has q dipping to 1.855 off-axis, giving two q = 2 surfaces with the inner one on the negative-shear branch — and `_find_rational_surfaces` is deliberately reverse-shear-safe, so it hands the layer build exactly those surfaces.

Pre-existing on develop, not introduced here, so it is its own PR. Cherry-picked onto this branch so the scan can run that shot; the copy resolves away once #431 merges.

Also `6781b639f`: a single failing surface no longer aborts the whole scan — the per-surface guard caught only `ArgumentError`, so one bad surface took down the run instead of being noted and skipped.

### Other changes since you last looked

- Rebased onto the rebased #403; `sing_lim!` moved to `src/ForceFreeStates/Surfaces/Finding.jl` in the #400 split.
- The `diiid_slayer_n1` γ `noise_threshold` bump (1e-1 → 2.5e-1) is **pulled out of this PR** so a widened acceptance criterion is reviewed as its own decision rather than riding along with a feature. It is a real effect — same-source fresh re-runs step by 0.076/0.122/0.145 Hz — but the underlying cause is the nondeterministic threaded root search (#418), and that is what should actually be fixed. #431's harness run is a clean illustration: a provably inert change trips the current bound.

jhalpern30:
Please ping me again when this is ready for review - it seems to still be being developed. Turning off auto-notifications on this for now to avoid my inbox being blown up

logan-nc:
Same @d-burg. It says 
> [d-burg](https://github.com/d-burg) marked this pull request as ready for review [15 hours ago](https://github.com/OpenFUSIONToolkit/GPEC/pull/298#event-30141184713)

above but I see no comment explaining how/why it got promoted from draft. The SLAYER stack is big rn so I am a bit lost - hold our hands through this please 🙏

d-burg:
I put together a review package for this PR — the overlap criterion and cut conventions as one diagram, the always-recorded / opt-in-applied wiring, the ten-equilibrium convention evidence, and the bugs found en route.

https://claude.ai/code/artifact/71b5b47a-7b76-45e8-9714-bd43041bcde4

logan-nc:
@d-burg lets merge this one way or another this week. Responses to the review package artifact questions:

> Does the convention choice survive contact with dmlim?

I am fine with this for now. The result is perhaps a very small fraction of "ideal-ish" plasma getting truncated out - it is still much better than what we had before and pushing further out into the known-stochastic region feels unjustified as well as numerically harder.


> The two width channels still disagree by about two surfaces. |δ_s| (Riccati, the dispersion-relation equivalent sheet thickness) cuts at 0.98978; Eq. (100) (the spatial decay scale) cuts at 0.99939. 

I believe @d-burg declared Eq. (100) to be the more consistent one to the paper and thus agree this is where we should start. Besides, 0.98978 is pretty deep and starts to cut into the pedestal... so more thought would be needed about what that means → future task to be considered if/when this has been used a bit more so we have more experience.

> Should the shipped decks stop being a no-op?

No. All DIII-D like examples should get `psihigh = 0.9995` like already done in the SLAYER and IMAS examples. It is an op when this is used.

> The edge_q_law gate change is exercised by diiid_n1 and diiid_slayer_n1 (both auto grids, both unchanged) but by no limited deck on an auto grid — every Solovev/LAR case uses grid_type = "ldp" with fixed mpsi. The limited-plasma branch is argued from measured r² values, not covered by a harness case.  Should a Solovev auto-grid case be added?

Nah. Too much of a change. Lets just add a super quick test that does solovev with auto-grid and confirms it is correctly classified as limited - no full example needed.

> Cost lands on runs that never asked for the feature.

This is fast and it's always valuable to have the basic outputs - so when users come to developers with "my run is funny" we can quickly look it up and see if the truncation they used was "reasonable". This is the number one gotcha for new users.

logan-nc (approved):
Approved once the task list associated with the response to the review package questions is addressed.

src/ForceFreeStates/Surfaces/Finding.jl
```diff
@@ -119,6 +120,20 @@ function sing_lim!(intr::ForceFreeStatesInternal, ctrl::ForceFreeStatesControl,
     intr.q1lim = profiles.q_deriv(profiles.xs[end]; hint=Ref(profiles.npts_minus_1))
     intr.psilim = equil.params.psihigh_resolved
 
+    # Resistive-layer overlap imposes an UPPER BOUND on the domain: past the first pair of
```
jhalpern30: This should probably just be in the docstring (which needs to be updated with the extra argument anyway)

src/GeneralizedPerturbedEquilibrium.jl
```diff
@@ -479,8 +482,50 @@ function run_local_stability(ctrl::ForceFreeStatesControl, equil::Equilibrium.Pl
...
+end
+
+# Run the resistive-layer overlap scan for the run's toroidal mode number, or return `nothing`
+# when it cannot be run (no kinetic file, multi-n, or the scan itself refuses). Reads the kinetic
+# file directly rather than reusing `kinetic_profiles`, which is a `KineticProfileSplines` built
+# for the NTV path and carries a different field set than the layer builders take.
+function _layer_overlap_scan(path::Union{Nothing,AbstractString},
+    intr::ForceFreeStatesInternal, equil::Equilibrium.PlasmaEquilibrium)
+    (path === nothing || !isfile(path)) && return nothing
+    if intr.nlow != intr.nhigh
```
jhalpern30: Why not just take the most conservative truncation limit? Like min(layer_overlap(n) for n in nlow:nhigh)?

src/GeneralizedPerturbedEquilibrium.jl
```diff
@@ -479,8 +482,50 @@ function run_local_stability(ctrl::ForceFreeStatesControl, equil::Equilibrium.Pl
...
+        npsi = length(data.psi)
+        omega = data.omega_E === nothing ? zeros(npsi) : collect(Float64, data.omega_E)
+        profiles = Utilities.KineticProfiles(; psi=collect(Float64, data.psi),
+            n_e=collect(Float64, data.n_e), T_e=collect(Float64, data.T_e),
+            T_i=collect(Float64, data.T_i), omega=omega,
+            omega_e=zeros(npsi), omega_i=zeros(npsi))
+        # Eq. (100) is not covariant -- it is anchored to the toroidal-flux label of the paper's
+        # Eq. (30), so the scan is driven in that label regardless of the SLAYER default.
+        return Tearing.resistive_layer_overlap(equil, profiles; n_tor=intr.nlow, rs_method=:flux)
+    catch err
```
jhalpern30: Is there an actual reason for the try catch here or is this an AI artifact?

src/ForceFreeStates/CoreTypes.jl
```diff
@@ -172,6 +173,7 @@ gpec.toml.
     numunorms_init::Int = 100
     singfac_min::Float64 = 1e-4   # Matches Fortran STRIDE; required nonzero for the Riccati path.
     set_psilim_via_dmlim::Bool = true   # Safe default for diverted equilibria (most production use); set false for limited/analytical (LAR, Solovev). Auto-skipped for multi-n. See docstring.
+    psilim_from_layer_overlap::Bool = false  # Cap psilim where adjacent resistive layers overlap. Opt-in; the scan is recorded either way. See docstring.
```
jhalpern30, line 176: Is false the right default here? If this is a true physical limit, do we want people to have to know that they need to turn this on or they might accidentally set truncation out too far?

src/GeneralizedPerturbedEquilibrium.jl
```diff
@@ -479,8 +482,50 @@ function run_local_stability(ctrl::ForceFreeStatesControl, equil::Equilibrium.Pl
...
+        cand !== nothing && isfile(cand) && return cand
+    end
+    return nothing
+end
+
+# Run the resistive-layer overlap scan for the run's toroidal mode number, or return `nothing`
+# when it cannot be run (no kinetic file, multi-n, or the scan itself refuses). Reads the kinetic
+# file directly rather than reusing `kinetic_profiles`, which is a `KineticProfileSplines` built
+# for the NTV path and carries a different field set than the layer builders take.
+function _layer_overlap_scan(path::Union{Nothing,AbstractString},
```
jhalpern30: If this is a property of the equilibrium, should this be done during setup_equilibrium? And along the same lines, would it make sense for the layer_overlap to be a member of the PlasmaEquilibrium struct? Completely up to you, just a thought. I guess its a balance of scope/struct bloat vs. carrying along layer_overlap everywhere

jhalpern30 (approved):
I gave a very cursory review and did not read the entire chain. These comments are all optional and I'll leave it up to you if you decide any of them are worth it, its just things I saw when skimming the code; sorry if anything was repeated from the above. Don't want to add too much work on a PR that's been sitting for a while

The implementation specifics (i.e. dmlim logic, opting in to get truncation, warning the user if they requested a psihigh past truncation, no-ops on most shipped decks) all seem ok with me

src/GeneralizedPerturbedEquilibrium.jl
logan-nc: Agreed

src/GeneralizedPerturbedEquilibrium.jl
logan-nc: skimming these comments - but I think the `resistive_layer_overlap` itself might throw errors? I agree this setup is opaque and could/should be cleaner

docs/development/references.md
d-burg: The `references.md` entry is now the same shape as its neighbours, with a one-line *Describes*. The explanation lives on the docs site instead: a new **Integration domain truncation** section in `stability.md` covers the `qhigh`/`psihigh`/`dmlim` rules, the overlap criterion and how the scan finds it, why the cap is off by default, the truncation example, and the `LayerOverlap` datasets. Fitzpatrick (2025) is added to `citations.md`, cross-linked both ways (0d73b1303).

docs/development/references.md
d-burg: Removed. Both that sentence and the claim about other response codes are gone; the entry now only states what the paper provides (the Eq. 100 width and the Sect. 5.9 truncation).

docs/development/references.md
d-burg: Done. Fitzpatrick (2025) Eq. (100) and Sect. 5.9 are cited where the width is computed (`InnerLayer/SLAYER/LayerThickness.jl`), at the scan and its cut rule (`Tearing/LayerOverlap.jl`), and now in the `sing_lim!` docstring (14d60b6b9).

src/ForceFreeStates/Surfaces/Finding.jl
d-burg: Moved into the docstring, which now documents `psilim_cap` and the `dmlim` bound; the inline comment is down to one line (14d60b6b9).

Doing this turned up a real bug. `dmlim` stepped back only against `qmax`, so it could land past the cap, and past `qhigh` too: `qhigh = 4.0` gave `qlim = 4.2`. And because the cap sits exactly on a rational, whether it was exceeded came down to how `q(psilim_cap)` rounded. `dmlim` now steps back until it is inside `min(qmax, qhigh, cap)`, so `qhigh = 4.0` and `4.1` both give `3.2`; the tests put a cap on the exact rational and 1000 ulps either side. No shipped deck moves.

src/GeneralizedPerturbedEquilibrium.jl
d-burg: Done in 8cfbbee7e. The scan now runs for each `n` in `nn_low:nn_high` and the innermost overlap point binds. On the DIII-D-like equilibrium at `psihigh = 0.9999`, n=1 overlaps at 0.99975 and n=2 at 0.99935, so `nn = 1:2` binds at n=2. The recorded `LayerOverlap` group is the binding scan, and its `n` dataset says which.

src/GeneralizedPerturbedEquilibrium.jl
d-burg: Fair question: it was hiding real errors (the empty-scan `MethodError` was downgraded to a "scan failed" warning). `resistive_layer_overlap` already records expected per-surface failures in its `notes` and rethrows anything else, so the outer catch-all is gone (8cfbbee7e). The only guard left is on reading the kinetic file, since that is external input to an optional diagnostic and a bad file should skip the scan rather than abort someone's stability run.

src/ForceFreeStates/CoreTypes.jl
d-burg: Keeping it `false`, now with numbers. On the shipped DIII-D-like decks the overlap point is at ψ ≈ 0.9997, beyond every deck's `psihigh`, so the cap cannot bind there and turning it on changes nothing (measured: identical results flag on vs off, even at `psihigh = 0.9995`).

On setting the truncation out too far: the scan always runs and is recorded under `ForceFreeStates/LayerOverlap/`, and GPEC now warns whenever `psilim` lands past the overlap point, flag on or off (14d60b6b9). `DIIID-like_truncation_example` (`psihigh = 0.9999`) is the case where the cap binds, with a harness case tracking it.

src/GeneralizedPerturbedEquilibrium.jl
d-burg: I'd keep it out of the equilibrium. The overlap point depends on `n` (now on the whole `nn` range) and on the kinetic profiles, neither of which `PlasmaEquilibrium` carries, so it belongs to the run rather than the equilibrium. It is computed once in `prepare_force_free_states!` and only goes to `sing_lim!` and the HDF5 writer, so it isn't carried around anywhere else.

logan-nc:
@d-burg I was excited to see the commits roll in! Whats the status now? Still working through any todos?

PR #306, opened 2026-06-25
VACUUM (3D) - taking advantage of field period symmetry for nonaxisymmetric vacuum calculations

jhalpern30:
## Summary
This PR prepares the Julia vacuum module for 3D stellarator use by adding an NFP-exploiting fast path for 3D vacuum response that preserves the full-torus answer at lower cost. It takes advantage of the block-circulant structure of the boundary integral operator matrix when there is an `nfp` = # of field periods symmetry of the boundary. This has computational advantages that scale like `nfp^3` for compute and `nfp` for storage without losing ANY information in the solve.

In other words - there is still toroidal mode decoupling in stellarator geometry, but instead of no toroidal modes being coupled, only modes with identical `k = n mod nfp` are coupled, and these modes are deemed a "mode family". The code can be solved for each of these mode families individually, similar to DCON. The structure I have implemented allows for both 1) all mode families to be solved in a single call (but solved independently to still take advantage of the speedup) and 2) for a single mode family, which is even faster.

The 3D VACUUM code can now accept inputs for a list of toroidal modes, not necessarily equally spaced, along with the boundary over only one field period to better interface with the 3D plasma response.

Note that unlike #190, there is no loss of accuracy when implementing this improvement since the toroidal modes are completely decoupled.

## Other required code changes

The only real numerical differences are in the 3D components of the vacuum code. However, cosmetic changes were necessary/came up elsewhere:

1. Making `VacuumInput` input arrays endpoint exclusive (helps for when I have to mirror the field periods to construct the full boundary)
2. Making `VacuumInput` accept a vector of modes versus the `mlow, mhigh` etc. And propagating this to the fourier coefficients
3. Having `compute_fourier_coefficients` return the complex basis, rather than the sin and cos terms. This led to a major rework throughout the rest of the code that cleaned up a lot. This probably should have been a separate PR, but I didn't know I would have to do this ahead of time and its integrated in what I was doing here already. I removed the benchmarking script which at this point is very stale and not worth keeping up to date
4. Separating the 2d and existing 3d axisymmetric code into a 2d and 3d code

## Still need to do (but not in this PR)
1. The block-circulant code can probably be further optimized for reducing memory
2. `grri/grre` matrix construction for perturbed equilibria are not integrated or tested in 3D yet
3. Potentially stellarator symmetry too?


jhalpern30:
## Benchmarks
### Numerical Improvement
I created a test where I initialize a boundary at some specified resolution and nfp and run both the nfp optimized code and with the same boundary but with nfp = 1 after I mirror the boundary such that it goes through the dense solve of the code. There is clearly a great speedup/memory reduction that increases with nfp, showing I am taking advantage of the symmetry. Note that these do have AdaptiveArrayPooling integrated but were tested with `MAYBE_POOLING[] = false`, a cool testing feature Min-Gu showed me that allows you to turn it off for selective functions to see what the memory usage looks like. 

```
small  (N=1152)  N=3456 (24×48/period × nfp=3)  M=1152  modes=15  residue-classes=3
------------------------------------------------------------------------------
  full (dense N×N)        time= 0.769 s  @alloc=  193 MiB
  reduced (block-circ.)   time= 0.306 s  @alloc=  109 MiB
  speedup: 2.51×   max rel err: 1.51e-15

medium (N=2304)  N=6912 (48×48/period × nfp=3)  M=2304  modes=42  residue-classes=3
------------------------------------------------------------------------------
  full (dense N×N)        time= 6.549 s  @alloc=  769 MiB
  reduced (block-circ.)   time= 2.246 s  @alloc=  453 MiB
  speedup: 2.92×   max rel err: 1.59e-15

large  (N=4608)  N=9216 (48×48/period × nfp=4)  M=2304  modes=42  residue-classes=4
------------------------------------------------------------------------------
  full (dense N×N)        time=14.378 s   @alloc= 1351 MiB
  reduced (block-circ.)   time= 3.068 s  @alloc=  521 MiB
  speedup: 4.69×  max rel err: 3.80e-15
```
### 3D GPEC Integration
For an integration test with 3D GPEC + n_stride capabilities, I ran the full workflow for Junyoung's sample NFP = 3 equilibrium with:
```
n_low = 1, n_low = 6, n_stride = 1 -> n = [1, 2, 3, 4, 5, 6] (previously working case)
[RESULT] dW_total  = 6.608533e-01
[RESULT] dW_plasma = 2.212641e-01
[RESULT] dW_vacuum = 4.395892e-01
``` 
and 
```
n_low = 1, n_low = 6, n_stride = 3 -> n = [1, 4]
[RESULT] dW_total  = 6.436734e+01
[RESULT] dW_plasma = 1.653465e+00
[RESULT] dW_vacuum = 6.271387e+01
```
and
```
n_low = 2, n_low = 5, n_stride = 3 -> n = [2, 5]
[RESULT] dW_total  = 6.608515e-01
[RESULT] dW_plasma = 2.212649e-01
[RESULT] dW_vacuum = 4.395866e-01
```
This result indicates that the most unstable mode lives in the n = [2, 5] decoupled family, and we can recapture it to within numerical error while only running with these modes. When running with the other n = [1, 4] family, we capture a more stable mode instead

jhalpern30:
@claude review

jhalpern30:
Regression results for solovev_n1, solovev_multi_n, diiid_n1
Summary: 21 unchanged
Summary: 15 unchanged
Summary: 46 unchanged

jhalpern30:
@jy021125 wanted to make sure you are tagged on this one - this should enable the n_stride = nfp implementation we discussed over email + some runtime improvements to speed and memory. I have pushed the corresponding changes needed in the python to the open PR on your github, so if you want to test this, it should work with the python on the `vacuum_interface` branch and the julia on this PR's branch.

Let me know if you find any bugs testing this out!

logan-nc:
@jhalpern30 is this waiting on me?

jhalpern30:
@logan-nc yes

jhalpern30:
@logan-nc nevermind, there were annoying merge conflicts from develop that I have to fix. Will ping you once its up to date

jhalpern30:
@logan-nc ok now this is good - was struggling with the issue described in #338.

Unit tests now pass and regression harness indicates agreement, as expected since the changes were primarily cosmetic here:
```
Regression Report: diiid_n1
Summary: 48 unchanged
Regression Report: solovev_n1
Summary: 21 unchanged
```

src/Vacuum/DataTypes.jl
```diff
@@ -216,13 +249,13 @@ function PlasmaGeometry(inputs::VacuumInput)
     @assert length(inputs.x) == length(inputs.z) == length(inputs.ν) "x, z, and ν must have the same length"
 
     # Interpolate arrays from input onto mtheta grid
-    θ_in = range(0.0, 2π; length=inputs.mtheta_in) # VacuumInput uses [0, 2π] grid
-    θ_out = range(; start=0, length=inputs.mtheta, step=2π/inputs.mtheta) # VACUUM uses [0, 2π) grid
+    θ_in = range(; start=0, length=inputs.mtheta_in, step=2π/inputs.mtheta_in)
```
logan-nc, line 252: Out of curiosity, why did you change the grids to be endpoint exclusive? Was it a bug? It looks like at least these local lines were self consistent (i.e. you had to change both the rang and the spline endpoint condition) but I remember bugs in these things being tricky to track down during the vacathon.

benchmarks/benchmark_fourier_transforms.jl
logan-nc, line 1: You didn't like this file? Honestly, it still seems like what does and doesn't belong in Benchmarks is not well enough defined. Someone should try to clarify it more in claude.md I guess... but that is mission creep that doesn't belong in this PR. I am ok letting the directory grow and get culled organically for now 😅

logan-nc (approved):
Looks like a great addition and some really nice cleanup of dirty real/imag coding from fortran ports.

src/Vacuum/DataTypes.jl
jhalpern30: The main motivation was because the rest of the VACUUM code uses endpoint exclusive grids and it makes it easier to mirror the equilibrium via field periodicity without duplicating endpoints. 

This wasn't fixing a bug, but rather making a small change that existed due to a quirk of interfacing the endpoint-inclusive DCON grids with the endpoint-exclusive VACUUM grids. I think we opted for inclusive during the hackathon since that was before Min-Gu added in the nice :endpoint=exclusive API for periodic BCs that make this relatively clean

benchmarks/benchmark_fourier_transforms.jl
jhalpern30: I think I looked, and I would have removed over half the file since the things its benchmarking were getting tossed, so it was just worth deleting outright.

I am also of the opinion that benchmarks is a weird folder - it seems like a conglomeration of people's scripts that were used for testing that no one else would ever run. I am fine with keeping them until they become so out of date that they're annoying to maintain

PR #224, opened 2026-04-14
KineticForces — Complete PE→KF pipeline and validate DIIID-PENTRC benchmark

logan-nc:
## Summary

Completes the PerturbedEquilibrium → KineticForces data pipeline so the Julia NTV calculation runs end-to-end from PE state through JBB deweighting, bounce averaging, pitch/energy integration, and the outer ψ ODE. Validated on the DIIID kinetic benchmark against Fortran PENTRC.

## DIIID benchmark results (vs Fortran PENTRC reference)

| Metric                | Julia       | Fortran     | Error  |
|-----------------------|-------------|-------------|--------|
| T_total_fgar (N·m)    | 1.782       | 1.780       | 0.09%  |
| dW_total_fgar         | 0.141       | 0.137       | 2.68%  |
| T_total_tgar (N·m)    | 0.964       | 0.949       | 1.55%  |
| dW_total_tgar         | 0.117       | 0.112       | 4.81%  |

All four metrics within <5% of Fortran in ~277 s runtime.

## Key changes

- `KineticForces/Compute.jl` — psi ODE bounds clipped to `intr.psilim` (prevents 10⁸ overshoot from extrapolating `xi_modes` beyond DCON limit); δW stored as real scalar in Re slot of `total_energy` per Logan 2013 Eq. 19; separate psi/inner tolerances threaded through.
- `KineticForces/KineticForcesStructs.jl` — `set_perturbation_data!` implements full JBB deweighting (S,T,X,Y,Z application in m-space → inverse DFT → divide by J·B² → forward DFT), matching `pentrc/inputs.f90:828-868`. Tolerances tightened to `atol_xlmda=1e-8, rtol_xlmda=1e-5, atol_psi=rtol_psi=1e-5`.
- `KineticForces/BounceAveraging.jl` — added `1/(ntheta-1)` Riemann-sum normalization to match Fortran `spline_int` over `linspace(0,1,ntheta)` for `wb, wd, bj_integral, wmu_ba, wen_ba`.
- `KineticForces/EnergyIntegration.jl` — refactored from ODE-based `energy_integrand!` to scalar `energy_integrand_scalar(x, p; imag_axis=false)` returning `ComplexF64` for cleaner QuadGK integration.
- `Equilibrium/KineticProfiles.jl` — fixed `_read_kinetic_table` column ordering (row-stack instead of column-major reshape).
[...]

PR #310, opened 2026-07-02
kinetic NTV δW inflated vs Fortran: χ₁ over-scaling + three bounce-averaging bugs

### LINKED ISSUE #4: Refactor EQUIL
I believe it's crucial to establish a consistent and well-defined architecture for each major component. To that end, I am proposing the following structure for the `Equilibrium` module.

#### **Proposed Directory and File Structure**

The `Equilibrium` module will be organized within its own directory:

-   **(Folder) `Equilibrium/`**
    -   `Equilibrium.jl`
    -   `types.jl`
    -   `equilibrium_io.jl`
    -   `direct.jl`
    -   `inverse.jl`

#### **Responsibilities of Each Component**

1.  **`Equilibrium.jl`**:
    -   Acts as the main orchestrator and top-level module.
    -   Defines the primary public-facing API, `setup_equilibrium`, which will be the single entry point for external code to generate an equilibrium profile.
    -   It coordinates the flow between the I/O and solver components.

2.  **`types.jl`**:
    -   Serves as the single source of truth for all core data structures used within the `Equilibrium` module.
    -   This includes definitions for `EquilibriumInput`, `DirectRunInput`, `InverseRunInput`, and the final `PlasmaEquilibrium` object.

3.  **`equilibrium_io.jl`**:
    -   Encapsulates all I/O and data parsing logic (analogous to the legacy `read_equil.f`).
    -   It will contain a function (e.g., `EquilibriumIo.prepare_solver_input`) that takes an `EquilibriumInput` object and, based on `eq_type`, returns either a `DirectRunInput` or `InverseRunInput` object, ready for the appropriate solver.

4.  **`direct.jl` & `inverse.jl`**:
    -   These correspond to the direct.f and inverse.f files.
    -   They take a `DirectRunInput` or `InverseRunInput` object and are responsible for returning the final, fully-calculated `PlasmaEquilibrium` object.

**Example Implementation:**

```julia
# In Equilibrium.jl

# Internal helper function that uses multiple dispatch.
# Adding a new solver in the future only requires adding a new method here.
_run_solver(input::DirectRunInput) = Direct.run(input)
_run_solver(input::InverseRunInput) = Inverse.run(input)

# The public-facing API
function setup_equilibrium(equil_input::EquilibriumInput)
    # 1. Prepare solver-specific input data via the IO module
    solver_input = EquilibriumIo.prepare_solver_input(equil_input)

    # 2. Dispatch to the correct solver based on the type of `solver_input`
    final_equilibrium = _run_solver(solver_input)

    println("--- Equilibrium Setup Complete ---")
    return final_equilibrium
end
```

I welcome any feedback.

### LINKED ISSUE #269: KineticForces — dW_kinetic (imaginary NTV energy) ~17–19% off Fortran on DIIID benchmark
## Summary

On the DIIID PENTRC benchmark (PR #112, "KineticForces Module"), the **NTV torque** metrics clear the 5 % acceptance bar against Fortran GPEC, but the **imaginary energy-conversion-rate** metrics — `dW_kinetic` — do not. They drift ~17–19 % from Fortran. This was flagged during PR review and explicitly accepted as a **non-blocking follow-up**; this issue tracks resolving it.

## Evidence

Torque (headline result) — passing:

| Metric | Julia | Fortran | Error | Status |
|---|---|---|---|---|
| `T_total_fgar` (N·m) | 1.805189 | 1.780406 | 1.39 % | ✅ < 5 % |
| `T_total_tgar` (N·m) | 0.991272 | 0.948962 | 4.46 % | ✅ < 5 % |

Imaginary energy-conversion-rate `dW_kinetic` — out of band:

| Metric | Julia | Fortran | Error | Status |
|---|---|---|---|---|
| `dW_total_fgar` | 0.160012 | 0.136829 | **16.94 %** | ⚠️ |
| `dW_total_tgar` | 0.133295 | 0.112297 | **18.70 %** | ⚠️ |

This is a regression from PR #224's pre-merge values (`dW_fgar = 2.68 %`, `dW_tgar = 4.81 %`). An A/B test on merged HEAD with the pre-port energy integrator swapped back in gives identical numbers, so **the energy-integrator port itself is ruled out** as the cause.

## Leading hypothesis

Develop commit [`083b0cc0`](https://github.com/OpenFUSIONToolkit/GPEC/commit/083b0cc0) ("EQUIL - BUG FIX - Snap periodic theta endpoint at 2D interpolant construction sites") changed how 2D equilibrium interpolants evaluate near the periodic-θ boundary. Bounce-data computation reads those interpolants, and the **imaginary** part of the kinetic resonance integral is far more sensitive to that boundary than the real (torque) part — consistent with T passing while dW shifts by ~12–13 % from #224's pre-merge value.

## Proposed next steps

1. A/B the bounce-data interpolant reads with the pre-`083b0cc0` θ-endpoint handling to confirm or deny the hypothesis.
2. Inspect the imaginary part of the kinetic resonance integral near the periodic-θ boundary for sensitivity to endpoint treatment.
3. If confirmed, decide whether the fix belongs in the interpolant boundary handling or in how `compute_bounce_data` samples θ.

## Reproducer

```
BENCHMARK_TGAR=1 julia --project=benchmarks/.benchmarks_env benchmarks/benchmark_diiid_kinetic.jl
```

## References

- PR #112 — validation comment, sections 2 (NTV torque benchmark) and 5 (dW disclosure).
- Branch [`feature/kinetic-validation-proofs`](https://github.com/OpenFUSIONToolkit/GPEC/tree/feature/kinetic-validation-proofs/benchmarks/kinetic_validation) — raw `(ψ, dT/dψ, T)` arrays under `benchmarks/kinetic_validation/kf_{fgar,tgar}_profiles.dat`.
- Suspected cause: commit `083b0cc0`.

jaesun57:
## Summary

Closes #269. The kinetic δW drift against Fortran GPEC (dW_fgar 16.94 %, dW_tgar 18.70 % in the
issue) is resolved: **all four DIIID PENTRC benchmark metrics are within the 5 % acceptance bar
at production tolerances** (converged dW_tgar sits at 5.49 %, the known t/p-boundary residual —
see the tolerance table below), and the kinetic-DCON eigenvalue path is verified unregressed at
0.4 %. The issue's leading
hypothesis (develop `083b0cc0`, periodic-θ endpoint snap) is **tested and ruled out**; the real
root cause was a set of independent bugs in the KineticForces bounce-averaging chain.

Four bugs fixed on this branch:

| # | Bug | Commit | Effect on DIII-D |
|---|---|---|---|
| 1 | `set_perturbation_data!` multiplied ξ^α by χ₁ although `clebsch_alpha` is already the physical ξ^α (= Fortran `gpec_xclebsch`, which stores `xmsout/chi1`, gpout.f:5643) | `c8b1f6af` | ξ^α over-scaled ×1.648 |
| 2 | `bo` was `equil.params.b0` = bt0, the edge-extrapolated vacuum field at rmean (equil_out.f:268). Fortran PENTRC uses the **axis** field `\|F(0)\|/ro` (dcon_interface.f:1414). Since λ = μ·bo/E, the trapped/passing boundary and every bounce average were shifted | `0858f63b` | bo off 2.2 %, wb(λ) off up to ~3.5 % |
| 3 | Bounce θ-integrals were trapezoid sums; Fortran fits `spline_fit(...,"extrap")` (C² cubic, 4-pt Lagrange endpoints) and integrates exactly via `spline_int`. With the 1/√v_par integrable endpoint singularities this is a leading-order scheme difference | `8c0721f1` | ~3.5 % systematic on trapped ∫J·B/√v_par |
| 4 | v_par and the bounce points were computed from the **periodic** tspl; Fortran builds a separate `vspl` fit **"extrap"** and takes both `vpar=vspl%f(1)` and `spline_roots(vspl)` from it (torque.F90:599-602, 677 — "more consistent w/ bnce pts than direct from tspl"). Also: transit start t1/t2 must be the nodal knot `xs[ibmax]` (torque.F90:610-611, 648-649), not the refined dB/dθ=0 extremum; and roots must be complete + descending like `spline_roots` | (this PR, uncommitted) | off-node v_par near θ=0/1 seam and turning points |

Fix #4 exploits the identity `extrap-cubic(1−(λ/bo)B) = 1−(λ/bo)·extrap-cubic(B)` (extrap endpoint
derivatives are linear in the nodal data), so a single λ-independent `B_extrap` per surface
replaces Fortran's per-λ `vspl` exactly. Root-finding is `Roots.find_zeros` on the same cubic
(replaces a fixed 256-point sign-change scan that could miss near-marginal root pairs), sorted
descending to match `spline_roots` (spline.f:1782-1785).

## Validation — three-way (Fortran / Julia-before / Julia-after)

Fresh Fortran reference generated with the current binaries (v1.5.5-512) on
`docs/examples/DIIID_ideal_example` (dcon → gpec → pentrc) and `DIIID_kinetic_example`
(kin_flag=t dcon). "Before" = branch baseline `c8b1f6af` (χ₁ fix in, bounce code original);
"after" = this PR. Both Julia runs consume the same Fortran `gpec_xclebsch_n1.out` ξ, so the
comparison isolates the KF operator.

### NTV torque and δW (DIIID, n=1, nl=4, fgar+tgar)

| Metric | Fortran | Julia before | err | Julia after | err | Issue #269 |
|---|---|---|---|---|---|---|
| T_total_fgar (N·m) | 1.776725 | 1.729446 | 2.66 % | 1.753980 | **1.28 %** ✅ | 1.39 % |
| dW_total_fgar | 0.136937 | 0.127035 | 7.23 % | 0.132776 | **3.04 %** ✅ | **16.94 %** ⚠️ |
| T_total_tgar (N·m) | 0.944563 | 0.928392 | 1.71 % | 0.928948 | **1.65 %** ✅ | 4.46 % |
| dW_total_tgar | 0.110929 | 0.105831 | 4.60 % | 0.105898 | **4.54 %** ✅ | **18.70 %** ⚠️ |

The after-fix dW matches PR #224's pre-merge quality (dW_fgar 2.68 %, dW_tgar 4.81 %). Underlying
bounce quantities: wb/wd match Fortran to ~1e-4 across λ at ψ = 0.3/0.5/0.7; normalized dJdJ < 1 %
away from the trapped/passing boundary. Remaining tgar ~4.5 % is the t/p-boundary dip node +
λ-integration scheme (LSODE vs QuadGK), where both codes are marginal.

### Kinetic-DCON eigenvalue (kin_flag=t path, kinetic_source="calculated")

| Quantity | Fortran | Julia before | err | Julia after | err |
|---|---|---|---|---|---|
| Re(et₁) | 1.050683 | 1.046407 | 0.41 % | 1.046386 | **0.41 %** ✅ |
| Im(et₁) | −0.301242 | −0.302353 | 0.37 % | −0.302321 | **0.36 %** ✅ |

The bounce fixes do **not** regress the self-consistent eigenvalue path (before ≈ after to ~3e-5):
et is a ψ-integrated quantity and is insensitive to the local quadrature details that dominate the
NTV δW. Both clear the 5 % (Re) / 20 % (Im) acceptance with large margin.

### ξ^α (χ₁ fix) — corrected attribution

**The benchmark above cannot test bug #1.** Commit `c8b1f6af` removed the `./chi1` in the benchmark
*and* the compensating `.*chi1` in `set_perturbation_data!`; those cancel exactly, so the operator
receives the same ξ^α before and after. The benchmark feeds Fortran's ξ directly and bypasses the PE
pipeline. **The dW improvement in the table above therefore comes from bugs #2–#4, not bug #1.**

Bug #1 is real and is validated against the Fortran *consumer*: `read_peq` passes the
`gpec_xclebsch` column straight into `xs_m(3)` with no χ₁ factor (`pentrc/inputs.f90:807`), and
`set_peq` documents it as "the contravariant clebsch displacement xi^alpha … NOTE this is NOT DCON
CONVENTION -> xmsmns/chi1 in DCON" (`:758-759`); the `force_xialpha` fallback rebuilds the same array
with an explicit `(1.0/chi1)`. So PENTRC's `xs_m(3) = xms/chi1 = clebsch_alpha`, and develop's extra
`×χ₁` over-scaled the tangential displacement. Its effect is visible only through the Julia PE
pipeline, where it is a factor of 148 on the NTV torque (see the regression section below).

**Coverage gap this leaves:** the only case exercising bug #1 (`diiid_n1`) has no Fortran ground
truth, and the case that has ground truth (the benchmark) cannot exercise it. Closing that needs a
PENTRC reference run driven by Julia's own PE ξ — added as follow-up 5.

Before the fix the tangential displacement fed to the kinetic operator was `ξ^α · χ₁`
(χ₁ = 2π·ψ₀ = 1.648 on this equilibrium); after the fix it is identical to the Fortran
`gpec_xclebsch` column. The old NTV benchmark masked this with a compensating `./chi1`
(removed in `c8b1f6af`) — which is also why the bug never showed in CI (coverage gap, see below).

### Figures (attached)

Three-way overlays (Fortran black dashed / before orange / after crimson):

- `ntv_fgar_profiles.png`, `ntv_tgar_profiles.png` — dT/dψ, cumulative T, dδW/dψ, cumulative δW
- `kinetic_dcon_et.png` — Re/Im(et₁)
- `xi_alpha_chi1.png` — Σ_m |ξ^α|(ψ): before = ×1.648 over-scale, after ≡ Fortran

kinetic_dcon_et
ntv_fgar_profiles
ntv_tgar_profiles
ntv_totals
xi_alpha_chi1

Reproducer scripts (three-way dump drivers + plotter) are kept out of this PR to keep the diff
src-only; available on the local validation setup (or on request).

## The 083b0cc0 hypothesis — tested and ruled out

The issue's leading hypothesis blamed develop `083b0cc0` ("Snap periodic theta endpoint at 2D
interpolant construction sites") for the dW drift. Three independent findings rule it out:

1. **The snap is a 3-ulp data correction.** Reverting it makes the code fail to run at all —
   the strict FastInterpolations `PeriodicBC` validator rejects the unsnapped array with
   `y[1]=7.188034825477611` vs `y[end]=7.188034825477608`: a relative drift of 4e-16.
2. **Measured propagation is machine epsilon.** Building the periodic cubic with the endpoint
   snapped each way (the full ambiguity the commit resolves) changes the interpolant by
   **3.1e-15 relative** (dense-grid max). A 12–13 % dW shift would require ~4×10¹³ amplification.
3. **The recovery is bounce-side.** The fixes above restore dW to #224 pre-merge quality while
   `083b0cc0` remains in place, untouched.

The issue's intuition that Im is more boundary-sensitive than Re was correct — but the sensitive
boundary is the **bounce-integrand quadrature** (1/√v_par endpoints), not the equilibrium
interpolant θ-endpoint.

## Relation to PR #224's open items

- **"Matrix Path Gap"** (5–50 % rel_frob on A_k…H_k vs Fortran, smooth ψ-dependent scaling
  ‖F‖/‖J‖ ≈ 1.01–1.35, no m-structure, not quadrature-order): the bo scale error (2.2 % on λ),
  the trapezoid θ-quadrature systematic, and nlmda 64→128 are exactly this signature. The 0.4 %
  et₁ agreement after the fixes suggests the plateau is largely closed; re-running
  `compare_kinetic_matrices.jl` (validation-proofs branch) is a cheap follow-up confirmation.
- **Tolerance sensitivity**: #224 documented dW error exploding as the outer-ψ tolerance loosens
  past 1e-5 (2.7 → 12.8 → 53.8 → 73.7 %). Current `KineticForcesControl` defaults are
  `atol_psi=rtol_psi=1e-2`. Convergence check of the after-fix numbers (ψ-steps 435 → 3165):

  | Metric | default 1e-2 | tight 1e-5 | verdict |
  |---|---|---|---|
  | T_total_fgar | 1.28 % | **0.45 %** | T was outer-ODE-limited; converges toward Fortran |
  | dW_total_fgar | 3.04 % | 3.37 % | converged — residual is scheme, not tolerance |
  | T_total_tgar | 1.65 % | **0.32 %** | converges toward Fortran |
  | dW_total_tgar | 4.54 % | 5.49 % | converged; honest residual ~5.5 % (t/p-boundary follow-up) |

  **No tolerance explosion occurs with the fixed code** — dW moves < 1 pp between 1e-2 and 1e-5,
  so the #224 blow-up mechanism is gone (it was fed by the bounce-side errors). The converged
  dW_tgar sits marginally above the 5 % bar (5.49 %), which is the known t/p-boundary residual
  tracked in follow-up 3, not an outer-quadrature artifact.

## Follow-ups (not blocking)

1. ~~Add a regression case tracking `kinetic_forces/fgar/{total_torque,total_energy}` through the
   post-PE path — the coverage gap that let #269 ship.~~ **Done in this PR** (`bf13143d`):
   `total_energy` added to `diiid_n1` (torque was already tracked).
2. Re-run the #224 kinetic-matrix comparison to confirm the rel_frob plateau closed.
3. tgar residual (~4.5 %): t/p-boundary node treatment / λ-grid matching.
4. Revisit `atol_psi/rtol_psi` defaults vs #224's 1e-5 floor.
5. Add a Fortran-referenced check that exercises the χ₁ convention through the PE pipeline (a PENTRC
   reference run driven by Julia's own PE ξ). Today no case has both Fortran ground truth and the
   PE-pipeline ξ, so bug #1's fix rests on the Fortran source convention plus the physicality of the
   corrected magnitude.

---

## Post-review update (develop merged, review addressed)

**Reconciled with #322.** develop — including the merged #322 `BounceScratch` preallocation, which
rewrote the same functions — is merged in (`77c514d5`). The resolution keeps this branch's physics
and #322's allocation discipline: all per-λ quadrature buffers live in `BounceScratch`, and the
per-λ quadrature block measures **0 bytes allocated**.

**Custom spline helpers removed** (`1e462651`). Per review, verified `cubic_interp(…; bc=CubicFit())`
+ `FastInterpolations.integrate`/`cumulative_integrate!` matches the hand-written helpers to ~2e-16,
then benchmarked both on the DIII-D NTV case (fgar, nl=4, 8 threads, 3 warmed runs):

| backend | mean | min | T_fgar | dW_fgar |
|---|---|---|---|---|
| custom helpers | 92.4 s | 89.0 s | 1.768519035 | 0.136126468 |
| FastInterpolations | 89.2 s | 86.9 s | 1.768519035 | 0.136126468 |

Identical results, slightly faster → ~90 lines deleted. `_refit!` calls the package-internal
`FastInterpolations._solve_system!` (no exported API refits without allocating); documented in code.

**Comment pass** (`3ea624b9`) — one-line comments, single Fortran anchor per function, no issue
references or error percentages, `B_extrap` documented. Comments and docstrings only; no code changed.

## Regression harness (local vs develop `50440732`)

`diiid_n1` — **46 unchanged, 2 changed**, both intended:

| Quantity | Diff | Status |
|---|---|---|
| NTV torque FGAR (Re,Im) [N·m] | 8.237e+01 (99.33 %) | CHANGED |
| NTV kinetic energy dW FGAR [J] | 7.468e+00 (100.93 %) | CHANGED |

Both moves come **entirely from bug #1 (the χ₁ line)**, isolated by re-running this case at HEAD with
only `clebsch_alpha .* chi1` restored: `T = 84.02 − 16.19i`, `dW = −8.095` — i.e. back to develop's
`82.93 − 14.80i` / `−7.399`. The three bounce-averaging fixes account for the ~1 % residual.

The 148× factor (not the naive 1.648² ≈ 2.7) is because ξ^α is rescaled *alone* — ξ^ψ is untouched —
which shifts the `divx`/`dbob` balance through a near-cancellation rather than scaling the
perturbation coherently. The corrected `T ≈ 0.56 N·m` is also far more physical for this case than
`83 N·m`. Note this case is the synthetic TkMkr equilibrium with coil forcing and has **no Fortran
reference**; it is not the benchmark case below.

Eigenvalues, Δ′, island widths, Chirikov, resonant field, PE energies, and the ballooning/Mercier
profiles are bit-identical or at roundoff.

`solovev_n1` — 21 unchanged (ideal path untouched).
`solovev_kinetic_calculated` / `solovev_kinetic_nuzero` — eigenvalues shift ~1e-5 absolute
(0.00 % relative), ODE steps move ≤14 of ~1800: roundoff amplified by the adaptive outer ψ-ODE.

**The refactor itself is numerically inert.** Tracking `ntv_energy_fgar` across the branch:

```
50440732 (develop)   --                   --
77c514d5 (physics)   7.468e+00 (100.93%)  ** CHANGED **
local    (refactor)  8.3e-17              OK
```

All 255 `runtests_kinetic.jl` tests pass.

## Independent physics audit

A `fortran-physics-reviewer` pass re-derived all four fixes against `torque.F90`, `spline.f`,
`dcon_interface.f`, and `gpout.f` independently of this description — **PASS on all nine claims**,
with the spline-quadrature algebra and the affine-fit identity checked by hand. Notable
confirmations: `gpout.f:5637` is literally `xmsout = xmsout/chi1` ("remove DCON normalization"),
proving the ×χ₁ was a double-application; `nlmda = 128` is the Fortran default (`torque.F90:53`);
and dropping `SINGULAR_EPS` from `(pl + (1−σ)/pl)` is safe and more faithful, since `pl` is `cis` of
a real so `|pl| = 1` exactly.


src/KineticForces/BounceAveraging.jl
```diff
@@ -141,15 +141,103 @@ function _powspace_antideriv(x::Vector{Float64}, pow::Int)
...
+# The Fortran bounce integrals are NOT Riemann/trapezoid sums: torque.F90 fits
+# the θ-samples with equil/spline.f `spline_fit(...,"extrap")` (C² cubic with
+# endpoint derivatives from a 4-point Lagrange fit) and integrates the fitted
+# cubic exactly via `spline_int`:  ∫ over interval = h/12·(6(f_i+f_{i+1}) +
+# h·(d_i−d_{i+1})).  The 1/√v_par integrand near bounce points makes the
+# quadrature scheme a leading-order effect (trapezoid gave a systematic ~3.5%
+# inflation of ∫J·B/√v_par in the trapped region → wb 3.5% low, dW off; see
+# issue #269), so we reproduce the Fortran scheme exactly.
+
+"""
```
jhalpern30: Are these really needed? We migrated from the Fortran spline package. If possible, use FastInterpolations API whenever available.

src/KineticForces/BounceAveraging.jl
```diff
@@ -141,15 +141,103 @@ function _powspace_antideriv(x::Vector{Float64}, pow::Int)
...
 
 
+# ============================================================================
+# Fortran spline-quadrature helpers (uniform grid)
+# ============================================================================
+# The Fortran bounce integrals are NOT Riemann/trapezoid sums: torque.F90 fits
+# the θ-samples with equil/spline.f `spline_fit(...,"extrap")` (C² cubic with
+# endpoint derivatives from a 4-point Lagrange fit) and integrates the fitted
+# cubic exactly via `spline_int`:  ∫ over interval = h/12·(6(f_i+f_{i+1}) +
+# h·(d_i−d_{i+1})).  The 1/√v_par integrand near bounce points makes the
```
jhalpern30: Avoid this many references to Fortran - this is a new code. Ok to provide enough for an interested party to reference the original fortran code, but summarizing like this pollute this codebase. This comment could be a line or two if we really need the spline wrappers below

src/KineticForces/BounceAveraging.jl
```diff
@@ -141,15 +141,103 @@ function _powspace_antideriv(x::Vector{Float64}, pow::Int)
...
+# quadrature scheme a leading-order effect (trapezoid gave a systematic ~3.5%
+# inflation of ∫J·B/√v_par in the trapped region → wb 3.5% low, dW off; see
```
jhalpern30: This whole part of the comment can be cut off. Specific error values will quickly go out of date as well as references to github issues

src/KineticForces/BounceAveraging.jl
```diff
@@ -141,15 +141,103 @@ function _powspace_antideriv(x::Vector{Float64}, pow::Int)
...
+
+
 # ============================================================================
 # Core bounce averaging
 # ============================================================================
 
 """
     compute_bounce_data(psi, n, l, q, bo, bmax, bmin, ibmax, theta_bmax,
-                        tspl, mfac, chi1, ro, dbob_m_f, divx_m_f,
+                        tspl, B_extrap, mfac, chi1, ro, dbob_m_f, divx_m_f,
```
jhalpern30, line 150: B_extrap needs to be described in the docstring

src/KineticForces/BounceAveraging.jl
```diff
@@ -301,32 +389,42 @@ function _build_lambda_grid(method_char::Char, lmdatpb::Float64, lmdamax::Float6
 end
 
 
+"""
+    _vpar_from_extrap(B_extrap, lmda, bo, θ) → v_par
+
+Parallel-velocity factor `v_par = 1 − (λ/bo)·B(θ)` evaluated from the **extrap**
```
jhalpern30: Way too long of a docstring for a short inline function - I think this is more suitable for where B_extrap is created/in the main docstring of tpsi!/something like that. If its this essential of a detail, shouldn't be hidden in the docstring of an inline

src/KineticForces/BounceAveraging.jl
```diff
@@ -301,32 +389,42 @@ function _build_lambda_grid(method_char::Char, lmdatpb::Float64, lmdamax::Float6
...
-        # Find zero crossings
-        bpts = Float64[]
-        for i in 1:nfine
-            if vpar_fine[i] * vpar_fine[i+1] < 0
-                # Bisect for better accuracy
-                θ_root = _bisect_vpar(tspl, lmda, bo, theta_fine[i], theta_fine[i+1])
-                push!(bpts, θ_root)
-            end
-        end
+        # Bounce points = roots of v_par(θ) = 1 − (λ/bo)·B_extrap(θ) in (0,1).
```
jhalpern30: Explanation not needed. Something like "# Build v_par(θ) = 1 - (λ/bo)*B(θ) and find roots using extrapolated B field" would be sufficient

src/KineticForces/BounceAveraging.jl
```diff
@@ -490,31 +541,39 @@ function _bounce_integrate(
...
         djdpsi = tspl_f[5]
 
-        vpar = 1.0 - (lmda / bo) * B_val
+        # vpar from the extrap cubic of B (Fortran vspl, torque.F90:677), NOT the
+        # periodic tspl B_val — which stays as the numerator field below (Fortran
+        # bspl uses tspl%f(1)/(4) for the numerator, vspl%f(1) only for 1/√vpar).
+        vpar = 1.0 - (lmda / bo) * B_extrap(θmod)
 
         if vpar <= 0
-            # Zero crossing near bounce points — handle like Fortran lines 678-697
```
jhalpern30, line 564: The comments for this entire section could probably have stayed the same

jhalpern30 (commented):
Thanks for the fix! The physics changes look ok to me, but I'll let @logan-nc give the signoff on that since the kinetics are out of my area of expertise.

My main comments are ironically relating to comments - LLMs have a habit of absolutely going crazy with multi-line comments that pollute the codebase, and this PR introduces a lot of those that need cleaning up. I've started marking up a few of them, but most of the comments in this PR should be looked at with this lens and it can be cut down a lot. In general try to "humanize" the comments a bit more:

1. Avoid these long, multi-line comments around a single line that focus on what was changed versus what it does. LLMs have a habit of writing the logic behind what was changed, versus a simple statement describing the actual code that currently exists. The "what was changed" comments will quickly go out of date. Similarly, avoid adding a multi-line comment in place of a removed line of code
2. Avoid referencing Fortran and specific lines of Fortran explicitly, which will again go out of date. Focus on describing the physics of what was implemented, and if helpful just add a "performs the same calculation as XXX in the Fortran" if that would be helpful

Feel free to ping me again once you've done a pass through, happy to take a look at the whole PR then.

logan-nc:
Yes thank you! Great work.

I agree with @jhalpern30. 
@jaesun57 please streamline the annotations and tag me in a comment when ready for my review.

What AI are you using? the `CLAUDE.md` should have given you instructions to not do these things. For example:
```
4. **No Fortran references — annotations must be self-contained and Julia-centric.** The Julia
   code is the production workhorse; a new user must not need to know the legacy Fortran GPEC
   code to read an example. Headers explain the *scenario and its unique aspect*; inline comments
   explain the *variable*. Do **not** cite Fortran namelist files (`dcon.in`/`equil.in`/`pentrc.in`),
   Fortran flag names (`kin_flag`, `sas_flag`, `electron_flag`, …), or legacy code names
   (`DCON`/`STRIDE`/`PENTRC`) as things the reader must know. Non-Fortran external-model citations
   (e.g. the analytic TJ model behind `tj_analytic`) are acceptable provenance. (This applies to
   the user-facing `examples/*` and `test/test_data/*` TOMLs; `regression-harness/cases/*` may
   retain algorithm names as developer metadata.)
```

Feel free to add further instructions if needed based on what it did and what @jhalpern30 has called out as poor choices for human readability.

If using another AI, we should symlink `CLAUDE.md` to whatever file it uses.

logan-nc:
@claude review this pr

logan-nc:
FYI @jaesun57 I have claude working through the review requests above. I'll push updates soon

src/KineticForces/BounceAveraging.jl
logan-nc: Good call — they were not needed. First I verified that `cubic_interp(xs, f; bc=CubicFit())` + `FastInterpolations.integrate` / `cumulative_integrate` reproduces the custom helpers to ~2e-16 (real, complex, cumulative, and a 1/√-type endpoint integrand). Then I benchmarked both backends end-to-end on the DIII-D-like NTV example (fgar, nl=4, 8 threads, 3 warmed runs each):

| backend | mean | min | T_fgar | dW_fgar |
|---|---|---|---|---|
| custom helpers (with preallocated workspace) | 92.4 s | 89.0 s | 1.768519035 | 0.136126468 |
| FastInterpolations in-place refit | 89.2 s | 86.9 s | 1.768519035 | 0.136126468 |

Identical to every printed digit and slightly faster, so the helpers are deleted (`1e462651`). The bounce quadrature now holds one real and one complex `CubicFit` interpolant in `BounceScratch`; per λ their `y` data is overwritten and refit in place, then integrated with the full-domain `integrate` / `cumulative_integrate!` fast path.

One caveat, called out in the code: no *exported* API refits an interpolant without allocating, so `_refit!` calls the package-internal `FastInterpolations._solve_system!`. That is the single internal dependency and it keeps the hot loop allocation-free — the per-λ quadrature block measures 0 bytes. If FastInterpolations later exports an in-place refit we can drop it.

src/KineticForces/BounceAveraging.jl
logan-nc: Removed — that section no longer exists (the helpers it described are gone). The remaining quadrature comment is two lines explaining the physics reason (1/√v_par endpoint singularities make the quadrature scheme leading-order) with a single mention that this matches the original Fortran spline integration.

src/KineticForces/BounceAveraging.jl
logan-nc: Removed. No error percentages or issue references remain anywhere in the source — also a CLAUDE.md rule I should have followed.

src/KineticForces/BounceAveraging.jl
logan-nc: Added to the `compute_bounce_data` `# Arguments` list:

```
- `B_extrap`: Endpoint-fit (non-periodic) cubic of B(θ) used for v_par and the
  bounce-point roots (the Fortran `vspl` equivalent)
```

src/KineticForces/BounceAveraging.jl
logan-nc: Agreed — cut to three lines. The substantive part (why one per-surface `B_extrap` can replace the per-λ Fortran `vspl`: the endpoint fit of `1−(λ/bo)B` equals `1−(λ/bo)` times the fit of `B`) moved to where `B_extrap` is actually constructed in `tpsi!` / `_setup_surface_state`.

src/KineticForces/BounceAveraging.jl
logan-nc: Shortened to essentially that:

```julia
# Bounce points: all roots of v_par(θ) = 1 − (λ/bo)·B_extrap(θ) in (0,1),
# sorted descending — the same order as Fortran spline_roots, which the
# marginally-trapped and deepest-well wrap logic below assume.
```
Kept the ordering note because the code below genuinely depends on it.

src/KineticForces/BounceAveraging.jl
logan-nc: Restored to concise one-liners throughout that section. The only multi-line comments I kept are the two genuinely non-obvious parity notes: the `wd` slot being held from the `wb` integrand in the after-midpoint fill, and why the exact `== 0.0` test in the backfill is safe (the value was set by `fill!`).

logan-nc:
## Review pass complete — @jhalpern30 ready for another look

Four commits on top of the develop merge. Summary of what changed and the verification.

### 1. Reconciled with #322 (`77c514d5`, `fe346d44`, `1e462651`)

develop (including #322's `BounceScratch` preallocation) is merged in. The conflict was real rather than mechanical — #322 threaded a scratch struct through exactly the functions this PR rewrote. Resolution keeps this branch's physics and #322's allocation discipline: every per-λ quadrature buffer now lives in `BounceScratch`, and the per-λ quadrature block measures **0 bytes allocated**.

### 2. Custom spline helpers replaced by the FastInterpolations API (`1e462651`)

Per @jhalpern30's question on the `_fortran_spline_*` helpers — they were not needed. Equivalence verified to ~2e-16, then benchmarked end-to-end on the DIII-D-like NTV case (fgar, nl=4, 8 threads, 3 warmed runs):

| backend | mean | min | T_fgar | dW_fgar |
|---|---|---|---|---|
| custom helpers | 92.4 s | 89.0 s | 1.768519035 | 0.136126468 |
| FastInterpolations | 89.2 s | 86.9 s | 1.768519035 | 0.136126468 |

Same numbers, slightly faster → helpers deleted (~90 lines). `_refit!` uses the package-internal `FastInterpolations._solve_system!` because no exported API refits an interpolant without allocating; this is documented in its docstring and is the single internal dependency.

### 3. Comment pass (`3ea624b9`)

Per the review: one-line comments describing what the code does, one Fortran anchor per function instead of per-line citations, no issue references, no error percentages, `B_extrap` documented in the docstring. That commit's diff is comments and docstrings only — no code lines changed.

### 4. Regression coverage for the gap that let this ship (`bf13143d`)

`kinetic_forces/fgar/total_energy` is now tracked in `diiid_n1`. Torque was already tracked; **kinetic δW — the quantity this PR fixes — was not**.

## Regression harness (local vs develop @ `50440732`)

**`diiid_n1` — 46 unchanged, 2 changed (both intended):**

```
NTV torque FGAR (Re,Im) [N·m]   [2 elem]  [2 elem]  8.237e+01 (99.33%)   ** CHANGED **
NTV energy FGAR (Re,Im) [J]     [2 elem]  [2 elem]  7.468e+00 (100.93%)  ** CHANGED **
```

Everything else — eigenvalues, Δ', island widths, Chirikov, resonant field, all four PE energies, ballooning/Mercier profiles — is bit-identical or at roundoff. The NTV quantities move by ~100% because NTV ∝ δB², and the χ₁ fix removes a ×1.648 over-scale of ξ^α; that squared factor is the whole point of the PR.

**`solovev_n1` — 21 unchanged.** Ideal path untouched.

**`solovev_kinetic_calculated` / `solovev_kinetic_nuzero` — self-consistent kinetic-DCON path:** eigenvalues shift by ~1e-5 absolute (0.00% relative), ODE step counts move by ≤14 of ~1800. Expected: these are ψ-integrated quantities largely insensitive to the local bounce quadrature, and the residual is roundoff amplified by the adaptive outer ODE.

### Refactor is numerically inert

Tracking the two NTV quantities across the branch shows the change lands entirely at the physics-fix merge, and the subsequent buffer/backend refactor moves nothing:

```
50440732 (develop)  [2 elem]  --                   --
77c514d5 (physics)  [2 elem]  7.468e+00 (100.93%)  ** CHANGED **
local    (refactor) [2 elem]  8.3e-17              OK
```

(`ntv_torque_fgar` likewise: 99.33% at `77c514d5`, then 1.7e-15.)

All 255 `runtests_kinetic.jl` tests pass.

## Independent physics audit

I also had the `fortran-physics-reviewer` agent re-derive all four fixes against `torque.F90` / `spline.f` / `dcon_interface.f` / `gpout.f` independently of the PR description. **Verdict: PASS on all nine claims**, with the spline-quadrature algebra and the affine-fit identity checked by hand rather than taken on faith. Two findings worth recording:

- `gpout.f:5637` is `xmsout = xmsout/chi1` with the comment "remove DCON normalization" — confirming `clebsch_alpha` is already physical ξ^α and the ×χ₁ was a double-application.
- Dropping `SINGULAR_EPS` from `(pl + (1−σ)/pl)` is safe and *more* faithful: `pl` is `cis` of a real, so `|pl| = 1` exactly and the guard was never doing anything.

The `nlmda` 64→128 change also matches the Fortran default (`torque.F90:53`), which the audit confirmed independently.

logan-nc:
## Correction to my summary above — two errors, and a finding that matters

@logan-nc caught both. Recording the investigation.

### 1. The regression table and the Fortran benchmark are *different cases*

I presented them as if comparable. They are not:

- **Fortran benchmark** (`DIIID_ideal_example`, real DIII-D EFIT, Fortran-supplied ξ): `T_fgar = 1.7685` vs Fortran `1.7767` (**0.46 %**), `dW_fgar = 0.13613` vs `0.13694` (**0.59 %**). Agreement with Fortran is intact and slightly better than the table in the PR body.
- **`diiid_n1` regression case** (`examples/DIIID-like_ideal_example`, synthetic TkMkr equilibrium, nl=6, coil forcing, ξ from the Julia PE pipeline): `T = 0.5597 + 0.1383i`, `dW = 0.0692`. **There is no Fortran reference for this case.** The `82.93 → 0.56` move is not a departure from Fortran.

### 2. `total_energy` is not complex — my tracked quantity was wrong

`Compute.jl:221` is `total_energy = complex(imag(total_torque)/(2*ctrl.nn), 0.0)`. The imaginary slot is structurally zero; δW is exactly `Im(T)/(2n)`. Tracking it as `(Re,Im)` was meaningless. Fixed in `7d289385` — now a scalar with `extract = "real_first"`, labeled `NTV kinetic energy dW FGAR [J]`:

```
NTV kinetic energy dW FGAR [J]   -7.398883e+00   6.914966e-02   ** CHANGED **
```

### 3. The finding: the 99.33 % move is the χ₁ commit alone, and the benchmark is blind to it

I said the ~100 % change was "the χ₁ squared factor." That hand-wave was wrong — a 1.648× change in ξ^α cannot give 148×. I isolated it: re-running `diiid_n1` at PR HEAD with **only** `clebsch_alpha .* chi1` restored gives

```
total_torque = [84.0223, -16.1909]      (develop: [82.9326, -14.7978])
total_energy = [-8.0954,  0.0]          (develop: [-7.3989,  0.0])
```

So **all** of the 148× swing and the δW sign flip come from that one line; the three bounce-averaging fixes account for the ~1 % residual. It is that large because ξ^α is rescaled *alone* — ξ^ψ is untouched — which shifts the `divx`/`dbob` balance through a near-cancellation rather than scaling the perturbation coherently.

**Is the new value right? Yes** — verified against the Fortran consumer, not just `gpout.f`. `read_peq` passes the `gpec_xclebsch` column straight into `xs_m(3)` with no χ₁ factor (`pentrc/inputs.f90:807`), and `set_peq` documents it: *"the contravariant clebsch displacement xi^alpha ... NOTE this is NOT DCON CONVENTION -> xmsmns/chi1 in DCON"* (`:758-759`). The `force_xialpha` fallback recomputes the same array with an explicit `(1.0/chi1)` factor (`:728`). So PENTRC's `xs_m(3)` is `xms/chi1`, exactly what `clebsch_alpha` holds. develop was feeding `ξ^α·χ₁` into the operator. The corrected `T = 0.56 N·m` is also far more physical than `83 N·m` for this case.

**But the benchmark cannot see this bug.** Before `c8b1f6af`, the benchmark divided by χ₁ and `set_perturbation_data!` multiplied it back; the commit removed *both*. Net identity — the operator receives the same ξ^α either way. So:

> The dW improvement in the PR body table (16.94 % → 3.04 %) is **not** attributable to bug #1. That benchmark bypasses the PE pipeline by feeding Fortran's ξ directly, so it is invisible to the χ₁ line. The benchmark improvement comes from bugs #2–#4 (bo, θ-quadrature, vspl). Bug #1's real effect appears only through the Julia PE pipeline — where it is a factor of 148.

I have corrected the PR body accordingly. **This also means bug #1 has no Fortran-referenced regression coverage**: the only case exercising it (`diiid_n1`) has no Fortran ground truth, and the case that has ground truth cannot exercise it. Worth a follow-up — a PENTRC reference run driven by Julia's own PE ξ would close it.

logan-nc:
## Status + handoff: where this PR actually stands

Writing up the full picture so this can be picked up cleanly. Branch is at `7d289385`, pushed, CI green. **Nothing below is committed yet — this is the recommended plan, not work already done.**

---

### 1. What the four fixes actually buy (this corrects the original PR body)

The headline claim in the original PR body — that the χ₁ fix drove the Fortran-benchmark dW error from 16.94% → 3.04% — **was wrong**, and I've corrected the body. The real accounting:

**The Fortran benchmark is structurally blind to bug #1.** Commit `c8b1f6af` removed *both* the benchmark's `./chi1` *and* the compensating `.*chi1` inside `set_perturbation_data!`. Those two cancel exactly, so the benchmark (`DIIID_ideal_example`: real EFIT, Fortran-supplied ξ) sees no change at all from the χ₁ fix. On that benchmark the improvement from **7.23% → 3.04% dW error comes entirely from bugs #2–#4** (axis `bo`, exact cubic θ-quadrature, extrap `v_par`/bounce-root source).

**Conversely, on `diiid_n1` (Julia PE-pipeline ξ, synthetic TkMkr, coil forcing) >99% of the swing is bug #1 alone.** Verified empirically, not inferred: re-running at HEAD with *only* `clebsch_alpha .* chi1` restored reproduces develop's numbers (`total_torque = [84.02, -16.19]` vs develop's `[82.93, -14.80]`). Bugs #2–#4 are the ~1% residual on that case.

So the review comment that "everything is in the noise except χ₁" is half right — it's true on `diiid_n1`, and false on the Fortran benchmark, where χ₁ is invisible and #2–#4 are the whole story. **Both sets of fixes should stay.** They just show up in different places, which is itself the reason for follow-up (5) below.

### 2. The regression quantity is a scalar, not a complex pair

`Compute.jl:221` stores δW = Im(T)/(2n) into the **Re** slot of a complex; the Im slot is structurally always zero. The earlier version of the regression case tracked it as a `(Re, Im)` pair, which was meaningless. Fixed in `7d289385` — `diiid_n1` now extracts `real_first` for `ntv_energy_fgar`.

### 3. The FastInterpolations internal-API refactor (`1e462651`) should be reverted

Commit `1e462651` swapped the self-contained `_fortran_spline_*` quadrature helpers for `FastInterpolations._solve_system!` — a **non-exported** function — to get an allocation-free fit-then-integrate in the per-λ loop. I originally defended this on speed grounds in the thread at :156. **That defense was wrong and I withdraw it.** Re-examined:

- **The speed edge is run-to-run noise.** Variant B (FastInterpolations): 86.9 / 87.4 / 93.2 s. Variant A (helpers): 97.9 / 89.0 / 90.2 s. Within-group spread exceeds the between-group gap (t≈0.9, p>0.3). There is no measurable win.
- **Neither variant allocates in the quadrature.** Both are already zero-alloc there, so the internal dependency isn't even buying allocation reduction — the preallocation commit `fe346d44` had already achieved that.
- **The real per-λ heap traffic is elsewhere** (see §4).
- `Project.toml` pins `FastInterpolations = "0.4.10"`, which caret-resolves to *any* 0.4.x. A private-API rename in a fast-moving package would break us silently.

The `clean-code-reviewer` and `julia-performance-optimizer` agents split on this (clean-code: revert; perf: keep and pin `[compat]`, citing CLAUDE.md's "don't reimplement spline integration"). **I side with revert.** The "don't reimplement" rule doesn't cleanly apply when the needed public primitive *doesn't exist* — and confirmed with the fast-interpolations specialist that **no public API does allocation-free fit-then-integrate** (no exported `refit!`, no total `integrate!`; `cubic_interp!` is fit-and-evaluate-at-query-points). Array pools don't close that gap either. So the choice is: depend on a private symbol for zero measured benefit, or own ~90 lines of well-tested, Fortran-faithful quadrature. The latter is the smaller liability.

### 4. The actual compute lever was never the quadrature

The per-λ allocations that matter are both in `_find_bounce_points_and_grid`:

- **`powspace()`** (`BounceAveraging.jl:422` / `:427`) rebuilds the θ-subgrid every λ — 5–8 arrays, ~5 KB per pitch angle.
- **`Roots.find_zeros`** (`:406`) allocates a closure + roots vector per trapped λ.

A `powspace!` writing into a preallocated `BounceScratch` buffer is the real optimization here. That was never touched by this PR.

---

### Recommended TODOs (in order)

1. **Revert `1e462651`.** Restore the self-contained `_fortran_spline_*` helpers and their preallocated `BounceScratch` workspaces from `fe346d44`, replaying the `3ea624b9` comment cleanup on top. Drops the `FastInterpolations._solve_system!` dependency entirely.
   - Numerically inert. Verify with `regress --cases diiid_n1 --refs 77c514d5,local` → expect **zero diff on every quantity**. No re-benchmark needed (the speed difference is noise).
   - *If a reviewer prefers to keep Variant B instead:* it must at minimum pin `FastInterpolations = "~0.4.10"` in `[compat]` and carry a guard test asserting `_solve_system!` still exists with the expected signature.
2. **Keep, unchanged:** the χ₁ fix, the `bo`/quadrature/`vspl` physics, the comment-hygiene pass, and the `diiid_n1` regression additions (`ntv_torque_fgar`, `ntv_energy_fgar`).
3. **Optional, this PR or a follow-up:** add `powspace!` writing into `BounceScratch` to eliminate the ~5 KB/λ grid allocation (§4). This is the genuine performance work; the quadrature refactor was not.
4. **Re-run the regression harness** against `develop` and post the table (mandatory per CLAUDE.md). Expect: kinetic quantities move on `diiid_n1` (the physics fix), all ideal quantities unchanged, `solovev_n1`/`solovev_multi_n` unchanged.

### Follow-ups (not blocking this PR)

5. **Coverage gap that made this whole thing hard to reason about:** no case has *both* Fortran ground truth *and* a Julia-PE-pipeline ξ. The Fortran benchmark supplies ξ externally (so it can't see bug #1); `diiid_n1` uses the PE pipeline but has no Fortran reference (so it can't validate the absolute number). Bug #1 therefore has **no directly Fortran-validated check anywhere**. Building one case that closes this is the highest-value next thing.
6. tgar ~5% trapped/passing-boundary residual.
7. `atol_psi` / `rtol_psi` defaults.

cc @jhalpern30 — the :156 spline-helpers thread is the one that changes: I'm now recommending we go back to your original position and drop the `FastInterpolations` internal call.

src/KineticForces/BounceAveraging.jl
logan-nc: Correction to my reply above — **your original instinct was right and I now recommend reverting to the self-contained helpers.**

I justified swapping them for `FastInterpolations._solve_system!` on the grounds that it was allocation-free and faster. Re-checked both claims:

- **Faster: no.** The gap is run-to-run noise (helpers: 97.9 / 89.0 / 90.2 s; FastInterpolations: 86.9 / 87.4 / 93.2 s — within-group spread exceeds the between-group difference).
- **Allocation-free: yes, but so were the helpers already**, once `fe346d44` moved their buffers into `BounceScratch`. The private-API call bought nothing over that.

And it costs something real: `_solve_system!` is non-exported, while `Project.toml` pins `FastInterpolations = "0.4.10"`, which caret-resolves to any 0.4.x — a rename upstream breaks us silently. Confirmed with the FastInterpolations specialist that **no public API offers allocation-free fit-then-integrate** (no exported `refit!` or total `integrate!`), so there is no third option that is both public and allocation-free.

The per-λ allocations that actually matter were never in the quadrature — they are `powspace()` rebuilding the θ-subgrid (~5 KB per pitch angle) and `Roots.find_zeros`, both in `_find_bounce_points_and_grid`. That is the real optimization target and is untouched by this PR.

Plan is to revert `1e462651`, restoring the helpers with their preallocated buffers and keeping the comment cleanup. Numerically inert. Full write-up in the status comment on the PR.

logan-nc:
## Update: quadrature backend resolved with a third option (supersedes the revert plan in my status comment above)

Pushed `e67c4a7d`. The status comment above recommended **reverting `1e462651` to the
self-contained `_fortran_spline_*` helpers**. I found a cleaner resolution that avoids *both*
options we disliked — the non-exported `FastInterpolations._solve_system!` **and** ~90 lines of
hand-written spline quadrature.

### The insight the earlier thread missed

The bounce θ-integrals live on a **fixed** unit grid (`range(0,1,ntheta)`; the `tdt` weights
carry `dθ/dx`). On a fixed grid the exact integral of the `CubicFit`-endpoint spline is a
**constant linear functional of the node samples** — so "fit then integrate" is just a
precomputed weight contraction:

- total: `∫ = w·y`
- cumulative: `fsi = W·y`

`w` and `W` are built **once per `ntheta` by evaluating the public `integrate` /
`cumulative_integrate!` on the unit basis vectors** (so they are bit-faithful to the operator
they replace, obtained through the public API — not a reimplementation), cached, and shared
read-only. The per-λ hot loop becomes `dot` / `mul!`.

Result — this clears every objection in the `:156` thread at once:

- **no non-exported symbol** — `_solve_system!` is gone; only `cubic_interp`/`integrate`/
  `cumulative_integrate!`/`CubicFit` remain, so the `FastInterpolations = "0.4.10"` caret pin is
  no longer a silent-breakage risk;
- **no hand-written spline helpers** — the deleted `_fortran_spline_*` code stays deleted;
- **bit-faithful** — reproduces `integrate`/`cumulative_integrate!` to ~1e-16 on smooth,
  1/√-endpoint, and complex integrands;
- **allocation-free** in the per-λ loop (`mul!`/`dot` into `BounceScratch`).

Diff is one file, +58/−38.

### Root finder — benchmarked as requested, keeping `Roots.find_zeros`

Benchmarked the bounce-point root finder against a preallocated bracketed sign-change scan on
realistic DIII-D `B(θ)`: the scan is **17–31× faster with ~80× fewer allocations** and matches
`find_zeros` to ~1e-14 in the bulk — **but** disagrees by ±1 near-tangent pair right at the
trapped/passing boundary (the known ~5 % tgar-residual region) and is non-monotonic in grid
resolution. That makes it **not numerically inert**, so I'm keeping `Roots.find_zeros` here and
leaving the scan as a follow-up to pair with the tgar-boundary work (both live at the same
marginal λ).

### Validation

- **Quadrature swap is inert.** `regress --cases diiid_n1 --refs 77c514d5,local` (spans the
  whole custom-helpers → private-API → public-weights history): **NTV dW FGAR diff 5.4e-9, NTV
  torque FGAR diff 1.7e-6**. Kinetic-in-the-loop eigenvalues (`solovev_kinetic_calculated`,
  `solovev_kinetic_nuzero`) move only at roundoff (et Re/Im ~1e-7…1e-6, 0.00 %).
- `regress --refs develop,local`: the only real move is `diiid_n1` **NTV torque +99.33 %** (the
  χ₁ fix); `solovev_n1`/`solovev_multi_n` unchanged (ideal diffs ~1e-8 + a few ODE steps).
- **255/255 `runtests_kinetic.jl` pass.**

### Note on the `diiid_n1` regression noise

That case runs `kinetic_factor=0`, so NTV is a post-hoc diagnostic and the quadrature change
cannot touch the ideal DCON ODE or the PE singular coupling. The near-resonant quantities
(Chirikov, island half-widths, resonant b^r) that appeared to swing ~16 % against one run come
out **unchanged** against `develop` — they're **run-to-run nondeterminism** from the
`Threads.@threads :greedy`/`:static` ideal/PE path amplified by the adaptive ODE, not the
quadrature. Worth a separate determinism follow-up; these are noisy regression targets.

logan-nc (approved):

Issue #309, opened 2026-06-28
Kinetic-aware equilibrium grid packing: place knots on rational and kinetic-resonance surfaces


logan-nc:
Sub-issue of #302.

### Motivation
#302 showed the auto grid (`log_asymptotic`, `mpsi=0`) places knots by a curvature/separatrix heuristic that is blind to where the solution is actually sharp — it under-resolves the edge and, to converge, over-resolves the mid-region (~4× the knots `ldp` needs). The right fix is to *place knots where the physics has sharp features* rather than globally tightening a tolerance.

With #112 (KineticForces/NTV) merged, the grid must also serve kinetic calculations, which add new sharp radial features the current heuristic ignores:

1. **Rational surfaces (q = m/n).** Orbit-pitch / bounce-harmonic resonances near rationals produce large, sharp kinetic features. The grid should land points *on* the rationals. (This also directly addresses the #302 near-separatrix-rational symptom, independent of kinetics.)
2. **Kinetic resonance surfaces.** The superbanana-plateau resonance near ω_E ≈ 0 and other bounce/precession resonances (Logan 2013, §IV–V) are sharp and localized. Their ψ-locations are cheaply estimable from cylindrical approximations of the resonance operator evaluated at thermal energy x = 1.

### Proposal: a two-pass "form → refine grid → reform" workflow
The equilibrium↔kinetic ordering is physical: resonance frequencies (ω_b, ω_d) need the formed geometry, and rational surfaces need the q-extrema that only exist after the first build. The solver also builds `psi_nodes` before the field-line integration that produces the geometry, so a true single pass is impossible. Equilibrium formation is fast and the existing rerun path re-forms from the same in-memory input (no file re-read), so a second pass is cheap.

Driver flow:
1. Form the equilibrium once on the default/coarse grid.
2. `refined_psi_nodes(equil, base_nodes; kinetic_profiles, n, m-range, …)` → augmented ψ vector: union of (a) rationals q=m/n (q forced positive, via the existing `sing_find!` bisection on the q-spline + q-extrema) and (b) kinetic-resonance ψ where Ω(x=1; ψ) = ℓ_eff·ω_b·√x + n·(ω_E + ω_d·x) changes sign (cylindrical ω_b/ω_d from Logan 2013), each locally packed with a few knots.
3. Re-form on the augmented grid via the existing in-memory rerun path; discard the first pass.

[...]

PR #295, opened 2026-06-21
Add D3D-like kinetic profiles and update example equilibrium

d-burg:
Adds kinetic profiles and an NTV-torque demonstration to the DIII-D-like example, and fixes a grid setting that was producing inaccurate ideal stability results. Closes #268.

Changes
- New kinetic profile file (`TkMkr_D3Dlike_Hmode_kinetic.gpeckf`): six columns
  psi_n, n_i, n_e, T_i, T_e, omega_E.
- Replaced the example equilibrium with a new synthetic and self consistent geqdsk that uses the tokamaker DIII-D limiter and coils (toroidal field sign now negative, matching the physical DIII-D convention).
- Added a [KineticForces] section to the example gpec.toml, which turns on the NTV-torque diagnostic. The stability solve stays ideal; NTV torque is computed as a post-PE diagnostic.
- Added an NTV-torque regression quantity to the diiid_n1 case.

Grid fix
- The example used `mpsi = 0` (auto), which under-resolves the edge (~11 edge points) and produced a spurious, sign-flipped delta_W and unphysical edge delta-prime (order 1e4). Switched to `grid_type = "ldp"` with a fixed `mpsi = 256`. Ideal delta_W and the inner-surface delta-prime values now agree with Fortran DCON.

Test updates
- Re-pinned the DIII-D parallel-integration regression for the corrected equilibrium and grid. `et[1]` and the inner-surface (q = 2, 3, 4) delta-prime are pinned tightly; the near-separatrix surfaces (q = 5, 6) are only checked for finiteness, since their values and signs are not grid-converged.

Notes
- The `diiid_n1` regression baseline shifts as expected from the equilibrium and grid changes; the new values are the correct ones.
- Two follow-ups are out of scope here: the auto-mpsi edge under-resolution in the equilibrium grid builder, and more thorough testing of `grid_type='log_asymptotic'`

image
image

PR #179, opened 2026-03-01
EQUIL + ForceFreeStates - Extend equilibrium splines to separatrix for diverted plasmas

logan-nc:
### Equilibrium: Edge inverse splines for diverted plasmas

- Adds `InverseCubicSpline{S}` type wrapping an iota = 1/q inner spline; callable returns 1/inner(ψ), enabling well-behaved evaluation of q and dV/dψ as psin → 1 in diverted plasmas (where q → ∞)
- Makes `ProfileSplines` mutable with a `q_spline::Union{S,I}` pointer field temporarily swapped to the edge inverse spline for ODE integration chunks that extend beyond psihigh
- Adds `direct_xpoint_find` using Newton iteration for Br=0, Bz=0 with physically motivated initial guesses on the inboard side; correctly classifies diverted vs limited topology (critical fix: initial R guess at `rs1 + 0.35*(ro-rs1)` rather than midpoint)
- For diverted plasmas, builds edge iota inverse splines over [0.9·psihigh, 1.0] with anchor iota(1.0) = 0 (CubicFit BC) enforcing the known separatrix physics
- Sets `qa = Inf` for diverted plasmas; `q_for_mode_range` fallback prevents `trunc(Int, Inf)` in mode range computation
- dV/dψ evaluation in `Free.jl` routes to `dVdpsi_spline_inv` when psilim > psihigh
- New `EquilibriumParameters` fields: `r_xpoint`, `z_xpoint`, `is_diverted`
- `ProfileSplines` fields: `q_spline_direct`, `q_spline_iota_inverse`, `dVdpsi_spline_inv` (all Nothing for limited plasmas → identical behavior to before)

### ForceFreeStates: Integration beyond psihigh for diverted plasmas

- Extends rational surface search in `sing_find!` above psihigh using the iota spline, stopping when surface spacing < `edge_layer_width` (new TOML parameter, default 1e-4)
- **Removed `set_psilim_via_dmlim` and `dmlim`** — integration limit now determined entirely by `psiedge < psilim` logic; `psilim` for diverted plasmas is set from the last edge rational surface found
- Moved `sing_find!` before `sing_lim!` in `JPEC.jl` so edge surfaces above psihigh are available when `sing_lim!` reads them
- `sing_lim!` sets `psilim = last_edge_surface + edge_layer_width/2` for diverted, giving a valid final ODE chunk beyond the last crossed surface
- Fixed `intr.mtheta` → `ctrl.mthvac` bug in `free_compute_wv_spline`
[...]

PR #312, opened 2026-07-02
EQUIL + GPEC - Two-pass measured-curvature auto psi grid with rational-surface knots

### LINKED ISSUE #302: auto mpsi under-resolves the edge, giving spurious dW and Delta-prime
With grid_type = log_asymptotic and automatic radial resolution (mpsi = 0), the
optimal-mpsi heuristic allocates very few edge grid points (about a dozen near the
boundary). The resulting edge under-resolution produces spurious ideal dW and
Delta-prime values, especially when a near-separatrix rational surface is captured.

Workaround: set a fixed mpsi (e.g. mpsi = 256), which resolves the edge and gives
stable dW / Delta-prime.

The fix should make the automatic mpsi selection allocate adequate edge resolution
(or warn when the edge is under-resolved). Note this is an auto-mpsi problem, not
log_asymptotic itself: log_asymptotic with a fixed mpsi is fine.

### LINKED ISSUE #308: Ballooning: guarantee pedestal resolution in auto-gridding, and make a default psi-window the standard for ballooning scans
Sub-issue of #302. #302 covers auto-mpsi under-resolving the edge (spurious dW / Δ′); this issue covers the ballooning-stability consequences and a paired performance default.

## 1. Auto-gridding must guarantee pedestal resolution for ballooning physics

High-n ballooning stability (Δ′, the s–α first/second boundaries in `src/ForceFreeStates/Ballooning.jl`) is physically most relevant in the pedestal, ψ_N ≈ 0.8–0.98. The edge under-resolution in #302 (N_edge ≈ 5 at the default `psi_accuracy = 1e-3`) under-resolves this whole band, not just the last rational surface — so the ballooning boundary profile there is unreliable. Whatever fix lands for the N_edge/N_mid imbalance in `make_optimal_mpsi` should explicitly ensure adequate, efficient resolution across the **0.8–0.98 pedestal band** (not merely a handful of points at ψ→1), for `log_asymptotic` auto-mpsi as well as `ldp` / power gridding.

## 2. Make a default psi-window the standard for ballooning scans (performance)

The α-boundary drivers (`ballooning_alpha_boundary` / `ballooning_alpha_boundaries`) loop over *every* equilibrium ψ node and run up to `n_scan` (~24) ODE shooting solves per surface — the dominant local-stability cost (~24.5 s / 54 GB on the DIII-D case). The tightly packed axis (ψ<0.1) and far-edge (ψ>0.98) nodes produced by `ldp` / `log_asymptotic` / power gridding waste solves where ballooning boundaries are rarely used. Proposal: by default evaluate the ballooning α-boundary scan only on

```
ψ_N ∈ [0.1, min(0.99, psi_edge, psihigh)]
```

which trims the worst axis and edge packing across all grid types while retaining the pedestal.

## Ordering / dependency

(2) depends on (1). Windowing a grid that is under-resolved in the pedestal just locks in the spurious values from #302, so the pedestal-resolution fix must land before or together with the default window.

## Implementation notes (from prior analysis)

- Clean approach for (2) is an **index window/mask over the existing grid** in the two α-boundary drivers, reusing the existing NaN sentinel (the drivers already `fill(NaN, npsi)` and `continue` on `xs[i] > 1.0`; outputs stay length-`npsi`, aligned to the ψ axis). A *separate* ballooning grid is too dirty: `prepare_ballooning_coefficients`, `salpha_reference`, and `resistive_interchange_h` read **precomputed nodal** bicubic partials (`rzphi_*.nodal_derivs.partials[k, idx, itheta]`) and spline node values (`profiles.<spline>.y[idx]`) that exist only at grid nodes — off-node evaluation re-architects all three and pays the geometry-bicubic cost.
- Leave `compute_ballooning_stability!` (di / dr / Δ′, locstab cols 1/2/4) **un-windowed or finite-filled**: its `locstab_fs` is splined and evaluated at singular surfaces (`intr.locstab(sing.psifac)`), so a NaN at a skipped node would poison the whole profile.

## Config question for reviewers (please weigh in)

I'd prefer **not** to add TOML window specifiers if everyone agrees the hard-coded default `[0.1, min(0.99, psi_edge, psihigh)]` is sufficient. Two things to confirm:

1. Is a fixed default window acceptable, or do we need a `[ForceFreeStates]` config knob?
2. Does `min(0.99, psi_edge, psihigh)` still span the full pedestal band of interest in your cases (i.e. it must not clip ψ_N ≈ 0.8–0.98)?

logan-nc:
Closes #302, closes #308. Implements the rational-surface half of #309 (kinetic-resonance knots remain a follow-up; the mandatory-node hook is in place).

## Problem

The one-pass auto grid (`log_asymptotic`, `mpsi=0`; the auto grid_type is renamed `"auto"` in this PR, with `"log_asymptotic"` kept as a working alias) sized knots from a-priori models blind to the formed solution: ~5 edge knots at the default `psi_accuracy=1e-3` (spurious dW/Δ′, #302), while tightening the tolerance over-packed the mid-region 4× because N_mid ∝ τ^(−1/2) vs N_edge ∝ τ^(−1/4).

## Approach

**Two-pass measured-curvature refinement** (`src/Equilibrium/GridRefinement.jl` + driver orchestration):

1. Pass 1 forms the equilibrium on a coarse fixed-128 log_asymptotic layout.
2. A knot density is built from **measured nodal data** — divided-difference fourth derivatives of the 1D profiles, the rzphi geometry channels along sampled θ-lines (the bicubic ψ-axis shares these knots), and the kinetic profiles when loaded (steep pedestal gradients attract knots, #308 part 1). Nodal values come from independent field-line integrations, so the estimate is immune to inter-knot spline ringing.
3. Mandatory knots are pinned on every rational surface q=m/n in the requested n range (`ForceFreeStates.rational_psi_nodes`, extracted from `sing_find!`), each with local geometric packing; a δ_min snap guard prevents near-duplicate nodes (#309).
4. Pass 2 re-forms via the existing in-memory rerun path (`override_psi_nodes` keyword through all solver paths) — no file re-read.

Key modeling decisions (each validated on the DIII-D example):
- **h³ derivative error model** `err(f′) ≈ h³|f''''|/24`: the stability physics consumes spline *derivatives* (q′ at rational surfaces, p′/V′ in the EL and ballooning coefficients). Every region scales as τ^(−1/3), fixing the #302 imbalance.
- **Curvature measured in ρ=√ψ**, where the equilibrium is regular at the axis; the ψ-space geometry channels diverge there and made the implied knot count grow without bound.
- **Local packing around rational surfaces** (spacing 0.06·τ^(1/3), radius 0.05) converges the Δ′ BVP, which samples the ψ-splined coefficient matrices around each surface — a pinned knot alone left Δ′@q=2 2× off.
- The core below ψ=0.03 uses the a-priori geometric density only (smallest-surface nodal data is integration/extrapolation noise).

Also included:
- **Ballooning α-boundary scan window** ψ_N ∈ [0.1, min(0.99, ψ_edge)] (hard-coded per the #308 discussion, NaN-sentinel index mask, locstab path untouched) — 93 of 200 surfaces skipped on the DIII-D case.
- **q-vs-ι benchmark** (`benchmarks/benchmark_q_vs_iota_edge.jl`): splining ι=1/q gives a ~1.2–2× constant-factor gain but the same convergence order, documenting that wholesale ι replacement is not warranted for grids ending inside the separatrix (PR #179 lineage owns separatrix-extension ι work).

## Validation (DIII-D-like example, n=1, defaults)

- 200 knots (vs ~1012 the old model needed for a converged edge): top eigenvalue within **0.01%** of ldp-512; Δ′ inner surfaces (q=2,3,4) within ~5%; near-separatrix surfaces q=5,6 match the dense references where ldp-256 was 35% off / wrong-signed — the two-pass grids are internally converged there (−217.5/−217.6/−218.1 across 268/293/454 knots) while ldp is still wandering (+205 → −146).
- `fortran-physics-reviewer` audit of the density mathematics: PASS (divided-difference table, sharp Hall–Meyer h³/24 constant, ρ-Jacobian, edge-floor algebra dlog=(4τ)^(1/3), equidistribution, merge semantics, rerun dispatch); its two MINOR robustness suggestions are applied.
- Full test suite green, including new `test/runtests_grid_refinement.jl` (38 assertions: merge invariants, equidistribution, quartic-exact d4, τ^(−1/3) scaling, Solovev two-pass round-trip).
- Regression harness vs develop: `solovev_n1` and `solovev_multi_n` **bitwise-unchanged** (fixed-grid path untouched). `diiid_n1` moves as expected — equilibrium scalars at spline-resolution level (≤1e-5 relative); grid-sensitive rows (energies, Δ′, island widths, resonant field, PE torque) shift because the develop baseline was computed on the defective under-resolved grid; the new values match the dense references above. **The diiid_n1 baseline should be re-pinned on merge.**
- The DIIID parallel-integration testsets now build the two-pass auto grid exactly as the driver does and pin its values — the auto grid is the production default, so that is what regression guards (near-separatrix Δ′ entries asserted finite-only, per the grid-convergence evidence; the ldp-512 equivalence table is recorded in the PR comments).

## Notes for reviewers

- PR #295's `ldp`+`mpsi=256` workaround can flip back to `mpsi=0` auto after this merges (and re-pin its regression).
- Standalone `setup_equilibrium` callers with `mpsi=0` now get the coarse pass-1 grid unless they refine via `refined_psi_grid` + `override_psi_nodes` (documented in the docstrings and `docs/src/equilibrium.md`).
- #309's kinetic-resonance knots plug into `merge_mandatory_nodes`/`refined_psi_grid(...; mandatory=...)` without further Equilibrium changes.


logan-nc:
## Knot placement: auto two-pass vs ldp (for the record)

Radial knot packing — DIIID-like example

*Top: cumulative knot fraction (index/mpsi) vs ψ_N — where each grid spends its budget. Bottom: local spacing Δψ_N (log scale). Dotted verticals are the n=1 rational surfaces. In the top panel the two ldp curves overlap exactly (normalized sin² packing is N-independent). Regenerate with `benchmarks/plot_grid_knot_placement.jl`.*

Reading the figure: the auto grid (blue, **199 intervals**) coasts across the smooth mid-radius at ~10× coarser spacing than ldp, dives at every rational surface (the spacing notches), and packs the core, pedestal, and edge — which is exactly where ldp either wastes knots (mid) or runs out of them (near-separatrix rationals).

## The grid logic, in plain bullets

- **Pass 1**: form the equilibrium once on a coarse fixed layout (~128 intervals). Cheap; its only job is to measure the real solution.
- **Mid-region (0.03 ≤ ψ_N ≤ 0.9): measured spline error.** Take 4th divided differences of the pass-1 *node values* (profiles F, P, V′, q; the four rzphi geometry channels along θ-lines; kinetic n/T/ω_E when loaded) and size the local spacing so the cubic-spline **derivative** error stays below `psi_accuracy` (τ). Node values come from independent field-line integrations, so this measurement can't be fooled by inter-knot ringing. Curvature is measured against ρ=√ψ, where the geometry is regular at the axis.
- **Core (ψ_N ≤ 0.03): a-priori only — NOT the same as the mid.** The smallest flux surfaces' nodal data is integration/axis-extrapolation noise, so measurement is not trusted there; the density is the geometric-in-log(ψ) form outright.
- **Edge (ψ_N ≥ 0.9): measured, with an a-priori floor — also not the same as the core.** The measured density applies, but is floored by geometric-in-log(1−ψ) packing so a pass-1 undersampled edge can't hide. The floor constant comes from uniform *relative* q′ error and is independent of the separatrix slope A.
- **Rational surfaces: pinned + locally packed.** Every q=m/n in the requested n range gets a knot exactly on the surface plus geometric packing around it (this, not the pinned knot alone, is what converges Δ′).
- Equidistribute the resulting density into the final knot vector; **pass 2** re-forms the equilibrium on it from the in-memory input. Every region scales as τ^(−1/3), so one knob refines everything proportionally.

## Rename: `log_asymptotic` → `auto`

Agreed the old name misdescribed the method — there are no q→∞ asymptotics anywhere on the grid (all finite q; the log-form survives only as the edge/core *floor* shape). `grid_type = "auto"` is now the default; `"log_asymptotic"` remains a working alias (two-pass when `mpsi=0`, the three-region log layout when `mpsi>0`), so no existing TOML breaks.

## Reproduces ldp mpsi=512, and test pinning

Convergence evidence on this example (n=1; Δ′ = BVP diagonal, PEST3 convention as stored in `gpec.h5`):

| grid | knots | et[end] | Δ′(q=2) | Δ′(q=3) | Δ′(q=4) | Δ′(q=5) | Δ′(q=6) |
|---|---|---|---|---|---|---|---|
| **auto, τ=1e-3 (default)** | **200** | **0.078112** | 8.47 | −4.03 | −10.23 | **12.35** | **−144.9** |
| auto, τ=1e-4 | 454 | 0.078086 | 8.98 | −3.86 | −10.23 | 12.35 | −217.5 |
| ldp, mpsi=256 | 256 | 0.077800 | 8.79 | −3.88 | −10.21 | 16.62 | +205.2 |
| **ldp, mpsi=512 (reference)** | 512 | 0.078104 | 8.90 | −3.46 | −10.16 | 12.23 | −146.1 |

- The default auto grid matches ldp-512 on the eigenvalue to **0.01%** and on Δ′ to ~5% on the inner surfaces — at 200 knots vs 512. On the near-separatrix surfaces it agrees with ldp-512 where ldp-256 is 35% off (q=5) or wrong-signed (q=6). Tightening τ converges Δ′ further (τ=1e-4 column).
- **Test pinning is now on the auto grid** (commit 42d9cf4a): the DIIID parallel-FM and STRIDE-BVP testsets build the two-pass grid exactly as the driver does (`rational_psi_nodes` → `refined_psi_grid` → ingest re-form) and pin its values — the auto grid is the production default, so that is what regression must guard. The table above is the recorded proof of ldp-512 equivalence; the near-separatrix Δ′ entries are asserted finite-only since they are not grid-converged in any grid family at these counts.

logan-nc:
## Follow-up: default-convergence scan, core ψ→0 handling, and edge-logic assessment

### Should the default converge everything? Scan says: no default change is justified

Full-pipeline scan over (`psi_accuracy`, `sing_pack_coef`) on the DIII-D example (identical BVP path for every row; anchor row densest):

| τ | pack coef | knots | et[1] | Δ′(q=2) | Δ′(q=3) | Δ′(q=4) | Δ′(q=5) | Δ′(q=6) |
| 5e-5 | 0.02 | **574** | 1.31456 | 9.80 | −3.15 | −10.18 | 11.86 | −206.2 |
| **1e-3** | **0.06 (default)** | **199** | 1.31163 | 8.46 | −4.04 | −10.24 | 11.70 | −138.3 |
| 1e-3 | 0.04 | 214 | 1.31147 | 8.44 | −4.01 | −10.13 | 11.63 | −140.3 |
| 1e-3 | 0.03 | 221 | 1.31156 | 8.71 | −3.40 | −10.31 | 11.63 | −140.2 |
| 1e-3 | 0.02 | 235 | 1.31147 | 9.25 | −3.45 | −10.16 | 11.76 | −137.0 |
| 5e-4 | 0.03 | 272 | 1.31291 | 9.30 | −3.09 | −10.25 | 11.34 | −201.7 |
| 3e-4 | 0.03 | 313 | 1.31523 | 9.13 | −3.28 | −10.22 | 11.48 | −194.6 |

What the table shows:
- **Converged at the 199-knot default**: eigenvalue (0.2% of the 574-knot anchor), Δ′(q=4) (±0.5%), Δ′(q=5) (±2%).
- **Not convergeable by grid density**: Δ′(q=2) wanders 8.44 → 9.80 (±8%) and Δ′(q=3) −4.04 → −3.09 (±13%) *without settling* — adjacent settings (221 vs 235 knots) flip q=2 by 6%. The wandering is **correlated with the q=6 edge state**: rows with Δ′(q=6) ≈ −140 form one branch, rows with ≈ −200 another. The residual inner-Δ′ spread tracks the edge-truncation branch (where the ψ-limit/dW-peak lands for that grid), the same mechanism as the long-known q=6 bistability — it is not a knot-count effect, so spending 300–600 knots buys nothing.
- Conclusion: the lean τ=1e-3 default stands; the existing test pins (exact default grid) remain the right regression guard. The Δ′-vs-edge-truncation sensitivity deserves its own issue (BVP outer boundary condition / psilim landing), separate from gridding — happy to file it.

`sing_pack_coef` is now exposed as a keyword on `refined_psi_grid` for future convergence studies (default unchanged).

### How does the auto grid treat the bicubic splines blowing up near ψ→0?

As a **coordinate artifact, which is what it is**: the rzphi channels behave as ψ^(k/2) near the axis (R−R₀ ~ √ψ), so their ψ-derivatives diverge under refinement while their ρ=√ψ derivatives converge. The grid (a) measures curvature in ρ, (b) converts density with the dψ/dρ = 2√ψ Jacobian (√ψ-tight packing toward the axis by itself), and (c) below ψ=0.03 switches to pure geometric-in-log(ψ) packing — constant-ratio spacing ∝ ψ down to `psilow` (~37 knots at psilow=1e-4), i.e. **yes, asymptotically tight** — because nodal data from cm-scale flux surfaces is integration/extrapolation noise that refinement amplifies rather than resolves. Rational-surface pinning+packing still applies at any ψ, including inside the core region.

### Is the different edge treatment needed? Honest assessment

**Mostly no — one consistent logic already governs, and the docs now say so** (commit 053526ad). The measured ρ-space density supplies ~3× more edge knots than the edge model demands on this example (84 vs 29 in ψ>0.95), because the pass-1 layout is already log-packed at the edge — so the edge "floor" is *inactive insurance*, not a distinct treatment. The honest framing, now in the module docstring and `docs/src/equilibrium.md`: **one sizing rule** (cubic derivative error h³|f''''|/24 ≤ τ·scale) applied to **three curvature sources** — measured divided differences (mid + edge), the analytic separatrix log model (edge floor, normally inactive), and the analytic power-law axis model (core, where it *replaces* measurement because tiny-surface nodal noise is irreducible in a two-pass scheme). The core is the only place a genuinely different rule applies, and that is forced by data, not aesthetics.

logan-nc:
## "Curvature of what?" — pedestal-packing attribution, with data

**Definition:** the density measures |d⁴f/dρ⁴| (ρ=√ψ_N) — the quantity in the cubic-spline *derivative* error bound err(f′) ≈ h³|f''''|/24 ≤ τ·scale — for **each** of: the 1D profiles **F = R·B_φ, μ₀·P, dV/dψ, q**; the four **rzphi bicubic channels** (r², η-offset, ν, Jacobian) along every 8th θ-line (worst case over θ); and the **kinetic profiles** (nᵢ, nₑ, Tᵢ, Tₑ, ω_E) when loaded. The final density is the max over all of them. Note the develop example loads no kinetic profiles (no `[KineticForces]` section until #295 merges), so the flagship's pedestal packing rests on the equilibrium quantities alone today; the kinetic n/T/ω_E channels will add their own pedestal pull after #295.

**Is the pressure pedestal gradient appropriately included?** Yes — verified by per-source decomposition rather than asserted:

Knot-density decomposition by source

Band integrals (knots contributed per band if that source acted alone), DIII-D example, τ=1e-3:

| source | [0.30,0.85] | [0.85,0.90] | [0.90,0.95] | [0.95,0.9995] |
|---|---|---|---|---|
| F | 0.0 | 0.3 | 0.6 | 2.1 |
| **P** | 5.1 | **4.2** | **3.6** | 16.9 |
| dV/dψ | 2.4 | 3.9 | 3.6 | 53.5 |
| q | 5.1 | 3.0 | 3.5 | 55.3 |
| rzphi r² | 6.8 | 3.6 | 3.2 | 65.2 |
| **rzphi offset** | 16.3 | **5.0** | **4.2** | 76.6 |
| rzphi ν | 9.1 | 4.3 | 3.4 | 66.4 |
| rzphi jac | 2.4 | 3.9 | 3.6 | 53.5 |
| floors (core/edge/min) | 2.8 | 0.3 | 1.2 | 32.5 |
| rational packing | 25.6 | 9.0 | 4.7 | 12.6 |

Reading:
- **The pedestal band is measured-curvature-driven, not floor-driven.** At every node in ψ∈[0.8, 0.98] the winning source is measured (the η-offset geometry channel, 62–588 knots/unit-ψ), well above the edge floor (5–315). The floor only matters as insurance.
- **P contributes materially and comparably to q and dV/dψ** (~8 knots across [0.85,0.95] on its own; h≈0.012 pedestal spacing from P alone, scaling as τ^(−1/3)). The geometry channels contribute *more* — which is physically right: the offset/Shafranov-shift steepening **is** the Grad-Shafranov response to that same pedestal p′, and its splines live on the same knot vector. The pedestal is therefore multiply covered, and since everything is measured from the formed solution, the packing follows the pedestal wherever it sits — no hardcoded 0.9–0.95 band anywhere.
- One flagged (theoretical) weakness stays on record: `f_scale = max|f|` normalization suppresses large-offset/low-variation quantities — visible as F's tiny contributions above. Harmless here because F's pedestal information is carried redundantly by P/q/V′/geometry; if a future case ever shows a pedestal that only F sees (it shouldn't, physically), range-based normalization is the one-line fix.

Figure + one-paragraph summary committed to the docs (a145a473).

logan-nc:
@d-burg it seems to me that all the auto grid logic checks out. Default settings give ~200 points for DIIID-like examples. I adde dsub issues to yours to make sure the logic captures the pedestal region well for ballooning and kinetic resonance regions well for kinetic terms. The logic also now explicitly knots on rationals, which I think should be a big win for all applications but I hope for you in particular. 

Please test on the cases that prompted #302 and confirm all is good. If yes, then merge away.

PR #313, opened 2026-07-03
KineticForces - PERFORMANCE - Panel NTV psi quadrature at rational and kinetic-resonance surfaces with rtol-primary tolerances

### LINKED ISSUE #303: NTV torque psi-quadrature over-refines at rational surfaces, dominating runtime
In KineticForces, the outer psi-integration of the NTV torque integrand tpsi(psi)
uses adaptive quadrature. The integrand has near-singular resonant peaks at each
rational surface, and the adaptive quadrature over-refines there with no maxevals
cap, dominating runtime (about 30 min for a single n=1 case vs about 2 min for the
rest of the pipeline). Instrumentation showed ~1375+ tpsi evaluations clustered at
the rational surfaces (q = 2,3,4,5,6).

Proper fix: panel the psi-integral at the rational surfaces (handle the integrable
resonant peaks per-interval) rather than brute-force adaptive refinement; optionally
add a maxevals cap.

Interim mitigation already applied to the DIII-D-like example: relaxed
atol_psi / rtol_psi to 1e-2 (about 4x faster) and removed leftover per-evaluation
debug prints in Torque.jl.

logan-nc:
Closes #303. Built on #312 (`feature/two-pass-psi-grid` is merged into this branch, so its commits appear in this diff until #312 lands — **merge #312 first**, after which this PR reduces to the KineticForces changes).

---

## Update (2026-07) — merged `develop`, and made Δ′ grid-robust

`origin/develop` is now merged into this branch. `develop` had retargeted the DIII-D example to Ip=1.15 MA and switched *that example* to a fixed `ldp`/`mpsi=256` grid; the merge conflicts (DIIID `gpec.toml`, `EquilibriumTypes.jl`, and the parallel-integration tests) were resolved to keep this branch's two-pass `auto` grid running on the new equilibrium.

That surfaced a real problem the merge exposed: on the two-pass `auto` grid the **delta-prime matrix diagonal was not grid-converged** — a `psi_accuracy` (τ) scan swung `dpm[1,1]` (q=2) by ~30%. Root cause: Δ′ extracts the subdominant small-solution coefficient through a `dpsi^{-2α}`-conditioned Frobenius projection, so it tracks the equilibrium-spline error in the *layer around each rational*; the old packing was τ-scaled and its merge snapped out the pinned rational's own neighbors, so the local stencil changed topology with τ. (`et[1]`, a whole-plasma integral, stayed robust to <1% — confirming this is a local singular-layer matching problem, not a global-response one.)

**Fix — `rational_psi_ladder`:** a geometric knot ladder straddling each rational with a fixed absolute innermost step `h0 = 1e-3`, independent of τ and matched to the equilibrium's native radial resolution, emitted as mandatory grid knots. Anchoring `h0` to the equilibrium resolution — *not* the far finer perturbed matching offset `singfac_min/|n·q1|` (~5e-5), which rings the re-splined equilibrium and blew Δ′ up to ~149 — is what makes Δ′ grid-robust. A min-spacing thinning prevents overlapping ladders from nearby rationals (low shear / multi-n) creating sub-`h0` intervals; `merge_mandatory_nodes` collapse now uses an absolute tolerance so fine ladder knots survive.

| DIII-D case | before (τ-scaled packing) | after (ladder) |
|---|---|---|
| `dpm[1,1]` spread across τ ∈ {2e-3…2.5e-4} | ~30% | **~4%** (q3, q4 <1%) |
| `dpm[1,1]` value | 6.4–8.9 (τ-dependent) | **8.40**, matches ldp `mpsi=512` (8.48) |
| Δ′-matrix runtime vs develop's ldp `mpsi=256` | — | **2.2× faster** (158 s vs 353 s) |

The converged value matches the finer ldp `mpsi=512` grid; develop's ldp `mpsi=256` (7.28) undershoots by under-resolving the rationals. In the `diiid_n1` harness, equilibrium, energies, NTV torque, and ‖resonant flux‖ are unchanged (<0.5%); the delta-prime and its downstream island/Chirikov diagnostics move intentionally (the ladder converges Δ′). A `fortran-physics-reviewer` audit passed: the fixed ladder is a faithful analog of DCON (equilibrium splined at native resolution, singular layer handled analytically by the Frobenius matching at `dpsi = singfac_min/|n·q1|`), and it flagged the overlapping-ladder edge case that the min-spacing thinning now guards.

Re-pinned the parallel-integration `dpm` diagonals at `rtol=0.1` on the ladder values (real parts only; `et_par` kept tight); added a grid-refinement test pinning the ladder min-spacing invariant in the multi-n regime. The NTV-quadrature description below is unchanged by this update.

---

## Problem

The outer ψ-integration of the NTV torque `tpsi(ψ)` used a single whole-domain adaptive quadrature with no interior breakpoints, no `maxevals` bound, and a discarded error estimate. The resonant torque-density peaks — at the rational surfaces and at the **kinetic resonances** (dominantly the ExB/superbanana-plateau resonance near ω_E ≈ 0) — forced deep adaptive bisection (1375+ evaluations, ~30 min reported in #303 on the DIII-D-like case). Separately, the `atol_psi = 1e-2` N·m default was an amplitude-sensitive trap: NTV ∝ δB², so a 10× weaker applied field gives a 100× smaller torque and a fixed absolute tolerance can silently dominate termination with O(1) relative error.

## Changes

**ψ-quadrature paneling at resonant surfaces**
- Rational-surface ψ locations the stability run resolved (`sing` + `kinsing`) are threaded into `KineticForcesInternal.sing_psis` and passed as quadrature panel boundaries — Gauss–Kronrod handles peaks at interval endpoints natively instead of hunting them by bisection.
- **Kinetic-resonance locations are identified and paneled for the case's full bounce-harmonic range**: `kinetic_resonance_psi_nodes(kinetic_profiles, equil; n, nl, ...)` scans for zeros of the trapped-branch resonance denominator at thermal energy, Ω_ℓ(x=1; ψ) = ℓ·ω_b(ψ) + n·(ω_E(ψ) + ω_d(ψ)) for every ℓ ∈ −nl:nl, using the pitch-averaged RLAR closed-form frequencies already in `tpsi!` (single spline evaluations, no bounce averaging). The ℓ=0 node is the ω_d-shifted ExB resonance; ℓ≠0 nodes capture the bounce resonances. On the DIII-D case this finds 8 in-range surfaces that line up with the observed pedestal dT/dψ structure (see figure) — including the previously unexplained bump at ψ_N ≈ 0.955.
- Panel boundaries and located resonance surfaces are persisted to the output (`kinetic_forces/<method>/panel_psi`, `resonance_psi`) for diagnostics/plotting; the per-method log line reports torque, error estimate, evaluation count, and panel composition.
- A `fortran-physics-reviewer` audit of the locator passed all checks: frequency conventions match `tpsi!`/`torque.F90` exactly (the `wdhat = q·T/(2εR₀²ZeB₀)` closed form verified algebraically identical), the denominator correctly uses ω_E alone (diamagnetic terms are numerator-only), and leff = ℓ matches the trapped branch. Passing-branch (ℓ+nq transit) resonances are a documented non-goal (broad/Landau-like; quadrature resolves them without dedicated panels).

**Unified helpers (one source of truth per concept, no drift)**
- New `src/KineticForces/Utils.jl`: `find_sign_change_roots(f, grid)` (grid sign-scan + Brent) — also replaces the two pre-existing inline copies of the same idiom in `Torque.jl` (dB/dθ extrema).
- `kinetic_resonance_psi_nodes` is the single kinetic-resonance locator, shared-signature-ready for the #309 grid-knot packing (`refined_psi_nodes` = `rational_psi_nodes` ∪ this). Rational-surface identification stays unified under `_find_rational_surfaces` (shared by `sing_find!` and #312's `rational_psi_nodes`).
- `psi_panel_points` owns sort/dedupe/bound-filtering of the merged panel list.

**Tolerance redesign (rtol-primary)**
- `atol_psi`: 1e-2 → **0.0**. Convergence is controlled by `rtol_psi = 1e-2` (~2 significant figures — the limit of the NTV model's validity; tighter tolerances buy digits the physics can't back).
- New `maxevals_psi = 2000` runaway guard; the error estimate is checked and a clear warning fires when the quadrature terminates without meeting tolerance, or when a user-set nonzero `atol_psi` dominates termination (the weak-field silent-garbage scenario).
- `psilims` stays `[0.0, 1.0]`. (An interim commit floored it at 0.1 to dodge a near-axis dT/dψ spike; measurement showed that was over-optimization — at rtol=1e-2 only 7 of 345 develop evaluations land below ψ=0.1 and the core carries a real 0.3% of the torque — so the floor was reverted. The near-axis feature itself turns out to be two *located* ℓ≠0 resonance surfaces, now paneled like the rest.)

**Coverage**
- New `examples/Solovev_kinetic_NTV_example` (ideal FFS → PE → NTV torque) — previously **no** shipped config exercised this code path — plus a `solovev_kinetic_ntv` regression case pinning the fgar torque and ψ evaluation count.
- Removed the inert `atol_psi`/`rtol_psi` keys from the a10 example (it stops after kinetic stability and never reaches the torque quadrature).

## Benchmarks (DIII-D-like case from #295, n=1, C-coil 1 kA, nl=6, 5 rationals)

| Code | Panels | atol/rtol_psi | Torque (N·m) | ψ evals | KF stage |
|---|---|---|---|---|---|
| develop | – | 1e-3 / 1e-3 (pre-#303 mitigation) | — | 1375+ | ~30 min (reported) |
| develop | – | 0 / 1e-2 | 82.93 − 14.80i | 345 | 158 s |
| **this PR** | 5 rational + 8 kinetic resonance | **defaults** (0 / 1e-2) | 83.16 − 14.92i | **210** | **108 s** |
| this PR | 5 rational + 8 kinetic resonance | 1e-9 / 1e-3 | 83.18 − 14.95i | 450 | 296 s |

The defaults run agrees with the tightly-converged answer to 0.024% in Re(T) at 1.6× fewer evaluations than develop at equal tolerance — and 210 = 14 panels × 15 nodes exactly: the initial pass alone converges, i.e. the panels sit precisely where the quadrature work is. Whole n=1 pipeline: ~3 minutes. Regularization note: the PE default `reg_spot = 0.05` must stay on for this post-PE diagnostic — with raw (unregularized) ξ the "torque" becomes the grid-capped ideal-MHD 1/(m−nq)² divergence (~66,000× larger on a Solovev test) rather than a physical answer.

Torque density with the identified resonance locations (rationals dashed gray; located kinetic resonances Ω_ℓ(x=1)=0 dash-dot violet — note the match to the pedestal structure and the small near-axis feature):

NTV torque density

Quadrature evaluation clustering — develop (no panels) goes near-vertical at the edge resonances; the paneled defaults run spreads effort almost uniformly:

Evaluation clustering

Cumulative torque integral — all runs overlay; ~95% of the torque accumulates in the pedestal, with the sharp rise starting at the ℓ=0 kinetic resonance:

Cumulative torque

## Regression

- `solovev_n1`: 21/21 unchanged; `diiid_n1`: 46/46 unchanged vs `feature/two-pass-psi-grid` (this PR's changes are isolated to `[KineticForces]`-configured runs).
- `solovev_kinetic_ntv`: NTV torque and evaluation count move intentionally with the quadrature improvements (final baseline pinned at this head).
- Unit tests: 222 kinetic (new: root-scan, resonance-node scan with analytic locations, panel assembly, tolerance warnings) + 39 grid-refinement, all green.

## Notes for reviewers

- The pre-existing `solovev_kinetic_calculated`/`solovev_kinetic_nuzero` regression cases are broken on develop and this branch alike (`[SOL_INPUT]` in a separate `sol.toml` that `build_inputs_from_toml` doesn't resolve) — separate fix needed; not touched here.
- PR #295's interim `atol_psi/rtol_psi = 1e-2` lines can simply be dropped after this merges — the new defaults are equivalent or better.
- The figures are committed under `docs/src/assets/` (same pattern as #312's grid figure) and can back a future docs page on NTV tolerance settings.


logan-nc:
FYI @d-burg I had claude go another round with this now that a few other things merged. 
The auto grid looks pretty good to me. I am seeing the dpm regression values change outside the pinned range with changes in either the auto tau or the ldp mpsi though. 

I had claude working through a smarter auto grid. It tried denser packing near the rationals, but that failed. It was looking into at least just consistent packing around the rationals, but I ran out of tokens. That seems like a hack to force "robustness" anyways - why trust any given fixed packing? Do you have any better ideas for how to make a smart grid that gets the "right" answer for dpm?

logan-nc:
## Handoff — corrected understanding of the Δ′ grid work, and what's left

**Read this before touching the Δ′ / grid code.** The two commits below are already pushed, and the PR description above **overclaims**. This post corrects the record and lists the work remaining.

### State of the branch

| Commit | What |
|---|---|
| `0ff8008a` | Merge of `origin/develop` (brings the Ip=1.15 DIII-D equilibrium; conflicts in DIIID `gpec.toml`, `EquilibriumTypes.jl`, parallel-integration tests resolved to keep this branch's two-pass `auto` grid) |
| `2b27889e` | `rational_psi_ladder` — geometric knot ladder straddling each rational, fixed step `h0=1e-3` |

Working tree clean, nothing pending.

---

### Corrected understanding

**What is genuinely true.** The original bug was real: the two-pass `auto` grid's near-rational stencil *changed topology with τ* (`psi_accuracy`) — the SING_PACK floor was τ-scaled (`h_s = 0.06·τ^{1/3}`), knots were placed by an integer-`N` global equidistribution, and `merge_mandatory_nodes` snapped out the pinned rational's own neighbors. `dpm[1,1]` swung ~30% with τ. The ladder does fix that pathology, and it made the run 2.2× faster than develop's ldp-256.

**What is NOT true — the PR overclaims.** Two defects, both confirmed:

**1. `h0 = 1e-3` is a magic constant, not "native resolution."**
The code comment (`src/ForceFreeStates/Sing.jl` const block) and the PR text claim `h0` is "matched to the equilibrium's native radial resolution." **That is false.** EFIT `nw` is reachable via `equil.ingest.sq_xs` (`EquilibriumTypes.jl:355`, InverseIngest `:391`); native `Δψ_N = 1/(nw−1)`. For the DIII-D g-file (nw=257) that is **3.9e-3** — so `h0=1e-3` is **~4× *finer* than native**, tuned empirically on one equilibrium.

Exposure: **`examples/efit_fixedbdy_separatrix_example/eq_eps0.0500000_k1.000_d0.000.geqdsk` is nw=129** (Δψ ≈ 7.8e-3), where `h0=1e-3` is ~8× finer than native. **Untested.** Analytic equilibria (Solovev/LAR) carry `ingest === nothing` (`EquilibriumTypes.jl:402-404`), so any Δψ-derived `h0` needs a fallback.

**2. The τ-"convergence" is largely circular — Δ′ is not converged.**
The ladder pins the near-rational stencil *independent of τ by construction*, so τ-invariance of Δ′ is substantially **guaranteed, not earned** — τ no longer controls the thing that determines Δ′. The real test is refining the **near-rational** resolution `h0`, and that does **not** converge:

```
h0 = 2e-3 → dpm11 = 8.65
h0 = 1e-3 → 8.40      ← the shipped value
h0 = 5e-4 → 8.49
h0 = 2e-4 → 8.77
h0 = 5e-5 → 149       ← rings / blows up
```

Non-monotone, then divergent. `dpm ≈ 8.4` is a **plateau / sweet spot, not a limit**.

**Honest characterization:** `dpm[1,1] = 8.4 ± ~5%`, independently corroborated by ldp `mpsi=512` (8.48); ldp `mpsi=256` (7.28) is under-resolved. It is **not** a converged value, and the PR should not say it is.

### Root cause — the Δ′ extraction is unconditioned (this is the real problem)

- `sing_get_ua` (`Sing.jl:821-848`): `pfac = dpsi^alpha` (`:840`); big column `./= pfac` → carries `dpsi^{−α}` (`:841`), small column `.*= pfac` → `dpsi^{+α}` (`:842`). Faithful port of Fortran `dcon/sing.f:782,795,796`.
- `sing_get_ca` (`Sing.jl:1118-1139`): builds that basis as `temp1` and does a **plain `temp2 .= lu!(temp1) \ temp2` at `Sing.jl:1131` — with no equilibration.** `cond(temp1) ~ dpsi^{−2α}`.

That amplification is what turns equilibrium-spline/grid noise near a rational into a spurious Δ′ — and it is **why packing finer rings instead of converging**. Fortran does the same raw solve (`sing.f:837-841`, `zgetrf`/`zgetrs`, no equilibration), so this is a faithful port of a numerically fragile step, not a Julia-introduced bug. The grid ladder treats the *symptom*; this is the *cause*.

### The likely true fix (not yet attempted)

**Column-equilibrate `temp1` inside `sing_get_ca`**: scale its columns by their norms (diagonal `D`), solve `lu!(temp1*D) \ temp2`, rescale the result rows by `D`. Mathematically exact (equilibrate/un-equilibrate cancels).

- **~6 lines, entirely inside `sing_get_ca` (`Sing.jl:1118-1139`).** No signature change, no caller change, no ripple into `ca_l`/`ca_r` (`Riccati.jl:1134`, `:1175`) or the PEST3 combine (`Riccati.jl:725`).
- Expected: removes the `dpsi^{−2α}` blow-up → Δ′ should **converge** as `h0` refines instead of ringing → makes `h0` far less critical and would let us claim real convergence.
- **Caveat:** it will *not* fix the **imaginary** parts. Those come from a separate near-cancellation — the PEST3 four-term subtraction (`Riccati.jl:725`), where `dp_raw` entries are 1e4–1e5× the result (`ForceFreeStatesStructs.jl:247`), already mitigated by `Double64` (`Riccati.jl:692`).

### Ground truth is obtainable

Fortran **STRIDE emits the Δ′ matrix for the same PEST3 combination**: `ode_calc_delta_prime` (`~/Code/gpec/stride/ode.F:1045`), written to `delta_prime.out` (`ode.F:1240-1250`) and NetCDF var `Delta_prime` (`stride/stride_netcdf.f:223-226`). Running STRIDE on `examples/DIIID-like_ideal_example` gives a **direct comparand** and settles whether 8.4, 7.28, or something else is right. This should anchor everything below.

---

## TODO

**1. Correct the overclaims (do first — these are live inaccuracies in pushed code).**
- `src/ForceFreeStates/Sing.jl`: const-block comment + `rational_psi_ladder` docstring — drop "matches the equilibrium's native radial resolution" (it is ~4× finer than nw=257's 3.9e-3). State the truth: an empirically chosen step in a stable band, above the ringing threshold (~1e-4) and below native.
- `test/runtests_parallel_integration.jl` (~`:545-557`, the dpm pin comment): reword "grid-robust" → τ-desensitized *by construction*; values corroborated by ldp-512 but **not** a converged limit (±~5%).
- PR #313 "Update (2026-07)" section: same correction — do not claim convergence.

**2. Frobenius conditioning fix — the real fix.**
- Implement column equilibration at the LU in `sing_get_ca` (`Sing.jl:1131`), self-contained variant.
- Re-run the h0 sweep. **Acceptance: `dpm[1,1]` converges as `h0 → 1e-4…5e-5` instead of ringing to 149.**
- Physics-review it (`fortran-physics-reviewer`): it deviates from the Fortran raw LU, so justify it as a numerically-equivalent-but-better-conditioned solve.

**3. Re-evaluate whether the ladder is still needed.**
- If (2) makes Δ′ converge, the ladder may be doing far less than claimed. **Test: does the conditioning fix *alone* (ladder reverted to a single mandatory rational knot) remove the original 30% τ-swing?**
- If yes → the ladder is redundant complexity; revert or greatly simplify it.
- If no → keep it, and do (4).

**4. Make `h0` resolution-adaptive (only if the ladder survives (3)).**
- `h0 = c · Δψ_in`, `Δψ_in = 1/(length(equil.ingest.sq_xs) − 1)` — reachable at the two-pass call site (`src/GeneralizedPerturbedEquilibrium.jl:238`), passed via `rational_psi_ladder`'s existing `h0` kwarg.
- Fallback required when `equil.ingest === nothing` (analytic Solovev/LAR).
- Choose `c` deliberately (c≈0.25 reproduces today's DIIID `h0=1e-3`). **Unresolved:** whether the right anchor is the 1D profile grid (`sq_xs`) or the 2D `psi_rz` map — confirm which bounds `fieldline_int` (`DirectEquilibrium.jl:494`), since the re-form resamples a cubic spline of the raw nw-point arrays rather than re-solving Grad–Shafranov.

**5. Validate against ground truth + a second resolution.**
- Run Fortran STRIDE on the DIII-D example; compare `Delta_prime` to the Julia dpm diagonal.
- Run the pipeline on the **nw=129** equilibrium (`examples/efit_fixedbdy_separatrix_example`) — checks the over-optimization exposure. Currently untested.

**6. Regression coverage (deferred).**
- `fortran-physics-reviewer` flagged that a multi-n / low-shear case would exercise the overlapping-ladder regime end-to-end. The grid min-spacing invariant *is* unit-tested (`test/runtests_grid_refinement.jl`, Solovev n=1..2), but full-pipeline Δ′ in that regime is not.

---

### Reference numbers (DIII-D, mpert=27, the `delta_prime_matrix` testset path)

| grid | dpm11 (q=2) | dpm22 (q=3) | dpm33 (q=4) | et[1] |
|---|---|---|---|---|
| ladder h0=1e-3, τ=1e-3 (shipped) | 8.402503 | −5.802715 | −15.97267 | 0.808 |
| ldp mpsi=256 (develop) | 7.280089 | −5.211187 | −15.81737 | 0.8006 |
| ldp mpsi=512 | 8.477646 | −5.539578 | −16.03544 | 0.8030 |

`et[1]` is grid-robust to <1% everywhere — it is a whole-plasma integral of the *dominant* solution, so it never touches the `dpsi^{−2α}` cancellation. That contrast is the cleanest evidence the problem is local singular-layer extraction, not the global response.

**Scan script** (not committed; recreate under a scratchpad): replicate the `delta_prime_matrix` testset setup (`test/runtests_parallel_integration.jl:485-560`) parameterized by `grid_type`/`mpsi`/`psi_accuracy`/`h0` → `FF.rational_psi_ladder(...; h0)` → `EQ.refined_psi_grid` → `setup_equilibrium(...; override_psi_nodes)` → `eulerlagrange_integration` → `free_run!` → `compute_delta_prime_matrix!`; return `et[1]` and `real(dpm[i,i])`.

**Harness baseline** (`regress --cases diiid_n1 --refs origin/develop,local`, last run): equilibrium, energies, NTV torque, ‖resonant flux‖ all <0.5%; delta-prime norm moved 36% (intended); runtime 158 s vs 353 s.

logan-nc:
## Correction + ladder removed (supersedes the "Update (2026-07)" section and commit `2b27889e`)

Empirical re-measurement on the current branch head (post-`develop`-merge) does **not** support the "grid-robust Δ′ ladder" claims, so `rational_psi_ladder` has been removed (commits `d16c7c7e`, follow-ups `9a8dc3e0`/`c06d9464`). The measurements, on the DIII-D case (real `dpm` diagonal, `delta_prime_matrix` testset path):

**h0 scan (near-rational resolution, τ=1e-3):**
```
h0:    2e-3   1e-3   5e-4   2e-4   1e-4   5e-5
dpm11: 8.59   8.34   8.34   8.45   8.49   8.76      → ±3%, NO ringing
```
The reported "h0=5e-5 → 149" does **not** reproduce.

**τ (psi_accuracy) scan:**
```
τ:            2e-3    1e-3    5e-4    2.5e-4     spread
single_knot:  7.88    6.19    7.92    9.91       ~48%
ladder:      10.06    8.34   10.26    9.97       ~21%   (mean ~9.7, ABOVE ldp-512 = 8.48)
```
The ladder reduces the swing only to ~21% (not ~4%) and biases dpm11 *above* the ldp-512 reference. The pinned 8.40 was a coincidental sweet spot at the default τ.

**Conditioning fix is a no-op.** Column-equilibrating the `sing_get_ca` LU solve gives bit-identical results — the solve is not the source of the sensitivity.

**Conclusion.** The ideal-MHD Δ′ diagonal is intrinsically grid-uncertain (~±30–50% across the accuracy knob); `et[1]`, NTV torque, and ‖resonant flux‖ stay grid-robust to <1%, so the sensitivity is local singular-layer matching, not the global response. Δ′ is now treated as a documented rough diagnostic (single mandatory rational knot + the general measured-curvature grid), pending the resistive-layer work. Net: **−~150 lines, −4 one-example-tuned constants.**

### What changed
- Deleted `rational_psi_ladder` + `SING_LADDER_*` constants; driver pins bare rational surfaces (`rational_psi_nodes`).
- Re-pinned the `delta_prime_matrix` regression diagonals honestly (grid-sensitive diagnostic, not converged).
- Minor: named the ψ-panel merge tolerance (`PANEL_MERGE_ATOL`); dropped the interim `atol_psi` override from the DIII-D example (this PR *is* that follow-up; numerically inert since rtol already bound termination).

### Regression (working tree vs `develop`)
`solovev_n1`: unchanged. `diiid_n1`: globals/energies/**NTV torque (0.67%)** preserved; the grid-sensitive Δ′ diagnostics move **intentionally** — delta prime ~27%, Chirikov ~9.5%, island ~4.6%, ‖resonant field‖ ~7.8%; runtime faster (298 vs 341 s). These `CHANGED` rows are expected, not a regression.

### Generality
With the ladder (and its one-equilibrium-tuned `h0`) gone, the auto grid runs cleanly with zero equilibrium-specific tuning on analytic Solovev, nw=257 DIII-D, and the previously-untested nw=129 EFIT.

### One item worth a physics glance
The regression shows `PE toroidal torque` sign-flipped at small magnitude (−0.051 → +0.051) under the grid change — flagging in case it matters.

logan-nc:
Consolidated into #312. This PR was built on `feature/two-pass-psi-grid`, and rather than maintain two entangled PRs, the full package is now on that branch: `feature/two-pass-psi-grid` has been fast-forwarded to include all of this branch's KineticForces NTV ψ-quadrature paneling, the Δ′ knot-ladder removal (see the [correction above](https://github.com/OpenFUSIONToolkit/GPEC/pull/313#issuecomment-5079015877)), and the merged `develop`.

#312 now carries the combined grid + NTV package and will be cleaned up (PR description reconciled, full `test/runtests.jl` + complete regression harness) before merging to `develop`. Closing this in favor of #312.

Follow-up issue #333 (PE toroidal-torque Hermiticity check) remains open and is independent of the merge.

PR #325, opened 2026-07-10
PerturbedEquilibrium & KineticForces: parallelize serial workflow loops (coupling, NTV torque) + reduce allocations

logan-nc:
# PerturbedEquilibrium & KineticForces: parallelize serial workflow loops (+ allocation cleanup)

Three performance improvements to the perturbed-equilibrium / kinetic-diagnostic workflow on
the DIII-D-like reference case. Each is verified to leave results unchanged (bit-identical or
within regression tolerance) and independent of thread count.

## 1. Parallelize the NTV-torque bounce-harmonic loop (`KineticForces`)
The NTV-torque outer ψ-integral (QuadGK `BatchIntegrand`) stays serial — its small per-batch
node counts make threading the ψ loop fork-join-bound — but the **inner bounce-harmonic loop**
(`2·nl+1` harmonics per ψ node) carries the ~40 ms of per-node work and is embarrassingly
parallel. Threaded it: each harmonic's `tpsi!` writes per-`intr` scratch buffers/hints, so
each thread gets its own `deepcopy(intr)` (the pattern the self-consistent matrix build uses),
and per-harmonic results are reduced in fixed harmonic order.
- **fgar NTV torque: 35.3 s → 8.2 s at 8 threads (4.3×).**
- **Bit-identical**: `82.1808181112855 − 14.6098675986818i` to every digit across the serial
  baseline, threaded t=1, and threaded t=8.

## 2. Parallelize the singular-coupling loop over rational surfaces (`PerturbedEquilibrium`)
Each rational surface builds its own vacuum Green's functions (allocation-local
`compute_vacuum_response`) and writes disjoint coupling rows / `sing[s]`, so the loop is now
`@threads :static` over resonant pairs (BLAS pinned to 1 across it).
- **Singular coupling: ~0.54 s → ~0.18 s at 8 threads (~3×)**; results identical across thread
  counts and equal to serial to ~13–15 significant figures (BLAS-pin reordering).

## 3. Reduce allocations in field reconstruction (`PerturbedEquilibrium`)
`reconstruct_physical_fields`: in-place mode DFTs, `@view` eigenmode slices, `cholesky!`, BLAS
pin. **~30% fewer allocations (594 → 420 MB), bit-identical.** (A memory cleanup — reconstruction
is bandwidth-bound, not GC-bound.)

## Verification
Regression fingerprints of every affected output (`b_n`/`xi_n`, `C_resonant_*`, `resonant_*`,
and the NTV torque) match the serial baseline and are thread-count-independent; the repo's
coordinate-invariance, coils, full-run, and HDF5-replay suites pass.


logan-nc:
@claude review this pr

logan-nc:
Addressed the review findings in 424463c6 (threading-correctness pass):

- **Removed the `sing[s].grri/.grre` data race** — these were write-only (never read anywhere in `src/` or `test/`), and a surface `s` legitimately repeats across rows in multi-n runs (integer q resonates at several n), so the shared-field write was a genuine race. Dropped the dead store; behavior-neutral.
- **Guarded both BLAS pins with `try/finally`** (`SingularCoupling`, `FieldReconstruction`) so an exception in a threaded loop can't leak `BLAS.set_num_threads(1)` into the rest of the session.
- **Documented** the top-level-`@threads` requirement of the `threadid()`-indexed buffers and the `nthreads × sizeof(intr)` memory scaling of the per-thread `deepcopy`.

Verified thread-safety of the remaining shared reads: the vacuum solve uses a **task-local** pool (`@with_pool` → `get_task_local_pool`), and FastInterpolations' `hint` is an external caller kwarg (never internal mutable state), so concurrent hint-less spline calls are race-free.

**Regression:** `regress --cases diiid_n1,solovev_kinetic_calculated --refs develop,local` is **bit-identical** to develop (all quantities `0.0e+00`), covering the singular-coupling, field-reconstruction, and NTV-torque paths.

Follow-up (separate PR): a **multi-n perturbed-equilibrium regression case** to exercise the repeated-`s` path and a threaded CI job — tracked in a new issue.

Issue #93, opened 2025-12-10
Large non-hermitian component of plasma response matrix when calculating crit


jhalpern30:
As discussed in #91, when calculating crit (the smallest eigenvalue of Wp), the original Fortran code forces Hermitian symmetry of the matrix. According to the DCON paper, this matrix should already be Hermitian by construction, so this was confusing. As part of this PR, we added in a warning if the non-Hermitian part of this matrix becomes too large, which I arbitrarily set at 1e-3; however, in some of the basic examples such as the DIIID ideal example, this warning is being raised. It seems like if this warning if being thrown on the simplest of examples, it likely should not be a warning and just be fixed or removed.

I think ultimately, we need to figure out if there being a large non-Hermitian component is a bug that we need to fix and are masking by forcing the Hermitian symmetry or if it not affecting the results and the warning should just be removed.

logan-nc:
Note that hidden in #24 is an update that cleans up the non-hermitian warnings such that it only prints one waring at the end of the integration telling you how many steps were non-Hermitian (no warning is printed if none were).

jhalpern30:
Progress update: I have been doing some investigating on this and will be adding some updates in an upcoming PR. I think this is an inherent issue in the serial EL integrator, where condition number of the solution matrix is becoming absolutely massive (and which isn't fixed by the fixups). After iterating with Claude a bit, it seems like the Ricatti path completely fixes this issue - since this seems to be what we are trending toward (and DCON has used this implementation for decades without too much problem), I think its ok to flag the bad conditioning when it appears (in the same place as the current non-Hermitian error). Unfortunately, I wasn't able to track down a full fix for this since I think its quite a complex issue.

## To provide a bit more context from some of Claude's outputs that might help in the future:
### Why is Wp Hermitian?
`Wₚ⁻¹ = U₁·U₂⁻¹` is built from the Euler-Lagrange fundamental solution matrix `u` (blocks
`U₁ = u[:,:,1]`, `U₂ = u[:,:,2]`). Glasser (2016) — `docs/resources/2016-Glasser-...pdf`,
Eqs. 26/28/55/57/63/66 — shows `Wₚ` must be Hermitian because of the symplectic invariant
`U†JU = J` preserved by the Euler-Lagrange ODE. Any measured anti-Hermitian part is pure
numerical noise, not physics. Physically, Wp is Hermitian because delta W_p is real.

### Why isn't it actually Hermitian?
[...]

PR #330, opened 2026-07-16
VACUUM - To be (Hermitian) or not to be (deprecating force_wv_symmetry)

jhalpern30:
Brought about by my investigation for the plasma response matrix in #93, I realized that we can deprecate the `force_wv_symmetry` parameter. 

The response matrices are inherently Hermitian - they compute real quantities (energies), so they must have all real eigenvalues and the formulation of the code expects this. However, numerical errors allow for non-Hermitian components to appear. In ForceFreeStates, this is highly nontrivial (again, see #93), but in Vacuum, its must simpler. The non-Hermitian error is only due to the error induced from discretization and can be easily decreased by increasing the number of grid points: 

image

In general, this is much less than the non-Hermitian errors seen in the plasma response matrix. I chose the value of 1e-4 relatively arbitrarily to be just a bit tighter than the arbitrary one for the plasma response matrix tolerance we've been using. 

The `force_wv_symmetry` parameter (`lsymz`) in the Fortran, allowed a flag to turn this on/off in the vacuum code. However, for the above reasons, this was pretty much always left as true. Because of this and my above arguments, I have deprecated this "always-on" parameter in favor or always symmetrizing the matrix, and including a simple warning statement for large Hermitian errors, requesting the user to increase the grid size (this will go hand in hand with just increasing the accuracy needed to converge the boundary integral too), so I doubt users will ever really run into this, but just in case.

Since this parameter defaulted to true, this doesn't change the outputs. However, I've added it to the list of deprecated parameters for backwards compatibility.

As an example of usage, I ran DIIID ideal as normal (and nothing changed), and then dropped `mthvac` down to an unrealistic 32, and the warnings pop up like this
```julia
[ Info:    ψ = 0.995,  q= 6.188,  max(S) = 8.60e+00,  steps = 51
[ Info:    ψ = 0.995,  q= 6.200,  max(S) = 8.19e+00,  steps = 52
┌ Warning: Vacuum energy matrix Wᵛ is non-Hermitian above tolerance 0.00948512437201021 > 0.0001 before symmetrization. Increase vacuum grid resolution to reduce it.
└ @ GeneralizedPerturbedEquilibrium.Vacuum ~/Github/JPEC/src/Vacuum/Vacuum.jl:48
```


jhalpern30:
@logan-nc relatively quick one here motivated by background reading I was doing to prep for the perturbed eq updated

logan-nc (approved):
Looks good!

PR #336, opened 2026-07-29
ForceFreeStates - BUG FIX - Pass direction arg to IntegrationChunk in kinetic EL crossing

logan-nc:
## Summary

`cross_kinetic_singular_surf!` constructed its placeholder `IntegrationChunk` with a stale **4-argument** call:

```julia
IntegrationChunk(0.0, 0.0, false, 0)          # (psi_start, psi_end, needs_crossing, ising)
```

but `IntegrationChunk` is an `@kwdef struct` whose `direction` field (added later for bidirectional parallel FM) makes the positional constructor require **5** arguments. The `ising=0`/`direction=1` defaults apply only to the *keyword* constructor, not the positional one, so the 4-arg call throws:

```
MethodError: no method matching IntegrationChunk(::Float64, ::Float64, ::Bool, ::Int64)
```

The fix passes `direction=1`, matching the two sibling placeholder call sites (`cross_ideal_singular_surf!` and the Riccati crossing), which already use the 5-arg form.

## Why it was latent

The line is only reached on the **serial Euler–Lagrange shooting** solver (`use_parallel=false`) when a **kinetic singular surface** is present (`kinetic_factor>0`, self-consistent Mode A). Until now, self-consistent kinetic runs always routed through the parallel-FM/Riccati path, which bails out *earlier* with `"kinetic_factor > 0 not implemented yet in Riccati"`. So this call site was never executed and the defect stayed hidden.

## Verification

- Empirically reproduced the `MethodError` for the 4-arg call and confirmed the 5-arg call succeeds.
- With the fix, a self-consistent kinetic run (`use_parallel=false`, `kinetic_factor=1`) finds and crosses the kinetic singular surface and completes end-to-end with converged output.
- Regression harness (`diiid_n1`, `solovev_n1`): no attributable change — the touched path is unreachable in these ideal/perturbative cases (confirmed by isolation; the only residual scatter is pre-existing threaded-reduction non-determinism in near-singular PE diagnostics).

https://claude.ai/code/session_01SsLUydP2gJbE1tGoaHDiS1

### DIFF
--- a/src/ForceFreeStates/EulerLagrange.jl
+++ b/src/ForceFreeStates/EulerLagrange.jl
@@ -676,7 +676,7 @@ function cross_kinetic_singular_surf!(
     ksurf = intr.kinsing[ising]
     dpsi = ksurf.psifac - odet.psifac
 
-    params = (ctrl, equil, ffit, intr, odet, IntegrationChunk(0.0, 0.0, false, 0))
+    params = (ctrl, equil, ffit, intr, odet, IntegrationChunk(0.0, 0.0, false, ising, 1))
     du1 = zeros(ComplexF64, intr.numpert_total, intr.numpert_total, 2)
     du2 = zeros(ComplexF64, intr.numpert_total, intr.numpert_total, 2)
 


logan-nc:
@claude review this

logan-nc:
Addressed the review:

- **`ising` consistency (point 1):** amended to `IntegrationChunk(0.0, 0.0, false, ising, 1)`, passing the in-scope `ising` to match the sibling call sites. Still functionally inert (the chunk is discarded in `sing_der!`), but consistent.
- **regression-guardian (point 3):** run against current `develop` (not a stale baseline) — `diiid_n1`, which is the case that exercises the kinetic FGAR quantities. **No attributable change:** `NTV torque FGAR` and `dW FGAR` are OK, and the rows that show a delta (delta prime ~4.5%, resonant field ~4.6–6.6%, island widths ~2.5%) are the known threaded-reduction non-determinism in near-singular PE diagnostics — identical to the develop-vs-develop baseline, not caused by this change. (The ideal `diiid_n1` path never reaches `cross_kinetic_singular_surf!`, so the fix can't affect it.)
- **Test coverage (point 2):** deferring — this is a runtime `MethodError` on a slow self-consistent path, so a tracked regression case would be noise rather than signal. Noted for a future lightweight guard if the placeholder call sites are revisited.

PR #342, opened 2026-07-30
GPEC - DEPRECATION - deprecating mat_flag, ode_flag, cyl_flag

jhalpern30:
Deprecating some flags that have stuck around for a while and should be removed before cleanups start for GPEC 2.0.

- `ode_flag`: a trivial flag that ran the ODE. I can't think of a reason why you'd ever want to not run it
- `cyl_flag`: centered the m grid around 0, I don't think anyone used this
- `mat_flag`: this was determining whether or not to dump the matrices to HDF5. While I think this is potentially useful, it shouldn't be in control and should be in some sort of DEBUG namelist once we get that up and running

I added all deprecated parameters to the deprecated keys list.

Note that for some of the benchmarks, the formatter wasn't run on them so the diff is massive - all I did was remove ode/mat_flags.


jhalpern30:
@claude review

jhalpern30:
@claude review

matt-pharr (approved):
This is overall good cleanup and I like getting rid of fortran flags. However, I should note just so you know @jhalpern30  that there exist a number of reasons someone might want to not run the ode integrator, like if they only want ballooning analysis, for example. We have a big suite with many capabilities now and not all of them require ode, so yeah.

There is also the matter of mat_flag being always on now, which as claude mentioned, will be cumbersome for big scans creating massive h5 files.

However, I think that both of these concerns are for now moot, because I want to restructure the code to work better in scripting form, e.g.

```
eq = Equilibrium("input.geqdsk")
basis, stability = ForceFreeStates(eq, :riccati, :ideal) <-- could also calculate plasma permittivity
rmp = RMPField("coils.dat", amplitude={1: 1000}) # set n=1 amplitude to 1kA
# or, rmp = RMPField(surfmn_output) / etc

perturbed = PerturbedEquilibrium(basis, rmp)
outputs = calculate_quantities(perturbed, ["C_xe", "delta_per_coil", "b_pen"])
```

When we have something like this calculate_quantities() function, these flags will be irrelevant and we will want to get rid of all of them anyways, so this is a good step towards that.

jhalpern30:
@matt-pharr agreed - perhaps the better wording instead of "fully deprecate" is "remove vestigial Fortran debug flags to make room for our own implementations later on"

PR #343, opened 2026-07-30
Reducing CLAUDE.md to a managable size

jhalpern30:
From conversations with @priyanshlunia and @matt-pharr - our CLAUDE.md file has gotten to be absolutely massive (takes up 45k. This PR massively reduces its size (from 750 lines down to ~150) by moving a lot of that context to separate markdowns in `docs/development`, with pointers in CLAUDE.md to look there when needed. This should reduce token usage and also adhere's to what JD has been mentioning that newer models require less context to effectively code

I did a quick check to remove obviously stale references in the files as well, but I did not update it for anything new. 

Matt and Priyansh, just looking for someone who has more experience with CLAUDE.md management to take a quick look at this to see if it looks reasonable/actually offers improvement


matt-pharr (approved):
This looks good! Thanks for doing this @jhalpern30

PR #344, opened 2026-07-30
CI - IMPROVEMENT - Gate the Julia test suite on Julia changes, and general workflow cleanup

matt-pharr:
Only changed github tests. Changes: 

1. Make it so that the julia tests only run on PRs that change julia code in src/ and tests/
2. Added formatter test. Currently disabled, plan is to enable it once we reformat the code in the repository and start enforcing formatting standards.
3. Made superseded tests cancel, in case someone pushes two commits back to back, so we don't waste monthly limited container time
4. Added reasonable timeouts to all the jobs
5. Updated to latest checkout and julia test workflow packages


jhalpern30 (approved):
These all look like great changes - especially the auto-formatter check, I've had too many PRs where I change one line on my end but the entire file gets diffed

logan-nc:
❤️

Issue #349, opened 2026-08-01
Auto psi-grid: accuracy-driven density is grid-dependent — pursue a self-consistent / one-pass construction


logan-nc:
# Auto psi-grid: the accuracy-driven density is grid-dependent — pursue a self-consistent / one-pass construction

## Summary

The two-pass auto grid (`grid_type = "auto"`, `src/Equilibrium/GridRefinement.jl`) sizes knot
density from an *accuracy requirement*: place knots so the cubic-spline derivative error of the
profile/geometry channels meets `psi_accuracy`. The premise is that this density is a property of
the **equilibrium**. It is not — the measured curvature that drives the density **grows with the
grid it is measured on**, so the grid the generator produces depends on the pass-1 grid rather than
converging to a unique accuracy-defined grid.

This does not affect the ideal Δ′ convergence (fixed separately by rational-surface bracketing —
see the accompanying PR), but it undermines the auto grid's stated guarantee and couples pass-1
resolution to pass-2 grid size (hence ODE cost).

## Evidence

Forming the equilibrium on uniform grids of increasing resolution and evaluating
`implied_knot_count` (the density integral ∫ρ) at a **fixed** `tau` on each (DIII-D-like example):

| mpsi | implied(τ=1e‑3) | implied(τ=1e‑4) |
|------|-----------------|-----------------|
| 128  | 418 | 900 |
| 256  | 436 | 939 |
| 512  | 709 | 1508 |
| 1024 | 1551 | 3310 |
| 2048 | 2643 | 5649 |

A grid-invariant density would saturate; instead ∫ρ grows ~monotonically with the measurement
grid. The same effect makes `implied/used ≈ 2` (and climbing) in two-pass runs — long read as
"underpacking," it is actually this non-convergence of the measurement.

## Root cause

Near the separatrix the flux geometry is genuinely near-singular (q → ∞ ~ −A·ln(1−ψ_N); the r²,
jacobian and ν channels compress at the X-point). Each finer equilibrium solve resolves sharper
edge curvature, so the divided-difference `|f''''|` estimate diverges as resolution increases. Two
regularizations were tried and **both fail** (documented so they are not re-attempted):

- **Fixed-ψ-width stencil** for `_fourth_derivative_nodes` (anchor the 4th-difference to a fixed
  width instead of the 5 adjacent nodes): ∫ρ still grew 418→2643. The dominant edge growth is the
  2D geometry channels, whose singularity a wider window cannot bound.
[...]

PR #238, opened 2026-05-18
Feature: add SLAYER and GGJ tearing growth rates

d-burg:
## Summary

Adds tearing-mode growth-rate analysis as a new `Tearing` umbrella module under `src/Tearing/`, with three layers:

- **InnerLayer** — inner-layer Δ(Q) physics: GGJ (Pletzer–Dewar) and SLAYER (Fitzpatrick Riccati) models, with switchable Spitzer/neoclassical resistivity.
- **Dispersion** — physics-agnostic root finder: `SurfaceCoupling`, `MultiSurfaceCoupling`, `CoupledFull` (2m×2m det(D′−D(γ))), and `CoupledFortranMatch` residuals; AMR Q-plane scan + triangulation-based growth-rate extraction; spurious-root detection (geom + γ-gap + polyline concavity)
- **Runner** — TOML-driven orchestration (`Tearing.Runner.run_slayer`), kinetic-profile loading, HDF5 output.

Also includes the supporting work that landed alongside:
- Full 2m×2m D′ matrix exposed from `ForceFreeStates` via `delta_prime_raw` / `pest3_decompose`
- New `Riccati.jl` solver in `ForceFreeStates` for the ideal plasma response (with ~30–40% perf work)
- `Utilities`: `NeoclassicalResistivity`, `KineticProfiles`, `PhysicalConstants`
- Regression cases: `solovev_slayer_n1`, `tj_epsilon_pole`
- LAR β-scan / ε-scan and TJ_epsilon_pole examples

73 commits, ~+13.9k / −0.75k LOC across 92 files.

## Relationship to `perf/riccati`

The inner-layer growth-rate solvers here consume the outer-region Δ′ produced by the Riccati-based ideal-MHD solver on [`perf/riccati`](../tree/perf/riccati). This branch is designed to subsume `perf/riccati` once that work merges: it currently shares a common ancestor (`c6c845ff`) and is 39 commits ahead but 11 commits behind `perf/riccati`. Plan is to merge the final state of `perf/riccati` in here once it's ready to merge and before this PR moves out of draft, so the final history includes the outer-region Δ′ improvements that the dispersion residuals depend on.

## Current status — WIP / draft

- Core stack (InnerLayer + Dispersion + Runner) and regression cases pass on Solovev and TJ-like analytical equilibria.
[...]

PR #350, opened 2026-08-01
EQUIL - BUGFIX - Fix ideal Δ′ non-convergence on the auto psi-grid

logan-nc:
## What & why

The two-pass auto psi-grid (`grid_type = "auto"`) produced a **non-convergent ideal Δ′**: as
`psi_accuracy` tightened, the Δ′ diagonal oscillated erratically (dpm[1,1] ≈ 7.2–10.1 on the
DIII-D-like case) instead of converging to the fixed-grid reference.

**Root cause.** The grid pinned a knot *on* each rational surface. The Δ′ asymptotic matching
stencil `[ψ_s ± singfac_min/|n·q′|]` then straddled that knot, so the cubic 3rd derivative the
singular extraction samples was inconsistent across the crossing. Confirmed by a clean in-run
natural experiment: surfaces whose knot happened to fall *outside* the stencil (e.g. q=4) were
rock-stable, while those straddling it (q=2/3/6) were erratic.

## The fix (grid layer only)

- **`bracket_mandatory_nodes`** — center each rational *between* a knot pair (never on a knot),
  blended to half the local spacing, so the matching stencil lies in one clean, locally-uniform
  spline segment and q‴ is single-valued across it.
- **Fixed-width (τ-independent) near-rational resolution patch** — Δ′ accuracy is a property of
  the rational surface, so its local resolution must **not** scale with the global accuracy target
  `psi_accuracy` (that only sizes the pedestal/edge grid). This is what makes the default
  `psi_accuracy = 1e-3` land on the converged value.
- **`enforce_min_spacing`** floor on the auto refined grid and the fixed `ldp`/`pow1` grids, so no
  grid packs its edge below the field-line-integration noise scale at high mpsi.

## Result

dpm[1,1] now converges **τ-independently to ~9.2** (matching the uniform mpsi ladder), default
`psi_accuracy=1e-3` included, vs the prior erratic 7.2–10.1. Equilibrium scalars are unchanged
(q95, β_n = 0.00%) — this is grid placement only.

| | develop (erratic) | this PR |
|---|---|---|
| dpm[1,1] @ τ=1e‑3 | 9.05 (lucky) | **9.199** (converged) |
| dpm[1,1] across τ | 7.2–10.1 | **9.2–9.3 (τ-independent)** |

## Regression (`diiid_n1`, develop vs this branch)

Equilibrium scalars unchanged. Δ′-dependent quantities changed **as intended**: Δ′ diagonal
(dpm[1,1] 9.05 → 9.199), island half-widths (7.5%), Chirikov (21%), resonant field (14%). Runtime
+13% (within budget). **These changes need reviewer sign-off** — the harness is a live develop-vs-local
diff, so `develop` becomes the new baseline on merge.

## Scope / follow-ups

- The accuracy-driven density remains **grid-dependent near the separatrix** (a real q/geometry
  singularity; the default converges to the *uniform* reference, ~3% above the *ldp* reference).
  That is out of scope here and tracked in #349 (self-consistent / one-pass grid redesign).
- **Not yet done:** a `kinetic_factor > 0` spot-check (the ideal deck has `kinetic_factor = 0`,
  and the near-rational knots' rationale is the kinetic-MHD path).
- Orthogonal grid change — cleanly **cherry-pickable onto `feature/tearing-growthrates` (#238)**
  if approved.

## ⚠️ Human review is mandatory

Do not merge without a third-party human reviewer's approval, and please confirm the regression
changes above are acceptable.

Closes/relates: #349 (deferred grid-invariance redesign)

https://claude.ai/code/session_0125dY2qbGtm3gky8TPv7vh6

---

## Scope update — this PR also restores the coil control surface to Fortran parity

The grid refinement here exposed a latent divergence from Fortran in the coil path, so this PR
is **no longer grid-layer only**. One behavioural change is included, declared here.

**`ForcingTerms/CoilFourier.jl`: the coil coupling surface now defaults to the outermost
computed surface (`psihigh`) instead of ψ_N = 1 — matching Fortran GPEC.**

### Fortran reference (PrincetonUniversity/GPEC v1.5.7-566-g902faac3)

Fortran evaluates the coil field at `psilim`, never at ψ_N = 1:

- `gpec/gpec.f:431` — `CALL field_bs_psi(psilim, coilmn(:,j), ...)`
- `coil/field.F:167-183` — `field_bs_psi` is the direct analog of `sample_boundary_grid`: same two
  splines, same square root, `rfac = SQRT(crzphi_f(1))` where `crzphi_f(1)` is `rzphi_rsquared`
  and `crzphi_f(2)` is `rzphi_offset`
- `coil/field.F:133` — the Fortran calls that mesh the **"control surface mesh"** in its own words
- `dcon/sing.f:170` — `psilim = psihigh`, moved only *inward* by `sas_flag`/`qhigh`/`psiedge`
- `equil/inverse.f:142` — `psihigh` is the last knot of the radial grid these splines are built
  on, so Fortran evaluates exactly ON the last knot and never extrapolates
- `docs/index.rst:60` — the external field is specified "on the surface of the GPEC plasma
  boundary **defined by the psihigh variable in equil.in**"

There is no ψ_N = 1 surface evaluation anywhere in the Fortran coil path, and no Taylor
extension to reach one. The ψ_N = 1 default was a port divergence.

Corroborating in-tree: **both** Fortran-comparison benchmarks already pass the correct surface
explicitly — `benchmark_against_fortran_run.jl:548` and
`benchmark_coil_ForcingTerms_against_fortran.jl:236`, both using Fortran's own `psilim`
attribute read from `dcon_output_n#.nc`. So the ψ_N = 1 default was never exercised on any
Fortran-validated path; only callers that omitted the keyword hit it.

### Why it surfaced now

Evaluating at ψ_N = 1 extrapolates a cubic past its last knot with no guarantee of remaining
physical. On the DIII-D-like example (psihigh = 0.995):

| ψ_N sampled | min `rzphi_rsquared` | θ points negative |
|---|---|---|
| 0.990 | +0.290 | 0/96 |
| **0.995** (= psihigh) | **+0.292** | **0/96** |
| 0.9975 | −1.84 | 1/96 |
| **1.000** (old default) | **−22.6** | **15/96** → `sqrt` DomainError |

The extrapolated distance is a fixed 1 − psihigh while the final spline interval shrinks with
edge packing, so the error grows with resolution. Bisection on `PASS1_INTERVALS` alone confirms:
`runtests_coils.jl` passes 8/8 at 128 and errors at 512 and 1024.

### Why fix it here rather than tune the grid

Δ′ is insensitive to `PASS1_INTERVALS` (dpm[1,1] = 9.246 / 9.291 / 9.239 at 256 / 512 / 1024)
while the coil test flips between 256 and 512. Tuning pass-1 down would mean choosing an
internal grid resolution to satisfy an unrelated extrapolation, and would silently walk back a
deliberate change in this PR. Fixing the root cause keeps `PASS1_INTERVALS = 1024` and leaves
the Δ′ convergence result untouched.

### Impact

The coupling surface moves inward by (1 − psihigh): **0.05 % of flux at the default
psihigh = 0.9995**, 0.5 % on the example, which lowers psihigh to 0.995 to capture q=6. Coil /
RMP amplitudes shift correspondingly and should be expected to differ slightly from previous
runs *that relied on the default* — runs passing `psi` explicitly, including both Fortran
benchmarks, are unaffected. Pass an explicit `psi` to recover the old behaviour.

### Both production paths use `psilim`

Strictly faithful is `psilim` (the *integration* limit), not `psihigh` (the *equilibrium spline*
limit); they differ whenever `dmlim`/`qhigh`/`psiedge` truncation fires — e.g. 0.998902 vs
0.9995 on a DIII-D case, a 6e-4 gap, the same order as the ψ_N = 1 extrapolation fixed above.
PPPL shipped a fix for exactly this confusion (`docs/releases.rst:281`: *"Fixes inappropriate
uses of psihigh, which may not be the end of integration psilim if sas_flag, qhigh, or
peak_flag are used"*).

The coil branch of `PerturbedEquilibrium` already passed `psi = ffs_intr.psilim`; the external
forcing-data branch did not, and so normalized on the wrong surface silently, with no error.
Both now pass `psilim`.

`psilim` is the ODE's last grid point to within `eps = 1e-10` (the final chunk ends at
`psilim*(1-eps)`, `EulerLagrange.jl:504`) and is snapped to `psi_store[end]` exactly in the
dW-peak truncation path (`Riccati.jl:1375,1870`, `EulerLagrange.jl:230`).

The `psihigh` default on the three signatures is retained and documented as the bare
`PlasmaEquilibrium` fallback — reachable only by direct callers with no ForceFreeStates solve,
for whom `psilim` does not yet exist — not as the intended control surface.

Full test suite passes locally.


logan-nc:
@d-burg it would be good to marge this by weds as awell

d-burg:
## Independent validation on FUSE-reconstructed DIII-D equilibria

Ran this branch on **12 equilibria from 6 DIII-D discharges** spanning β_N 0.3–3.4, q95 2.4–5.9 and 2–9 resonant surfaces (FUSE postdictive reconstructions; n=1, δm=8, dmlim=0.2, serial BVP). Three pinned builds so each effect is isolated: our production build, this branch's parent (f46b202f), and the parent + this PR. **The fix reproduces here, and the shipped defaults (`grid_type="auto"`, `mpsi=0`, `psi_accuracy=1e-3`) are exactly the configuration we'd recommend.**

### auto @ the default τ=1e-3 is converged

Error vs an ldp-1024 reference, 11 slices (one near-pole slice excluded — see below):

| grid | median \|err\| | max \|err\| |
|---|---|---|
| ldp-128 | 0.97 | 11.3 |
| ldp-256 | 0.22 | 7.5 |
| ldp-512 | 0.018 | 0.14 |
| **auto @ 1e-3** | **0.011** | **0.049** |
| auto @ 1e-2 | 0.036 | 0.44 |

τ-independent as claimed (1e-2 / 1e-3 / 1e-4 agree to ≈0.05, the sole exception being one small-\|Δ′\| case at the loosest τ).

### Placement, not count, is doing the work

Refined pass-2 node count actually selected:

| shot | nsing | τ=1e-2 | τ=1e-3 | τ=1e-4 |
|---|---|---|---|---|
| 97741 | 2 | 320 | 570 | 996\* |
| 148789 | 4 | 267 | 459 | 858 |
| 165024 | 8 | 277 | 479 | 900 |
| 161172 | 9 | 262 | 450 | 837 |

\* implied 1130, clipped by `REFINED_N_CAP`.

- **τ=1e-3 → ~450–580 nodes**: `auto` independently rediscovers the mpsi ≈ 512 our fixed-grid ladder shows is required.
- **At matched resolution auto beats ldp by 6–17×**: ~270 nodes gives 0.036 / 0.44 (auto @1e-2) against 0.22 / 7.5 (ldp-256). That is `bracket_mandatory_nodes` + the fixed-width near-rational patch earning their keep.
- **Density is curvature-driven, not rational-driven** — it slightly *anti*-scales with nsing (2 surfaces → 570 nodes, 9 surfaces → 450), only ~20 % spread across the set. Practically this is the biggest win: one τ works fleet-wide, no per-shot mpsi tuning.

### No regression for fixed-grid users

`enforce_min_spacing` leaves ldp Δ′ essentially untouched at mpsi=512 — median \|ΔΔ′\| = **0.0006**, max 0.029 across the set — at **+3 %** runtime, well inside your +13 % budget.

### One caution on the `uniform` reference

On these equilibria **`uniform` is not converged at mpsi ≤ 1024** and we would not use it as truth: uniform-512 vs uniform-1024 differ by up to 2.0, and uniform-1024 comes out **sign-flipped** relative to the converged ldp/auto value on 2 of the 6 discharges (−1.83 vs +0.94; −1.51 vs +0.42). ldp and auto agree with each other and are self-consistent under refinement; uniform lags, presumably for want of edge packing. Flagging it since the convergence argument here leans on the uniform ladder.

Related: on fixed grids convergence is a **damped oscillation, not a monotone plateau** — 7 of 11 slices *overshoot* at mpsi=256 before settling — so agreement between two coarse grids is not evidence of convergence. ldp-128 was bad enough to flip the sign of Δ′ on one discharge where the converged value is ≈ +0.9.

### Near-pole slices

One slice sits on a Δ′ pole (β_N 3.2 hybrid, 9 resonant surfaces). There the value spans ~900 across grid families and no grid family converges — ldp-512/1024 agree to 0.8 % at −738/−743 while the coarse grids return small positive values, i.e. refinement *converges onto* the divergence rather than removing it. Not a criticism of the PR — expected behaviour near a pole, and a reminder that single-slice Δ′ is not meaningful there.

Happy to share the scan driver or the raw CSVs if useful.

d-burg:
## Pushed: `develop` merged + the q=2 pin re-pinned on a converged ladder

Two changes on the branch, plus a correction to my earlier comment.

### 1. The failing pin was stale, and the ladder says your value is the right one

The old pin's own docstring flagged this: *"expect to re-pin whenever the grid generator changes (as here)"* and *"Replacing this with a converged-Δ′ pin on a fixed ldp grid is tracked as an open issue."* I ran that ladder on the fixture itself (`DIIID-like_ideal_example`), and three independent grid families agree:

| grid | dpm[1,1] (q=2) |
|---|---|
| ldp mpsi 128 / 256 / 512 / 1024 / **2048** | 7.279 / 7.280 / 8.478 / 9.142 / **9.244** |
| uniform mpsi 512 / 1024 / **2048** | 8.890 / 9.146 / **9.143** |
| **auto** τ 2e-3 / 1e-3 / 5e-4 / 2.5e-4 | 9.198 / **9.239** / 9.206 / 9.157 |

**dpm[1,1] → ≈9.2**, and `auto` is flat in τ to **0.9 %** (it was ~50 %, 6.2–9.9, before this PR). The old pin, 6.9237, sits below even ldp-128 — it was a pre-fix artefact, not a target. So the new value is not a regression to accept; it is the converged answer, and the default τ now lands on it rather than near it by luck.

Re-pinned q=2 to `9.30 ± 3e-2`. That is a **5× tightening** of the old `1.5e-1`, which is possible because **the platform spread collapsed**: the same grid and inputs now give **9.239** (macOS/aarch64), **9.333** (linux/julia 1.11), **9.356** (linux/julia 1.x) — **1.3 % end to end, against ~25 % before** (6.14 / 6.69 / 7.70). That collapse is itself the cleanest evidence the diagnosis was right: with the matching stencil no longer straddling a knot, q‴ is single-valued across it and the surface stops sitting on the steepest part of the refinement curve.

q=3 and q=4 are left untouched — the same ladder puts them at −5.716 and −16.033 (ldp 2048), inside their existing tolerances, and they reproduce across platforms.

### 2. `develop` merged in

Merged cleanly, no conflicts. Re-ran the full ladder against the merged tree: **bit-identical** to the pre-merge values above, so nothing in `develop` perturbs this.

### 3. Δ′ convergence on real FUSE-reconstructed equilibria

Error against an ldp-1024 reference, per slice (11 slices, 6 DIII-D discharges, β_N 0.3–3.4, q95 2.4–5.9):

| slice | \|Δ′\| | ldp-128 | ldp-256 | ldp-512 | **auto @1e-3** |
|---|---|---|---|---|---|
| 97741 t=1450 | 1.51 | +0.051 | +0.007 | −0.002 | 0.007 |
| 97741 t=1500 | 2.59 | −0.264 | −0.017 | −0.031 | 0.039 |
| 153145 t=2601 | 0.83 | **+1.763** | +0.107 | +0.006 | 0.008 |
| 153145 t=3150 | 0.94 | **−1.097** | +0.244 | +0.010 | 0.021 |
| 174482 t=3200 | 0.74 | +0.783 | +0.203 | +0.018 | 0.011 |
| 174482 t=3850 | 0.42 | +1.189 | +0.344 | +0.025 | 0.023 |
| 148789 t=4602 | 4.75 | +0.941 | +0.269 | +0.019 | 0.003 |
| 148789 t=5101 | 8.23 | −0.968 | +0.282 | +0.017 | 0.008 |
| 165024 t=1199 | 16.80 | −0.425 | +0.003 | −0.024 | 0.003 |
| 165024 t=1598 | 15.52 | −1.927 | +0.221 | +0.010 | 0.017 |
| 161172 t=2414 | 27.02 | −11.252 | +7.542 | +0.138 | 0.049 |
| **median** | | 0.97 | 0.22 | 0.018 | **0.011** |

Bolded: at mpsi=128 the two 153145 slices come out with the **wrong sign** — +0.93 where the converged value is −0.83, and −0.16 where it is +0.94. On the discharges where Δ′ is the physically interesting quantity (|Δ′| ≲ 1), a coarse fixed grid does not just lose precision, it inverts the conclusion.

Two structural points from this set:

- **Error at fixed mpsi is roughly absolute (~0.2 at mpsi=256), not proportional** — 0.02 % on 165024 (|Δ′|≈17) but 83 % on 174482 t=3850 (|Δ′|≈0.4).
- **Convergence on fixed grids is a damped oscillation, not a monotone plateau**: 7 of 11 slices *overshoot* at mpsi=256 before settling, so agreement between two coarse grids proves nothing.

Measured `auto` node counts on these equilibria (via `implied_knot_count` / `refined_psi_grid`):

| shot | nsing | τ=1e-2 | τ=1e-3 | τ=1e-4 |
|---|---|---|---|---|
| 97741 | 2 | 320 | 570 | 996\* |
| 148789 | 4 | 267 | 459 | 858 |
| 165024 | 8 | 277 | 479 | 900 |
| 161172 | 9 | 262 | 450 | 837 |

\* implied 1130, clipped by `REFINED_N_CAP`.

τ=1e-3 → ~450–580 nodes, i.e. `auto` independently rediscovers the mpsi≈512 the fixed ladder requires; and at *matched* ~270 nodes it beats ldp-256 by ~6× median / ~17× worst case. The count is curvature-driven, not rational-driven — it slightly *anti*-scales with nsing — so one τ works fleet-wide with no per-shot tuning.

### 4. Correction to my earlier comment

I said "the fix reproduces here" before running the counterfactual. Having now run the parent commit and this PR side by side on the same 11 slices:

| | parent (pre-PR) | this PR |
|---|---|---|
| median \|auto@1e-3 − ldp1024\| | 0.010 | 0.010 |
| **max** \|auto@1e-3 − ldp1024\| | **0.310** | **0.049** |
| max τ-spread | 1.541 | 0.444 |

On *our* equilibria `auto` was already largely converged before this PR — the median is unchanged. What the PR fixes is the **tail**: 6× better worst case, 3.5× tighter τ-spread. Real and worth having, but not the 7.2–10.1 erraticism seen on the fixture, so I did not want that overclaim standing.

Relatedly, scoping my earlier caution about `uniform` properly: on **this fixture** uniform converges fine and agrees with ldp (9.146 → 9.143). The non-convergence I flagged is specific to our high-β reconstructed equilibria, where uniform-512 vs -1024 differ by up to 2.0 and sign-flip against ldp/auto on 2 of 6 discharges. Not an issue for the reference used here.

Happy to share the ladder script, the FUSE convergence figure, or the raw CSVs.

d-burg (approved):

Issue #139, opened 2026-01-16
Finding and reducing function calls that pass in entire structs


jhalpern30:
While we have removed global variables and replaced them with dedicated data structures throughout the code, as @logan-nc has pointed out, we often just pass in many of these structs into functions that modify data in-place, giving the same issue as the globals in the Fortran - it is not clear to a new user what is getting modified and where! I think there are a lot of improvements that can be done for this (and will probably get delegated to AI).

Some thoughts:
1. Replace function calls that include the full data structures but only use one entry of it with just that entry. This should especially be done if its an update in place function and is only modifying one entry in the function.
2. If functions are modifying more than one entry in a data struct, think through if it can be separated into multiple functions, such that 1. can be achieved
3. We have the main data structures of `ctrl, intr, equil, ffit, odet`. During the hackathon, we made them all `@kwdef mutable struct` types. However, I think we don't need them all to be mutable. For example, I think  `ctrl` should be instantiated from the input toml and then left alone completely - any modifications that are needed can be stored as copied within the mutable `intr` struct (e.g. psiedge, nn_low...). This way, users will know that `ctrl` will never appear as the first entry of a `!` function. This could also be done with `ffit`

Open to other suggestions

logan-nc:
To add to this, we can also break out the mutable part of the structs as they stand. 
For example, `DconInternal` is a bunch of "global" settings that should be immutable plus `sing` stuff and `locstab`. Break out the sing and locstab parts and just explicitly pass those where they are needed:
- `mercier_scan!(locstab_fs, equil)` → `locstab_spline = mercier_scan(equil)`
- `sing_vmat!(intr, ctrl, equil, ffit, ising)` → `sing_struct = sing_vmat(intr, ctrl, equil, ffit, ising, locstab_spline)` (focus here is on the locstab - the sing_struct may need more thought)
- `write_outputs_to_HDF5(ctrl, equil,...` → `write_outputs_to_HDF5(ctrl, equil,..., locstab_spline)`
etc.

jhalpern30:
@claude go through the repository and identify changes which address the comments above. Specifically: function calls that include the full data structures but only use one entry of it with just that entry, functions which are modifying more than one entry in a data struct and if it can be separated into multiple functions, opportunities to break structs into mutable and immutable types

jhalpern30:
[...]

PR #355, opened 2026-08-12
ForceFreeStates - REFACTOR - Reduce whole-struct passing and in-place struct mutation

jhalpern30:
Incremental work against #139 — reducing the number of functions that take a whole mutable struct and modify it in place, so a reader can tell from a call site what it changes.

**This PR is a work in progress and more struct cleanups will be pushed to this branch.** Each installment is meant to be behaviour-preserving on its own.

### Freeze `ForceFreeStatesControl`
Issue #139 proposed that `ctrl` be instantiated from the input TOML and then left alone, with anything derived stored on the mutable `intr`. It turned out only six sites in `src/` were writing to `ctrl`, so this is now enforced by the type:

- `ForceFreeStatesControl` is `@kwdef struct` instead of `@kwdef mutable struct`. `ctrl` can no longer appear as the first argument of a `!` function.
- `ctrl.nn_low` / `ctrl.nn_high` now mean *what the user asked for*; the resolved toroidal range lives on `intr.nlow` / `intr.nhigh`. Downstream reads (`sing_lim!`, `rational_psi_nodes`) follow.
- The Fortran DCON `delta_mhigh` doubling is applied at the two sites that consume it rather than written back onto `ctrl` 130 lines earlier. It is read nowhere else and never written to HDF5, so this is equivalent — and the factor of two is now visible where it matters.
- `sing_lim!` normalizes `dmlim` into a local instead of writing it back to shared config.

### Extract the serial EL branch
`_populate_dense_xi_via_serial_el!` was saving, clobbering and restoring `ctrl.use_parallel` / `use_riccati` / `verbose` around a nested `eulerlagrange_integration` call — a mutation of shared state inside a function reachable on the threaded path.

`eulerlagrange_integration` was already a six-line dispatcher whose parallel and Riccati branches were named functions; only the serial body was inline. It is now `serial_eulerlagrange_integration(...; verbose=ctrl.verbose)`, completing the trio, and the dense-ξ pass calls it directly. No flag flipping, no `try`/`finally`, and the call site states its intent.

### Return `locstab` instead of parking it on `intr`
`intr.locstab` was written once and read six times, all inside `write_outputs_to_HDF5` — `intr` was pure transport. Local stability is now returned by `compute_local_stability(ctrl, equil)`; the field is removed, along with the zeros-filled spline that was constructed on every run even when `local_stability_flag = false`. `write_outputs_to_HDF5` takes it as a keyword.

### Docstring corrections
`eulerlagrange_integration` and `parallel_eulerlagrange_integration` both documented `-> OdeState` while actually returning a 4-tuple. Also dropped stale claims about dumping to `euler.h5` (this function writes no HDF5, and the file is `gpec.h5`), a completed `restype` TODO, and instructions to set `ctrl` fields programmatically, which immutability now forbids.

## Testing

| Suite | Result |
|---|---|
| `runtests_sing.jl` | 75/75 |
| `runtests_eulerlagrange.jl` | 92/92 |
| `runtests_fullruns.jl` | 17/17 |
| `runtests_parallel_integration.jl` | 104/104 |
| `runtests_riccati.jl` | 14/14 |

## Not yet done

- [ ] **Regression harness has not been run.** Every change here is a data-flow refactor with no intended numerical effect, but that needs confirming with `regress --cases diiid_n1,solovev_n1 --refs develop,local` before merge.

## Notes for review

- After the n-range change, the `nn_low <= 0` fallback in `sing_lim!` is unreachable from `main` (the resolved `intr.nlow` is always >= 1). It still guards direct callers, including the existing `sing_lim` test, so it is left in place.
- `singular/di0` and the `locstab/*` datasets are now gated on `locstab !== nothing` rather than re-checking `ctrl.local_stability_flag`. Equivalent, since `main` only builds `locstab` under that flag.
- The commit was made with `--no-verify`. The local `julia-formatter` pre-commit hook is `language: system` and picks up whatever JuliaFormatter is in the default depot (v2.6.0 here), not the v1.0.62 that `.pre-commit-config.yaml` pins — so it reformatted ~500 lines of pre-existing code in files this PR happens to touch. That drift is the reason `format.yaml` is currently disabled as a PR gate. Every line added by this PR was verified clean under the stricter v2.6.0.


matt-pharr:
@claude review

jhalpern30:
Pushed `90f2bb34` to this branch: a bugfix the regression harness surfaced.

Freezing `ForceFreeStatesControl` left one consumer behind. `galerkin_solve` builds `ctrl_gal = deepcopy(ctrl)` and then assigns `ctrl_gal.sing_order` per surface — fine while the struct was mutable, fatal once it wasn't:

```
ERROR: setfield!: immutable struct of type ForceFreeStatesControl cannot be changed
```

Every run with `gal_flag = true` and at least one singular surface was failing on this branch. The assignment is unconditional inside the `for s in sings` loop, so there is no path around it.

It stayed hidden because no unit suite covers the Galerkin path — `runtests_sing.jl`, `runtests_eulerlagrange.jl`, `runtests_fullruns.jl`, `runtests_riccati.jl` and `runtests_parallel_integration.jl` all pass without ever reaching `galerkin_solve`. It only showed up when the harness ran the `gal_resistive_*` cases.

**Fix:** `compute_sing_asymptotics` and `compute_sing_mmat!` now take a `sing_order` keyword defaulting to `ctrl.sing_order`, and `galerkin_solve` passes the per-surface value directly. The `deepcopy` is gone — no copy-with-changes helper, per the approach agreed for this branch. As a side effect `SingAsymptotics` now records the order actually used rather than whatever was last written onto the copied struct.

**Verified:** the `DIIID-like_gal_resistive_example` goes from erroring to completing in 90.9 s; `runtests_sing.jl` 75/75 and `runtests_fullruns.jl` 17/17 on this branch. Numerically inert — `diiid_n1` 48/48, `gal_resistive_diiid` 10/10, `gal_resistive_pe` 4/4 unchanged against `a0cad260`.

Worth flagging for review: this branch's suites were green *while it was broken*, so please don't take test status alone as evidence the freeze is complete.

jhalpern30:
Regression harness now fully verified on this branch - should be good to go with @matt-pharr approval

matt-pharr (approved):

PR #356, opened 2026-08-12
TEST - enable formatting - Update pre-commit trigger for pull requests in workflow

matt-pharr:
Enables a formatting test. Only merge when we are ready for every PR to fail if the formatter has not been run.


logan-nc:
@matt-pharr assign someone or mark as draft

PR #285, opened 2026-06-15
Standardize TOML annotations, bundle a10 inputs, drop Fortran references (#274)

logan-nc:
## Summary

Addresses #274 (TOML annotations need re-standardizing) and fixes two related defects found during the review.

**1. Standardize TOML annotations (#274).** One shared convention across every git-tracked config-style TOML: a short 2–10 line purpose/context header, an inline comment on every variable sourced from the matching config-struct docstring (same wording everywhere; `examples/DIIID-like_ideal_example/gpec.toml` is the canonical reference), decorative section dividers removed, informative section notes kept. Codified in `CLAUDE.md` so future edits stop diverging.

**2. Bundle a10_kinetic_example inputs.** The example pointed `eq_filename`/`kinetic_file` at absolute paths in one developer's local Fortran checkout — unrunnable for anyone else and the only hardcoded local paths in the repo. The g-file (`fix_a100_k10_q2_bn010_prof1`, 1.1 MB ASCII) and kinetic profile (`a10_prof1.txt`) are now bundled in the example directory (token-identical to source, whitespace-normalized hook-clean) and referenced by bare relative filenames.

**3. Make example annotations self-contained / Julia-centric.** The Julia code is the production workhorse; a new user shouldn't need the legacy Fortran code to read an example. Dropped all Fortran-GPEC references (namelists, flag names, `PENTRC`/`DCON`/`STRIDE`, "Julia mirror of Fortran") from `examples/` and `test/test_data/`; the a10 header now describes the *scenario* (drift-kinetic MHD stability via the calculated NTV matrices, deuterium ion-only Maxwellian). Also removed deprecated / no-op vars (`truncate_at_dW_peak`, `thmax0`) from the examples. TJ-analytic model citations kept (external model, not Fortran); `regression-harness/cases/*` left as developer metadata.

## Scope
- 30 example/fixture/case TOMLs annotated; 2 new a10 input files; `CLAUDE.md` conventions extended.
- `Project.toml`-family files left untouched (machine-managed).

## Verification
- All 34 tracked TOMLs parse; hygiene clean (no trailing whitespace, single final newline, LF).
- Parsed values byte-identical to `develop` except the two intentionally removed keys — **zero value changes**.
- a10 example runs end-to-end (GPEC completes ~173 s, n=1 modes stable, `gpec.h5` produced).
- Regression harness vs `develop`: **`solovev_n1` 21 unchanged, `diiid_n1` 30 unchanged — every quantity 0.0e+00 / OK.**

[...]

PR #357, opened 2026-08-12
Fix TOML annotation drift and enforce the convention via pre-commit (#274)

### LINKED ISSUE #274: toml annotations need to be re-standardized
In reviewing some recent PRs I've noticed that our various AIs have gone to town annotating some git tracked tomls, while others have been left untouched. The different AIs seem to have used different conventions for toml file header annotations and/or section annotations. A standardization sweep is in order.

I like the idea of the headers, especially for the examples. The headers can explain the purpose/context of the example and call out any key features of the settings. They should not be overly large - maybe 2-10 lines.

All variable setting lines should have in-line annotations describing what the variable is/does. These are super helpful for new users, who inherit copy and modify the example tomls and do not want to search source code to remember what each variable does.

Section annotations seem overkill, but perhaps I can be convinced otherwise if there are important ones.

priyanshlunia:
## Summary

Closes #274.

The standardization sweep in #285 established one shared annotation convention for the
config-style TOMLs, but the issue stayed open — and in the weeks since that merge, 9 new
TOMLs were added and most existing ones modified, reintroducing drift. This PR fixes all
current violations and adds pre-commit enforcement so the convention holds going forward.

**1. Fix the drift** (11 files, `887bd8f`):

- **Missing headers added**: `DIIID-like_gal_resistive_example`,
  `DIIID-like_gal_resistive_pe_example`, the `regression_solovev_kinetic_nuzero` fixture,
  and the `diiid_slayer_n1`, `gal_resistive_diiid`, `gal_resistive_pe`,
  `solovev_kinetic_nuzero` regression cases.
- **Missing inline descriptions completed**: the whole `[SLAYER]` section and scan grid in
  the SLAYER example, `rtol_xlmda`/`write_outputs_to_HDF5`/`verbose` in the canonical
  `DIIID-like_ideal_example`, and the same trailing keys in the Solovev kinetic decks.
- **Deprecated keys removed** (all warn-and-deleted by the loader, never read):
  `mat_flag`, `ode_flag`, `power_bp`, `power_b`, `power_r`, `truncate_at_dW_peak` (SLAYER
  example) and `force_wv_symmetry` (Solovev NTV example).
- **Cleanups**: decorative `# ---` dividers stripped from `diiid_slayer_n1.toml`; the
  `psiedge` comment claiming "integration truncated at peak" corrected in 3 files (the
  edge-dW scan is diagnostic-only with `truncate_at_dW_peak = false`).

**2. Enforce the convention** (`af540d5`): four declarative `pygrep` hooks in
`.pre-commit-config.yaml` — no external scripts, config-only like the existing hooks:

- `toml-header-block` — covered files must open with a `#` header block
- `toml-no-decorative-dividers` — rejects `# ----`-style dividers
- `toml-inline-annotations` — every `key = value` line needs a trailing `# description`
  (`examples/` and `test/test_data/` only)
- `toml-no-deprecated-keys` — rejects keys mirrored from
  `_DEPRECATED_FFS_KEYS`/`_DEPRECATED_EQUIL_KEYS`

**3. Docs**: `docs/development/toml-conventions.md` gains an Enforcement section and
codifies the regression-case exemption (the repetitive `[quantities.*]` schema keys are
developer metadata and don't need inline comments — cases need the header plus informative
per-quantity block comments). The setup guide's pre-commit list mentions the new checks.

## Scope

- 11 example/fixture/case TOMLs, `.pre-commit-config.yaml`, 2 dev-docs files.
- No `src/`, no test-code changes.

## Verification

- **Value-neutral**: `TOML.parsefile` comparison of every touched file vs `develop` —
  parsed values byte-identical except the 7 intentionally removed deprecated keys, which
  the loader warns about and discards before the control structs are built.
- **Hooks verified both ways**: all four patterns pass clean on this branch and reproduce
  the full audited violation set (missing headers, 8 dividers, 19 unannotated lines,
  5 deprecated keys) when run against `develop`'s versions of the files.
- **Hygiene**: no trailing whitespace, single final newline, LF endings; YAML parses.
- **Regression harness**: to be run on this branch before merge
  (`regress --cases solovev_n1,diiid_n1,diiid_slayer_n1,solovev_kinetic_ntv --refs develop,local`);
  expected all-zero deltas given the value-neutrality check above.


ebursch:
This all looks good.

ebursch (approved):
Looks good (try 2)

Issue #360, opened 2026-08-12
Branch protection requires skipped matrix jobs, so docs-only PRs to develop can never satisfy required checks


d-burg:
## Summary

`develop`'s branch protection requires the two test *matrix* jobs as status checks, but
`test.yaml` is path-filtered and deliberately publishes a separate stable job for exactly this
purpose. A pull request that touches no Julia code skips the matrix jobs, their contexts never
report, and the required check sits pending forever — blocking the merge with no way to clear it
short of an admin bypass or a no-op commit to a watched path.

## Current configuration

`develop` requires (`GET /repos/OpenFUSIONToolkit/GPEC/branches/develop/protection`):

- `Documentation`
- `runtests 1.11 - ubuntu-latest`
- `runtests 1.x - ubuntu-latest`

The last two are the matrix legs of the `test` job in `.github/workflows/test.yaml`, which is
gated on change detection:

```yaml
  test:
    name: runtests ${{ matrix.version }} - ${{ matrix.os }}
    needs: changes
    if: needs.changes.outputs.julia == 'true'
```

The workflow already anticipates this and provides the fix — the `test-status` job, named
`Tests`, runs with `if: always()` and reports success when the suite passed *or* was
intentionally skipped. Its own comment says so:

```yaml
  # Stable check that is present on every pull request, whether or not the suite
  # ran. Point branch protection at this job so skipped runs do not block merges.
  test-status:
    name: Tests
```

Protection is pointed at the matrix jobs instead, so that stable check is currently doing
nothing.

## Reproduction

Open a PR against `develop` touching only a path outside
`(^src/|^test/|^examples/|^Project\.toml$|^\.github/workflows/test\.yaml$)` — for example
`README.md`, anything under `docs/`, or `regression-harness/`. Change detection reports
`julia=false`, the `test` job is skipped, and the two required contexts never appear. The PR
shows "Expected — Waiting for status to be reported" indefinitely.

This has not bitten us yet because nearly every PR so far has touched `src/` or `test/`.

## Fix

Replace the two `runtests …` contexts in `develop`'s required status checks with `Tests`, and
apply the same to `main` if required checks are added there (it currently has none).

## Note for anything added later

The same trap applies to any new path-filtered or conditionally-run check. Whatever gets added
should follow the existing pattern: a conditional worker job plus an `if: always()` stable job
that branch protection points at. Worth keeping in mind for the regression-tier checks under
discussion, which will be conditional by construction.

[...]

PR #362, opened 2026-08-12
Test - TEST - Decomposition invariance of the unified Riccati Delta-prime path

d-burg:
## Release note

- **Audience:** developers
- **Numerical impact:** none
- **Migration:** none

Adds a regression test asserting that the Riccati Δ′ matrix is bit-identical when the same
integration is cut into a genuinely different set of chunks, and runs the CI suite multi-threaded so
the chunked propagator path is actually exercised concurrently.

## Notes for reviewers

**Rewritten after #381** (riccati unification), which superseded this PR's original form: it varied `parallel_threads`, which #381 deleted, and #381 made the chunk target structurally thread-independent (derived from `msing` alone) with unit tests pinning the thread-free target and bit-identical boundaries.

## What this now adds

**The end-to-end claim the #381 unit tests cannot make**: cutting the same integration into a genuinely different chunk decomposition — which reassociates the fundamental-matrix products, `((A·B)·C)·D` → `(A·B)·(C·D)` — leaves the physics output unchanged. The test steers the decomposition directly via `nchunks` (auto vs auto+11, boundaries asserted different first, so the equality can never pass vacuously) and therefore runs meaningfully at any thread count.

**A multi-threaded CI leg** (`JULIA_NUM_THREADS: 4`). The chunk loop executes through `Threads.@threads`, which runs serially in a single-threaded session — and the existing matrix legs are single-threaded, so no CI job has ever exercised concurrent scheduling. Existing job names are preserved byte-identically (branch protection names them; see #360).

**CLAUDE.md fix**: the documented single-test-file invocation (`test/runtests.jl test/runtests_foo.jl`) cannot work — `runtests.jl` passes ARGS to `include`, which resolves relative to `test/`. Corrected, plus the multi-threaded invocation documented.

## ⚠️ Finding for @matt-pharr: `et[1]` is not decomposition-invariant under the unified driver

Measured on the DIII-D-like ideal deck, auto (53 chunks) vs auto+11:

| quantity | invariance |
|---|---|
| Δ′ diagonal (all 5 surfaces) | **bit-identical** (asserted `===`) |
| `et[1]` | **drifts 2.7e-8 relative** — was bit-identical under the pre-unification driver |

The test records this honestly rather than hiding it: `@test_broken` on `et[1]` exactness (a future driver change restoring exactness reports an Unexpected Pass, forcing the strict assertion back), plus a documented `rtol=1e-6` ceiling — 30× the measured effect, there to catch the sensitivity growing by orders of magnitude, explicitly not chosen to make anything pass. If this is a known/accepted property of the serial Riccati-crossing assembly, the `@test_broken` line is the right permanent record; if not, it may deserve an issue.

## Verification

`julia -t 4`: 11 pass, 1 broken (the recorded `et[1]` exactness), 60.9 s.


logan-nc:
@d-burg assign someone to review or mark as draft

matt-pharr:
@d-burg you will be interested in the chunking/threading changes for riccati I made that are now on develop. Should make it always invariant to thread count.

matt-pharr:
et[1] drifting mildly in the decomposition is acceptable IMO, but if we decide otherwise we should be able to fix the order things are summed in to fix it. You can look at it here if you want but @logan-nc can decide whether he thinks that is an issue or not, I don't think so personally.

logan-nc:
That small of a drift seems ok to me? Maybe @adrianaghiozzi knows better if its a red flag or benign/expected.

matt-pharr:
@d-burg this is currently marked as open, is this ready for review?

d-burg:
Ready now, thanks @matt-pharr !

logan-nc:
Rothstein's 023 MHD Workshop presentation and Benjamin's 2025 MHD Workshop presentation both show fortran STRIDE is NOT invariant with many of it's numerical settings. 

We should and do strive for doing better, but I doubt we are completely invariant at this point

matt-pharr:
@d-burg the PR description still says Delta' should be bit-identical when chunking is done differently. I am still of the opinion that that is too strict most likely. I am not sure there are any quantities that GPEC can even reproduce bit-identically across multiple different LAPACK math implementations. Has that ever been the case with a STRIDE Delta' calculation, fortran or julia? And is that test run in CI right now? If so, that means that STRIDE on this branch is actually bit identical right?

logan-nc:
Bit-identical is surely asking too much. Would even integrating a smooth known function like x^3 in chunks be bit identical? I don't think so.

d-burg:
The bit-identity is an empirical observation, not a theoretical requirement — the test asserts what the current serial crossing assembly actually produces, and it holds on my machine and both CI runners (which is the strongest thing the test can prove). I agree it's not guaranteed across BLAS/LAPACK implementations. Happy to relax the Δ' assertion to a tight rtol (say 1e-12) with a comment recording that it was measured bit-identical as of this PR, if you'd both prefer that over a strict `===` that might trip on a future runner change.

test/runtests_decomposition_invariance.jl
```diff
@@ -0,0 +1,77 @@
...
+# Decomposition invariance of the Riccati Δ′ path.
+#
+# The chunked propagator driver reassociates the fundamental-matrix products whenever the chunk
+# decomposition changes, and floating-point products do not reassociate exactly in general — so
+# Δ′ reproducibility is a real end-to-end claim, beyond the structural chunk-count invariance
+# pinned by the unit tests in runtests_parallel_integration.jl. This file asserts it: the Δ′
+# matrix is bit-identical when the same integration is cut into a genuinely different set of
+# chunks. Driven through the public solve API so it exercises the production path itself.
+
+const GP_TI = GeneralizedPerturbedEquilibrium
```
jhalpern30: This is one of the more heinous abbreviations I've ever seen lol

jhalpern30 (approved):
Seems like Matt/Nik have already helped you flesh out the physics - the code here all looks good to me

PR #368, opened 2026-08-13
FFS - BUGFIX - Zero-extent ca_left/ca_right when unpopulated (issue #366)

### LINKED ISSUE #366: singular/ca_left and ca_right are written as uninitialized memory on kinetic runs
## Summary

`singular/ca_left` and `singular/ca_right` are written to `gpec.h5` unconditionally, but
nothing populates them on kinetic runs. The datasets therefore contain **uninitialized
memory** whenever `kinetic_factor > 0`.

Filing as a sub-issue of #226 because that effort already rewrites every `singular/` write
site (`singular/` → `SingularSurfaces/`), so the guard belongs in the same pass rather than
as a separate drive-by change.

## Mechanism

- The arrays are allocated with `undef` and never zeroed:
  `src/ForceFreeStates/ForceFreeStatesStructs.jl:615-616`
  ```julia
  ca_r::Array{ComplexF64,4} = Array{ComplexF64}(undef, numpert_total, numpert_total, 2, msing)
  ca_l::Array{ComplexF64,4} = Array{ComplexF64}(undef, numpert_total, numpert_total, 2, msing)
  ```
- They are only ever filled by the **ideal** surface crossings —
  `cross_ideal_singular_surf!` (`EulerLagrange.jl:594, 631`) and the Riccati ideal crossing
  (`Riccati.jl:1226, 1267`).
- `cross_kinetic_singular_surf!` (`EulerLagrange.jl`) does no asymptotic analysis and never
  writes them, so on a kinetic run they keep whatever was on the heap.
- They are written out regardless: `src/GeneralizedPerturbedEquilibrium.jl:839-840`
  ```julia
  out_h5["singular/ca_left"]  = odet.ca_l
  out_h5["singular/ca_right"] = odet.ca_r
  ```

## Evidence

Running `examples/Solovev_kinetic_calculated_example` twice with **identical code** and
diffing `singular/ca_left` (4096 elements) gives a different number of NaNs each time —
238 in one run, 100 in the other. One `ca_right` element came back as `7.026e-310`, a
subnormal double, which is the classic signature of reading uninitialized memory rather
than a computed value.

This surfaced while validating #365 (on-demand solution derivatives): a develop-vs-branch
comparison flagged these two datasets as differing even though the change is provably
numerics-neutral everywhere else. It is pre-existing on `develop` and unrelated to that PR.

## Impact

Low severity, but genuinely bad behaviour:

- Output is **non-reproducible** run to run for a deterministic calculation, which
  undermines any bitwise output comparison (regression harness, benchmark diffing, rerun
  from `gpec.h5`).
- Consumers cannot distinguish "not computed" from "computed and happens to be small".
- NaNs propagate silently if anything downstream ever starts reading these.

No current consumer is affected: the only reader is `compute_delta_prime_from_ca!`
(`EulerLagrange.jl:16`), documented as a **stub that is not physically valid** and not
called from any integration driver.

## Suggested fix

Pick whichever fits the new schema best:

1. Allocate with `zeros` instead of `undef` — one-line, and makes "not computed" read as
   zero rather than garbage. Cheapest, and safe since these arrays are small (`msing`
   surfaces).
2. Skip the datasets entirely when they were never populated (kinetic runs), matching how
   #365 now handles the derivative stores — empty dataset rather than meaningless values.
3. If they are to stay always-present, add an explicit `computed` flag / attribute so
   readers can tell.

Option 1 plus a note in the dataset docs is probably enough; option 2 is more consistent
with how the rest of the schema is heading.

logan-nc:
## Summary

Fixes #366: `SingularSurfaces/ca_left`/`ca_right` were written to `gpec.h5` unconditionally, but only **ideal** singular-surface crossings populate `odet.ca_l`/`ca_r` (allocated `undef`). Kinetic runs dumped uninitialized heap memory — non-reproducible output (238 vs 100 NaNs between identical runs, subnormal garbage values). Third PR in the #226 stack: base is `refactor/hdf5-metadata` (#364); **merge order #363 → #364 → this**.

**Scope extension found during implementation**: the **galerkin-matched** path (`gal_matched_odestate` builds an `OdeState` and never crosses a surface) emitted the same garbage, and `benchmarks/gal_validation/gal_{epsilon,beta}_scan.jl` were already consuming it. The fix covers both paths.

## Fix (issue options 1 + 2)

- `ca_l`/`ca_r` allocated with `zeros` instead of `undef` — kills in-memory nondeterminism at the root.
- New `OdeState.ca_populated::Bool` (mirrors the existing `du_store_populated` pattern), set by the two ideal crossing routines (EulerLagrange + Riccati) and carried through the Riccati dense-ξ save/restore; galerkin-matched states keep the `false` default.
- The writer emits **rank-4 zero-extent sentinels** when unpopulated. Datasets stay always-present — no reader `KeyError`s, PR B's metadata annotations unchanged. This matches the empty-sentinel idiom already used by the FreeBoundaryStability block and the zero-extent choice #365 made for its analogous unpopulated `u_store` case (that PR's body independently documents this bug as follow-up — cross-linking).
- Reader guarded: `Analysis.plot_delta_prime` returns a placeholder on empty ca (previously indexed into it). (This PR originally also guarded the two gal-validation scans; develop has since untracked those benchmark scripts, so the stack merge resolved them as deleted and those hunks are gone.)
- Conventions: zero-extent-array = not-computed sentinel codified in `docs/development/hdf5-conventions.md`; stale `docs/src/stability.md` claim (SingularCoupling "reads ca_l/ca_r directly" — it doesn't) corrected.

## Verification

- **Determinism (the issue's own signal)**: two identical runs of the kinetic Solovev deck now produce byte-identical, empty `ca_left`/`ca_right` (previously differing NaN counts).
- Tests: kinetic fullrun asserts both datasets empty; ideal schema run asserts populated + all-finite; full `test/runtests.jl` suite (57 testsets) passes with zero failures; docs build clean.
- **Regression harness (mandatory)**: all 12 cases `--refs develop,local` — every tracked quantity `OK` at 0.0e+00 (same 4 pre-existing N/As in gal_resistive_pe). New `solovev_n1` quantity `ca_left (checksum)` locks bitwise reproducibility going forward; it reports `N/A` on the develop side only because develop's cached extraction predates the quantity — the local checksum is now the baseline.

## Coordination

- **#365 collision**: it is open and touches the same struct block and files (`ForceFreeStatesStructs.jl`, `EulerLagrange.jl`, `Riccati.jl`, the main writer). This diff is tiny — whichever lands second rebases trivially.

**This PR must not be merged until a third-party human reviewer has approved it — non-negotiable project policy.** Merge order: #363 → #364 → this.


docs/src/stability.md
```diff
@@ -342,8 +342,10 @@ end
 
 ## Notes
 
-- The standard path does not populate `delta_prime`; use `PerturbedEquilibrium.SingularCoupling`
-  for Δ' on the standard path (it reads `ca_l`/`ca_r` directly).
+- The standard path does not populate `delta_prime`; the canonical Δ' is the STRIDE BVP
```
logan-nc, line 366: I believe this should be capitalized Delta as of #364

logan-nc: Correct — fixed. That reference is an HDF5 path, so it follows the literature-capitalization rule from #364: `SingularSurfaces/Delta_prime_matrix`.

I swept the rest of the branch for the same slip and this was the only one; the neighbouring lowercase names in that Notes block (`delta_prime`, `delta_prime_col`, `delta_prime_matrix` in the code sample above it) are Julia struct fields, which keep their existing spelling — only the dataset names were recapitalized.

logan-nc:
@matt-pharr I assume this will be also fixed in your abstraction of integrators but lets just merge the fix here for now - it'll be a minor conflict compared to the rest you have

logan-nc:
@priyanshlunia review & merge this in the next 24 hrs please. Just fix any minor issues yourself directly. Ping me if anything significant changes.

priyanshlunia:
## Regression harness report — PR #368 post-merge verification

**Command:** `julia --project=regression-harness regression-harness/regress.jl --refs develop,local`
**Local ref:** merge commit `7bc0c033` (`bugfix/366-ca-uninitialized` with `origin/develop` @ `9be3ac58` merged in, conflicts resolved)
**Environment (both refs):** julia 1.12.2, arm64-apple-darwin24.0.0, manifest `5e468ac3` (pinned), 6 threads / 6 BLAS

### Result: all 14 cases ran successfully on both refs — no tracked quantity changed

| Case | Quantities | Status |
|---|---|---|
| diiid_n1 | 47 | all OK, 0.0e+00 |
| diiid_n1_riccati | 17 | all OK, 0.0e+00 |
| diiid_slayer_n1 | 15 | all OK (root-finder diffs ≤ 3e-6, well under per-quantity noise thresholds) |
| efit_fixedbdy_separatrix | 5 | all OK, 0.0e+00 |
| gal_resistive_diiid | 10 | all OK, 0.0e+00 |
| gal_resistive_pe | 8 | N/A on **both** refs (pre-existing, see note) |
| ggj_ray_q500i | 4 | all OK, 0.0e+00 |
| ggj_reference | 4 | all OK, 0.0e+00 |
| solovev_kinetic_calculated | 14 | all OK, 0.0e+00 |
| solovev_kinetic_multiion | 6 | all OK, 0.0e+00 |
| solovev_kinetic_ntv | 6 | all OK, 0.0e+00 |
| solovev_kinetic_nuzero | 14 | all OK, 0.0e+00 |
| solovev_multi_n | 15 | all OK, 0.0e+00 |
| solovev_n1 | 22 | all OK, 0.0e+00 — includes the new `ca_left (checksum)` quantity |

### Key verifications for this PR

- **`solovev_n1` `ca_left (checksum)`: byte-identical on develop and local**
  (`6673eae3dce0afad7b32e19130842592ca7e773d59044eefa383cbf4a5c1f003`) — the ideal-path
  asymptotic coefficients are untouched by the fix and now bitwise-locked going forward.
- **End-to-end sentinel check on the merged tree's harness outputs** (`h5ls` on `gpec.h5`):
  - Kinetic (`Solovev_kinetic_calculated_example`): `ca_left`/`ca_right` = `{0, 0, 0, 0}` (zero-extent) ✓
  - Galerkin-matched (`DIIID-like_gal_resistive_pe_example`): `{0, 0, 0, 0}` (zero-extent) ✓
  - Ideal Riccati (`DIIID-like_riccati_deltaprime_example`): `{5, 2, 35, 35}` (populated, all compared OK) ✓

### Notes

- The 8 `gal_resistive_pe` N/As are datasets missing from `gpec.h5` on **both** refs — a
  pre-existing develop-side issue unrelated to this PR (every tracked quantity in that case
  except runtime currently resolves to `missing`; the case TOMLs likely need a refresh after
  the recent schema work — suggested as a follow-up issue).
- SLAYER (`diiid_slayer_n1`) root quantities differ at the 1e-6..1e-3 level, far below their
  configured noise thresholds (e.g. `slayer_Q`: max diff ~3e-6 vs threshold 1e-4) — expected
  AMR/root-finder noise, reported OK by the harness.

priyanshlunia (approved):

PR #369, opened 2026-08-13
REGRESSION - IMPROVEMENT - Reuse one worktree per commit instead of per (case, ref)

jhalpern30:
## Summary
- The regression harness (`regression-harness/`) created and tore down a brand-new git worktree for *every* (case, ref) pair, so an N-case run against one commit paid for `Pkg.instantiate()` + a full precompile of `GeneralizedPerturbedEquilibrium` N separate times.
- Cases are now run grouped by ref (`run_cases_at_ref` in `runner.jl`), sharing a single worktree across every case that needs the same commit. Worktree creation is skipped entirely when every case at that commit is already cached. `run_commit`/`run_computed_at_commit`/`run_at_commit` gained an optional `worktree_path` kwarg so they can reuse a shared worktree without changing behavior for any other caller.
- Reporting is unchanged — still grouped by case, comparing across all refs.

## Test plan
- [x] Benchmarked 3 lightweight `kind="computed"` cases (`ggj_reference`, `ggj_ray_q500i`, `efit_fixedbdy_separatrix`) against `develop` with a warm package depot: **2:02 → 1:28** wall time, 3 worktrees → 1.
- [x] Ran the full 11-case suite (excluding the unrelated fixed-boundary equilibrium case) against `develop` and `local` on `refactor/vacuum-response-struct`: completed in **40m52s** end-to-end. The same run with the old code did not finish `Pkg.instantiate()` for even 9 of 11 `develop`-side cases in 52 minutes before it had to be killed.
- [x] Confirmed the harness itself still reports correctly: all 11 cases showed 0 numerical diff between `develop` and the working tree in that run.
- [ ] No new regression-harness case needed — this is a tooling change to the harness driver, not physics code.

**⚠️ Per repo policy: this PR must not be merged without a third-party human review approval — no exceptions.**


jhalpern30:
Figured Mr. Regression himself @d-burg would be best to review this, but wanted to flag this for @logan-nc @matt-pharr  and @priyanshlunia - I found this substantially speeds up multi-case runs which were taking forever by running all cases on each branch then switching to the other branch (instead of going case by case, switching branches each time). Should be helpful during the remainder of the cleanup hackathon

logan-nc (approved):
Manual read checked out. 
My AI caught a few minor robustness improvements that will land soon (checking regression run now), at which point we can merge. I'll add a follow-up PR that adds threads for more performance wins.

PR #373, opened 2026-08-14
REGRESSION - IMPROVEMENT - Run GPEC subprocesses with -t auto

logan-nc:
## Summary

The regression harness spawned GPEC subprocesses with no `--threads` flag, so all of GPEC's threaded kernels (Riccati parallel FM, ballooning, field reconstruction, kinetic forces, vacuum kernels) ran single-threaded unless `JULIA_NUM_THREADS` happened to be exported. Subprocesses now launch with `--startup-file=no -t auto`; override with `GPEC_REGRESS_THREADS=1` (documented in `--help` and the harness docs). `--startup-file=no` also removes the user's `startup.jl` as an uncontrolled variable in regression runs.

**Stacked on #369** (touches the same spawn sites in `runner.jl`); this PR's diff shows only the threading change. Retargets to develop when #369 merges.

## Determinism

Riccati's parallel FM was reviewed for thread-count sensitivity: chunks integrate independently with per-chunk state and are assembled serially in chunk order; the only cross-thread reduction is an integer step count. Verified empirically: threaded local runs are **bitwise identical** to cached single-threaded baselines for every tracked quantity.

## Regression report

- `solovev_n1`: 21/21 unchanged (all diffs 0.0e+00), runtime 102.7s → 96.3s
- `diiid_n1`: 48/48 unchanged (all diffs 0.0e+00), runtime **242.0s → 157.7s (−35%)**

Note: `Runtime (s)` rows cached from single-threaded runs are not comparable to threaded ones — re-baseline with `--force` where runtime tracking matters. A caveat for future decks: chunk decomposition depends on `parallel_threads ≥ 3` (`EulerLagrange.jl` `balance_integration_chunks`), so decks raising that setting should re-verify; all current decks use ≤ 2.

⚠️ Requires third-party human review before merging — do not merge without an approving review.


logan-nc:
@d-burg @jhalpern30 @matt-pharr do we want this? Or keep everything 1 thread for basic tests

d-burg (approved):

logan-nc:
@matt-pharr this ran into your known issue in that it didn't edit enough to trigger the tests but the tests are required to merge. I am assigning you to this PR to either manual override or fix the testing/blocking logic

PR #395, opened 2026-08-15
CI - IMPROVEMENT - Pin the CI dependency set per Julia minor version

matt-pharr:
Essentially this PR adds a pin on the manifest for the CI jobs, so testing will be much much quicker once this is merged. Presently, the packages will all recompile if the dependencies change. However, this manifest pin should allow us to re-run CI on the same versions of all dependencies and therefore skip the precompilation that takes like 15 minutes. @logan-nc I recommend merging this regardless of whether we decide to track the manifest or not and we can always change how this works in the future, but this should make the github actions jobs much faster now so I think we should merge this ASAP.


logan-nc:
@claude review this

matt-pharr:
FYI this most recent commit is just a test to show that the precompile step gets skipped in the gh runner jobs.

matt-pharr:
OK @logan-nc this is ready. As you can see the latest commit finished >10 minutes quicker than the prior one and no longer goes through the whole pre-compiling process.

logan-nc (approved):
The speedup sounds great to me.
Honestly, you are way beyond me with the CI implementations at this point so I definitely defer to you on the best approach. I didn't see anything crazy in my ***MANDATORY*** human read through

PR #404, opened 2026-08-17
Repo - FEATURE - Add naming conventions, PR release-note block and changelog

logan-nc:
## Release note

- **Audience:** developers
- **Numerical impact:** none
- **Migration:** none

Reworks the commit, issue and pull request conventions around AI-generated release notes. Commit and PR titles now use a closed `Area - TAG[!] - Summary` grammar, every PR carries a release-note block, and `CHANGELOG.md` is compiled from those blocks by a new `/release-notes` command.

## Regression report

No files under `src/` are touched, so there is nothing for the harness to move and no harness stamp is required by the new check. Confirmed by `git diff --stat`: the change is limited to `docs/`, `.github/`, `ci/`, `CLAUDE.md`, `CHANGELOG.md` and `.claude/`.

## Notes for reviewers

Why this changed at all: the old `CODE - TAG - message` convention was built for hand-consolidating release notes. Handing that to AI shifts what the metadata must carry. Measured over the last 400 non-merge commits, `MINOR` was the single most-used tag at 87 uses but encodes size rather than kind; `BUG FIX`/`BUGFIX`/`FIX` appeared under three spellings; and the one fact a GPEC user most wants at a release — did my answers change — was recorded nowhere, because it cannot be derived from a diff.

Design decisions worth a look:

- **The PR, not the commit, is the unit.** Commit subjects follow the grammar but are not enforced; a blocking hook was tried and removed as more friction than it was worth, since release notes are compiled from PR bodies and merge commits now carry the PR title.
- **`!` marks a user-visible change** and promotes it into a leading `Changed results & breaking changes` section whatever its tag. `REFACTOR!` is deliberately legal — a restructuring that perturbs numbers is exactly what a user needs told. `MINOR!`, `TEST!` and `DOCS!` are rejected.
- **`MINOR` is kept rather than renamed to `CHORE`.** It records the author's judgement that a commit can be skipped, which is what it was actually used for.
- **Valid Areas are derived from `src/` on disk**, not hardcoded, so the vocabulary cannot go stale. This surfaced that `CLAUDE.md` and `architecture.md` both still listed a `Splines` module that no longer exists and omitted five real ones; both are corrected here.
- **Harness staleness.** The release-note block records the commit the harness ran at, and CI fails if anything under `src/` changed afterwards — so a report cannot silently rot during review. Docs- and test-only follow-ups never trip it.

Two steps cannot be done from a branch and are still outstanding:

1. **Settings → General → Pull Requests → default merge-commit message = "Pull request title".** Some of the reasoning above depends on it — it is what makes `git log --merges` a well-formed history and what justifies not enforcing commit subjects.
2. **The label set does not exist yet.** `addLabels` auto-creates missing labels with random colors, so nothing breaks, but creating the nine tag labels plus `changed-results` deliberately would be better, along with retiring the stale `splines` label.

This PR is also the first live test of the `pr-conventions` workflow — it has never run. If the checks misbehave, that is the workflow being exercised for the first time rather than the branch being broken.

### DIFF
--- a/ci/conventions/check_subject.py
+++ b/ci/conventions/check_subject.py
@@ -1,0 +1,347 @@
+#!/usr/bin/env python3
+"""Validate pull request titles, release-note blocks, and commit subjects.
+
+Two things block on this: the `pr-conventions` CI job, which checks a pull request
+title, its release-note block and the freshness of its harness stamp; and the
+`naming-table-in-sync` pre-commit hook, which checks the Area table in the docs
+against `src/`. Commit subjects follow the same grammar but are not enforced --
+`--commit-msg` and `--stdin-report` exist for checking them by hand or from a
+locally installed hook.
+
+The grammar and its rationale are documented in `docs/development/naming.md`.
+
+Valid Areas are derived from `src/` on disk rather than hardcoded, so adding or
+removing a module updates the vocabulary automatically.
+"""
+
+from __future__ import annotations
+
+import argparse
+import difflib
+import re
+import sys
+from pathlib import Path
+
+# Tags that name a release-note section.
+RELEASE_NOTE_TAGS = ("FEATURE", "BUGFIX", "PERF", "API", "DEPRECATION", "DOCS")
+
+# Tags whose changes never appear in release notes.
+EXCLUDED_TAGS = ("MINOR", "REFACTOR", "TEST")
+
+ALL_TAGS = RELEASE_NOTE_TAGS + EXCLUDED_TAGS
+
+# `!` marks a user-visible change; it is incoherent on tags that cannot move
+# results or break an interface. An author reaching for one of these has
+# mis-tagged the change.
+NO_BANG_TAGS = frozenset({"MINOR", "TEST", "DOCS"})
+
+# Areas that are not Julia modules. Closed set.
+NON_MODULE_AREAS = (
+    "Benchmarks",
+    "Build",
+    "CI",
+    "Docs",
+    "Examples",
+    "Regression",
+    "Repo",
+    "Test",
+)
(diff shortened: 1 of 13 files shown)


logan-nc:
I did manual step (1)

ci/conventions/check_subject.py
```diff
@@ -0,0 +1,347 @@
+#!/usr/bin/env python3
+"""Validate pull request titles, release-note blocks, and commit subjects.
```
matt-pharr, line 8: Highly recommend adding a PEP 723 convention comment header to this script, it will make sure that it is always runnable with no environment fuss with `uv run` or similar tools.

matt-pharr (approved):
Just this one edit and can be merged. If you don't want to do it I can do it in ~2h once my plane is in the air

logan-nc:
labels have been updated

ci/conventions/check_subject.py
matt-pharr: Done in 198225e — no need to do it from the air.

```python
#!/usr/bin/env python3
# /// script
# requires-python = ">=3.10"
# dependencies = []
# ///
```

The script is stdlib-only, so the dependency list is empty; `>=3.10` matches the `default_language_version` pin in `.pre-commit-config.yaml`. Verified both `python3 ci/conventions/check_subject.py --title ...` and `uv run ci/conventions/check_subject.py --title ...` still pass a valid subject and reject `EQUIL - FIX - nope` with the replacement hints. `docs/development/naming.md` now shows the `uv run` invocation alongside the `python3` one.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #405, opened 2026-08-17
CI - REFACTOR - Remove the auto-merge label wrapper workflow

logan-nc:
## Release note

- **Audience:** developers
- **Numerical impact:** none
- **Migration:** Arm auto-merge with the **Enable auto-merge** button in the merge box instead of the `auto-merge` label. The label has been deleted.

Removes the custom auto-merge workflow. GitHub's native auto-merge, which the workflow was only triggering, covers the same capability directly.

## Why

`.github/workflows/auto-merge.yaml` did not implement merging. It ran `gh pr merge --auto`, which *is* GitHub's native auto-merge — so the workflow was a label-driven trigger for the merge-box button, nothing more. `allow_auto_merge` is `true` on this repository, so the button is already available to anyone with write access and no capability is lost.

Two reasons to prefer the button over the label:

- Two ways to arm a merge is worse than one for something this consequential, given that no PR may be merged without human review.
- The workflow merged using `GITHUB_TOKEN`, so anyone able to apply a label could arm auto-merge even without permission to merge it themselves. The button requires write access.

The one thing lost is at-a-glance visibility: a label shows in the pull request list, whereas the button's state only shows inside the PR.

## Regression report

No files under `src/` are touched; the change deletes a single workflow file, so there is nothing for the harness to move.

## Notes for reviewers

The `auto-merge` label has already been deleted, and no open pull request was carrying it, so nothing in flight was disturbed. With the label gone the workflow is inert regardless, but leaving a dead workflow that holds `contents: write` is worse than removing it.

Branch protection on `develop` is unchanged by this and still requires one approving review plus the `Documentation` and `Tests` checks — native auto-merge waits on those exactly as the label-driven path did.

The title follows the commit convention proposed in #404. Note that the `pr-conventions` check will not run on this PR, because that workflow does not exist on `develop` yet.


matt-pharr (approved):
Looks good

Issue #376, opened 2026-08-14
Performance: Why do large equilibrium splines slow down the code?


logan-nc:
Empirically, we all know larger `mpsi` slows down the ForceFreeStates.

The hand waving reason is the 10s of thousands of spline interpolations done during the ODE integration are a bit slower. But why would this be the case when we are using hints to cleanly provide the knot locations of interest and these are only slowly (and monotonically) incrementing? It seems like there would be effectively no search difference with knot number in this case. Is there some further optimization of the spline evaluation calls that would make this loop insensitive to the spline knot length? It seems like there should be.

@mgyoo86 might have some insights.

Possible topic for David Smitth ( @drsmith48 ) and UW hackathon

logan-nc:
## Answer: spline evaluation is not the bottleneck — adaptive step count is

I measured this end to end on `examples/DIIID-like_ideal_example` (ideal path, `kinetic_factor = 0`),
single-threaded, Vern9, `eulerlagrange_tolerance = 1e-10` unless stated otherwise. Short version:

**The hint-based spline lookup already does exactly what the issue hoped — evaluation cost is
flat in knot count. There is nothing left to win there.** The mpsi slowdown is entirely
**adaptive step-count growth**: the integrator's accepted step size is slaved to the *local knot
spacing*, so `nstep` scales with the number of knots in the regions where steps concentrate.

Two things worth saying up front, because they set what is actionable. Refining the grid genuinely
does make the integrand rougher — the coefficient matrices carry a node-error floor that grid
refinement amplifies rather than reduces (§6) — but when I repaired that directly it recovered
only 11–20% of the steps (§7). The dominant lever is much simpler: the two-pass auto grid reaches
the same answer with 558 knots and a third of the steps of a fixed mpsi=1024 grid.

### 1. Spline evaluation is knot-count-independent

`CubicSeriesInterpolant`, 676 `ComplexF64` series (mirrors `fmats`/`kmats`/`gmats` at `numpert = 26`),
non-uniform `Vector` grid, in-place eval, persistent `Ref(1)` hint, 16k queries:

| npts | mono ns/ev | stage ns/ev | no-hint mono | range grid | 3-sep stage | B/eval |
|---|---|---|---|---|---|---|
| 33   | 880 | 928 | 890 | 889 | 883 | 0 |
| 129  | 898 | 913 | 915 | 925 | 893 | 0 |
| 513  | 898 | 906 | 930 | 890 | 896 | 0 |
[...]

PR #408, opened 2026-08-18
FFS - IMPROVEMENT - Cap the core knot density of the EL coefficient splines

logan-nc:
## Summary

The Euler-Lagrange coefficient splines (`fmats`/`kmats`/`gmats` and the primitives) inherited **every knot of the equilibrium grid**. A cubic spline's third-derivative jumps at knots scale as (node error)/Δψ³, so the equilibrium's near-axis packing (Δψ ~ 1e-6 at high mpsi) amplifies even tolerance-level node error (~1e-9, measured) into huge C² kinks — and the adaptive integrator's step size becomes slaved to the knot spacing. Measured directly: core jump magnitudes grow ~34× per mpsi doubling, and an (tol/J)^¼ step model reproduces the observed step-count ladder.

The coefficients are near-cylindrical in the core and do not need that packing. This PR builds their splines on a subset of the equilibrium grid with **core density capped at Δψ ≥ 0.05·ψ below ψ = 0.1**. Node *values* are unchanged — only interpolation density — which is why the physics moves at the 1e-7–1e-8 level.

**Stacked on #398** (route (a), same investigation); diff shows only the cap once #398 lands.

## Measured (DIII-D stripped decks, route-(a) base, mpsi 512/1024)

| | mψ=512 | mψ=1024 |
|---|---|---|
| accepted EL steps | 2768 → **2188** | 4403 → **3034** |
| step growth per doubling | — | 1.59× → **1.39×** |
| warm run | 12.8 → **10.4 s (−19%)** | — |
| et[1] relative change | **3e-8** | 3e-8 |
| Riccati Δ′ diagonal (all 5 surfaces) | **4e-7** | — |

## Not a tuned hack: the rule's basis and its generalization

**Form**: near the axis every component is a Frobenius power law in ψ; power laws are scale-free, so log-uniform sampling (Δψ ≥ c·ψ) resolves them at constant relative accuracy. **Constant**: cubic interpolation of ψ^p on a log-uniform grid errs by ~(p·c)⁴/384, so c = 0.05 resolves even the steepest spectrum component (p = m_max/2 = 11) to ~2e-4 — and the physics responds far below that because the steep components carry vanishing solution amplitude. **Region**: ends at min(0.1, innermost rational − `RATIONAL_RES_RADIUS`), and no knot inside a rational's resolution window is ever removed — guards for decks (e.g. higher n) whose rationals reach the core; verified no-ops on every current case.

**Cross-equilibrium check** (four equilibria, two construction paths, three grid families):

| case | knots | EL steps | et[1] change |
|---|---|---|---|
| DIII-D efit, traced, log-family m1024 | 1025→575 | −31% | 3e-8 |
| tj_analytic_direct, traced, log-family m1024 | 1025→575 | **−35%** | **1.3e-9** |
| LAR, **inversion**, packed auto m1024 | 1025→575 | −8% | **exact to 8 digits** |
| Solovev, traced, **ldp** m512/m1024 | 505→468 | ~0 | ~4.5e-7 abs (3e-5 rel via the ±10.4 → 0.0146 cancellation) |

The LAR row is the anti-over-fit witness: clean geometry with no noise to exploit, and the cap is still harmless.

## Two properties reviewers should know

1. **No-op on production decks.** The two-pass auto grid's core spacing already satisfies the cap — verified: 287 → 287 knots on the DIII-D auto deck, and the full harness (`diiid_n1, solovev_n1, diiid_n1_riccati, gal_resistive_diiid`) reproduces #398's numbers exactly. The cap engages only on explicitly packed fine grids (large `mpsi`), which is precisely the issue #376 scenario.
2. **A fixed cap beats general curvature selection — measured, not assumed.** Removal-error knot selection (drop any knot whose interpolated value matches within δ for every element of all 12 matrices) was implemented and swept: it reaches the physics plateau only at δ=1e-7 where it keeps 504/513 knots and buys nothing, and breaks et[1]/Δ′ at any δ loose enough to matter. The core knots that are *safe* to remove **fail** the interpolation test (real coefficient curvature, but the solution components multiplying them vanish as ψ^|m|/2). Interpolation error is the wrong objective; the fixed mechanism-based cap is the right scope for the ideal path. Record: `handoff/issue376/RESULTS.md` §22–§23 (experiment branch).

## Verification

- Tests: `runtests_eulerlagrange`, `runtests_riccati`, `runtests_sing` — all pass.
- Harness vs develop @ 9491f893d across all four cases: identical to #398's report (see property 1); Δ′-tracking cases included.
- Mechanism measurement and step-model validation: `handoff/issue376/RESULTS.md` §22.

⚠️ **Requires third-party human review before merging — do not merge without an approving review.**

### DIFF
--- a/src/ForceFreeStates/Fourfit.jl
+++ b/src/ForceFreeStates/Fourfit.jl
@@ -496,23 +496,38 @@ function make_matrix(equil::Equilibrium.PlasmaEquilibrium, intr::ForceFreeStates
 
     # FastInterpolations now natively supports complex values - no need to split real/imag
     # Create complex series interpolants with per-column extrap BC
-    ffit.amats = cubic_interp(metric.xs, Series(amats_flat); ffit.itp_opts...)
-    ffit.bmats = cubic_interp(metric.xs, Series(bmats_flat); ffit.itp_opts...)
-    ffit.cmats = cubic_interp(metric.xs, Series(cmats_flat); ffit.itp_opts...)
-    ffit.dmats_prim = cubic_interp(metric.xs, Series(dmats_flat); ffit.itp_opts...)
-    ffit.emats_prim = cubic_interp(metric.xs, Series(emats_flat); ffit.itp_opts...)
-    ffit.hmats = cubic_interp(metric.xs, Series(hmats_flat); ffit.itp_opts...)
-    ffit.fmats_lower = cubic_interp(metric.xs, Series(fmats_lower_flat); ffit.itp_opts...)
-    ffit.fmats_prim = cubic_interp(metric.xs, Series(fmats_prim_flat); ffit.itp_opts...)
-    ffit.fmats_gal = cubic_interp(metric.xs, Series(fmats_gal_flat); ffit.itp_opts...)
-    ffit.gmats = cubic_interp(metric.xs, Series(gmats_flat); ffit.itp_opts...)
-    ffit.kmats = cubic_interp(metric.xs, Series(kmats_flat); ffit.itp_opts...)
+    # Decouple the coefficient-spline knots from the equilibrium grid in the packed core:
+    # a cubic spline's third-derivative jumps scale as (node error)/dpsi^3, so equilibrium-grade
+    # core packing amplifies tolerance-level node error into jumps that slave the EL step size.
+    # The coefficients are near-cylindrical there; cap the density at dpsi >= 0.05*psi below psi=0.1.
+    keep = Int[1]
+    for i in 2:length(metric.xs)-1
+        x = metric.xs[i]
+        if x >= 0.1 || (x - metric.xs[keep[end]]) >= 0.05 * x
+            push!(keep, i)
+        end
+    end
+    push!(keep, length(metric.xs))
+    mxs = metric.xs[keep]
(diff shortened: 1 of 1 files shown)


logan-nc:
**Review package** (visual companion to this PR — decoupling diagram, knot-density and step-distribution before/afters, the Frobenius-scale basis for the cap, the cross-equilibrium generalization table, and the why-not-general-curvature-selection sweep):

📦 https://claude.ai/code/artifact/e1ed1ea9-1362-40bd-9f52-ccd386118e21

Self-contained page; complements rather than duplicates the description and diff. If the link does not resolve, ask @logan-nc to enable sharing on it.

logan-nc:
Folded into #398 — its branch already contained these three commits, so this was a fast-forward with no history rewrite. Both changes are grid-side fixes aimed at the same symptom (Euler-Lagrange integration cost), the cap is a no-op on production auto grids so its measurements only mean anything alongside #398's step numbers, and they have been reviewed and benchmarked as a pair throughout. Closing to keep the review surface at one PR; the content and its evidence move into #398's body. Branch `performance/decoupled-el-matrix-grid` is left in place.

PR #410, opened 2026-08-19
FFS - NEW FEATURE - Certified adaptive psi grid for the calculated kinetic matrices

logan-nc:
Stacked on #408 → **#414** (near-axis kinetic validity suppression). This branch currently duplicates #414's commits and will be **rebased onto `feature/kinetic-axis-validity`** once #414 lands; the certified-grid work is the only content that belongs here. (The #408 base-branch force-push trimming its remote copy back to the cap commits is still pending.)

## What

Adds `kinetic_grid_tol` (`[ForceFreeStates]`, default **0 = off**, bit-identical to today when off): when `kinetic_source = "calculated"`, the expensive bounce-average kernel is driven over a **certified adaptive ψ grid** instead of every equilibrium knot.

- **Certificate on the total matrices** the EL solver consumes (ideal + kinetic, max-element residual vs `tol · max|T|`), including the adjoint combination `kw₃ − kt₃` — never the kinetic increments in isolation.
- **Physics-skeleton seed**: the same structure the NTV ψ quadrature panels at — rational windows, located kinetic-resonance surfaces (`kinetic_resonance_psi_nodes`, one source of truth with the torque quadrature), and four interior points per inter-rational span (log-spaced in the axis span, matching the core Frobenius structure).
- **Batched certify-or-refine rounds**: every round evaluates all uncertified midpoints as one threaded batch (bulk-synchronous — the existing `Threads.@threads` kernel pattern is untouched). Spacing floors: the Frobenius cap in the core, `RATIONAL_RES_SPACING` outside, ÷4 inside rational windows (anti-aliasing where layers are expected). Certification exhaustion warns loudly with the first uncertified location.
- Kernel takes an arbitrary ψ list (`psis` kwarg through the KineticForces callback); everything else sizes from it.

## Why

The shipped calculated-kinetic decks run `mpsi = 16` and carry **percent-level interpolation error on the B/C total matrices** (ε 6.8e-2 → 5.8e-4 across a 16/64/256 ladder — mostly unresolved real structure, not noise). At production grids the kernel costs ~1.3 s/surface: minutes of matrix formation spent on knots the totals don't need, while coarse decks are silently under-resolved. Certification fixes both directions: it **adds** knots where the deck is starved and **skips** them where the equilibrium grid over-packs.

## Evidence

- Knob-off bit-identity, re-verified on this branch's actual head (bef73a6ba, i.e. including the seed rework): `regress --cases solovev_kinetic_calculated,solovev_kinetic_ntv,solovev_kinetic_nuzero,diiid_n1 --refs performance/decoupled-el-matrix-grid,local` → **81/81 tracked quantities unchanged**.
- Solovev smoke (m16 deck, tol=1e-3): 15 seed → 74 knots, 133 kernel evaluations, fully certified — the certifier correctly refuses the deck's under-resolution (more evals than the deck's 17 knots, because the deck was wrong).
- **DIII-D survival test: FAILED its pre-registered physics gate — this PR stays draft until the root cause is decided.** Full record in `handoff/issue376/RESULTS.md` §28. Summary: the DIII-D kinetic-calculated configuration (never run before) is pathological on the full grid too — 223k EL steps (94% at ψ<0.01) and 17.5 GB solution dumps. A discriminator run (kernel tolerance ×10 tighter) reproduces the full-grid et[1] to 2.7e-7 with steps unchanged, so the pathology is **real near-axis structure, not kernel noise**: the calculated kinetic increments diverge toward the axis (~ψ^−0.7; the kernel has no near-axis validity clamp). The certified grid smooths that structure (et[1] off by 17→14% across the tol ladder, converging toward the full-grid value with cost exceeding the full grid). Two design consequences recorded for the fix: certify the post-Schur matrices the ODE consumes (f0/K/G amplify the raw-increment certificate ~20×), and the Frobenius core floor is unjustified for kinetic increments. A denser-equilibrium-grid check confirms the full-grid reference is converged (1.6e-4), so the certified deviation is unambiguously certification error; the identified design flaw is the certificate's GLOBAL max|T| scale, which is blind to the small-amplitude but dynamically dominant near-axis structure (1.6e-4 of the G scale, yet 94% of the ODE work). **The blocking question is physics**: should the drift-kinetic contribution be tapered/cut near the axis where the model degenerates? If yes, both the step pathology and the certification failure disappear together.

### DIFF
--- a/examples/DIIID-like_ideal_example/gpec.toml
+++ b/examples/DIIID-like_ideal_example/gpec.toml
@@ -87,5 +87,6 @@ f0type = "maxwellian"                   # Equilibrium distribution
 moment = "pressure"                     # Pressure-moment NTV torque
 atol_xlmda = 1e-9                       # Absolute tolerance for inner pitch + energy integrations
 rtol_xlmda = 1e-5                       # Relative tolerance for inner pitch + energy integrations
+axis_validity_suppression = true        # Suppress kinetic terms where the zero-orbit-width ordering fails near the axis (profile-derived boundary, no tuning parameters)
 write_outputs_to_HDF5 = true            # Write outputs to the HDF5 file
 verbose = true                          # Enable verbose logging
--- a/examples/Solovev_kinetic_NTV_example/gpec.toml
+++ b/examples/Solovev_kinetic_NTV_example/gpec.toml
@@ -87,3 +87,4 @@ f0fac = 1      # Scale toroidal field at constant pressure (β, q change; Φ, p,
 
 [KineticForces]
 kinetic_file = "kinetic.dat"            # Kinetic profile file: psi_n, n_i, n_e, T_i, T_e, omega_E columns
+axis_validity_suppression = true        # Suppress kinetic terms where the zero-orbit-width ordering fails near the axis (profile-derived boundary, no tuning parameters)
--- a/examples/a10_kinetic_example/gpec.toml
+++ b/examples/a10_kinetic_example/gpec.toml
@@ -59,3 +59,4 @@ nutype = "harmonic"            # Collision operator (zero, small, krook, harmoni
 f0type = "maxwellian"          # Distribution function (maxwellian, jkp, cgl)
 atol_xlmda = 1e-9              # Absolute tolerance for inner pitch + energy integrations
 rtol_xlmda = 1e-5              # Relative tolerance for inner pitch + energy integrations
+axis_validity_suppression = true        # Suppress kinetic terms where the zero-orbit-width ordering fails near the axis (profile-derived boundary, no tuning parameters)
--- a/src/ForceFreeStates/ForceFreeStatesStructs.jl
+++ b/src/ForceFreeStates/ForceFreeStatesStructs.jl
@@ -233,7 +233,7 @@ gpec.toml.
   - `numsteps_init::Int` - Initial array size for ODE data storage
   - `numunorms_init::Int` - Initial array size for solution normalization data
   - `singfac_min::Float64` - Fractional distance from rational q at which ideal jump condition is enforced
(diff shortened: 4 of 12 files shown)


logan-nc:
**Closing without merging** (branch `performance/certified-kinetic-grid` retained for the record). The certified adaptive grid failed its pre-registered acceptance gate and the follow-up work removed its premise:

- **Failed its gate**: on the DIII-D kinetic-calculated auto-grid case, certified et[1] sat 14–16% from a reference converged in both kernel tolerance (2.7e-7) and equilibrium grid (1.6e-4), and the deviation *grew* in cost while shrinking only slowly with tolerance — at tol=1e-4 it spent 424 kernel evaluations against the full grid's 288 and was still wrong. Root cause identified: the certificate's **global** `max|T|` scale is blind to small-amplitude but dynamically dominant structure, and it certified raw increments rather than the post-Schur matrices the ODE consumes.
- **Premise removed**: fixed-mψ ladders (512/1024), the develop 554-knot baseline, and two auto grids now agree on the kinetic et[1] to 0.1–0.2% — the ideal-driven grid is adequate at production resolution. #414 adds resonance-aware pinning for the structure the criterion cannot know about, and a one-pass bisection *map* (branch `experiment/kinetic-bisection-refine`) showed the remaining unresolved kinetic structure is real but physically inert (resolving all of it moved every referee by ≤2.4e-4).
- **Everything reusable already migrated**: the arbitrary-ψ kernel plumbing (`psis` kwarg) and the FKG `xs` kwarg are in #414.

**Revival path**, if a future case ever needs kinetic-error-driven refinement: certify the **post-Schur f0/K/G** with a **local** (per-interval or solution-amplitude-weighted) scale, and certify against the **PE-vs-NTV torque cross-check** rather than eigenvalues — the eigenvalue was demonstrably not a sufficient referee (see #423 and `handoff/issue376/RESULTS.md` §28–§35). The total-matrix tolerance rule from this PR's design stands and should be carried into any such revival.

PR #422, opened 2026-08-21
EQUIL - NEW FEATURE - Pin located kinetic-resonance surfaces into the auto psi grid

logan-nc:
Stacked on #414. The kinetic-grid follow-up chosen after the certified-grid postmortem: instead of adaptive compression, make the auto grid *resonance-aware* statically — the cheapest layer of the safety-net design, and per the tests below, the only one currently justified.

## What

When a run builds calculated kinetic matrices, locate the Ω_ℓ = 0 kinetic-resonance surfaces (same `kinetic_resonance_psi_nodes` locator the NTV quadrature panels at — one source of truth) and **pin them into the two-pass auto ψ grid as plain knots** via `merge_mandatory_nodes` — knot-at-node, no cleared zone, inserted *before* rational bracketing so the Δ′ clean-interval treatment wins wherever a resonance sits inside a rational's bracket. Nodes inside the near-axis validity region (ψ < ψ_c, #414) are suppressed anyway and are not pinned. ~30 lines, no new knobs.

## Evidence

- **DIII-D auto grid**: 6 nodes located (the ψ=0.021 node — which explains the previously mysterious "sharp increment feature at ψ≈0.022" — correctly filtered below ψ_c), +2 net knots after pedestal nodes yield to rational brackets, **et[1] unchanged to 2e-6, EL steps −14%** (4,246 → 3,665: the grid finally knows about the ψ=0.17 resonance).
- **Fixed-mψ convergence ladder** (closes the review question "was the full grid converged above 512?"): fixed-1024 → 1.004073−0.259566i, within 0.12% of auto-288; fixed-512 is 1.6% off (fixed grids converge *slower* than the auto grid). Five grids across three families now cluster at 0.1–0.2%; nothing approaches the parked certified grid's 0.84–0.86.
- **Mid-radius resonance stress test** (`toroidal_rotation_factor = 0.2` parks Ω_ℓ=0 nodes at ψ = 0.578, 0.812): naive auto grid vs 1024-knot gold agrees to **1.24e-3** — the profile-curvature term already tracks the reshaped ω_E — and pinning reproduces it to 1e-6. The feared naive-grid physics risk is bounded at the convergence-scatter level for this machine; pinning is cheap insurance plus the step-count win.
- **Collisionless stressor** (nufac = 0.02, layers narrowed ~50×): naive auto grid vs 1024-knot gold again agrees to **1.244e-3** — the same margin as the collisional case. The sharp-resonance risk did not materialize under the strongest stressor constructible on this equilibrium.

## Design provenance

This replaces the parked adaptive approach (#410): detection of *possible* resonances is essentially free from the frequency profiles, so the grid can be made resonance-aware without any kernel-driven refinement loop. The remaining safety-net layers (a free post-hoc residual detector; one bounded probe-and-insert round) are designed and recorded in `handoff/issue376/RESULTS.md` §31 but **not implemented** — every stress test so far says the seeded auto grid is sufficient.

# ⚠️ **REQUIRES THIRD-PARTY HUMAN REVIEW BEFORE MERGING — NON-NEGOTIABLE.**

### DIFF
--- a/.claude/agent-memory/fortran-physics-reviewer/MEMORY.md
+++ b/.claude/agent-memory/fortran-physics-reviewer/MEMORY.md
@@ -5,3 +5,4 @@
 - [KineticForces (NTV) Audit Checklist](kinetic_ntv_map.md) — pentrc->KineticForces map, Logan 2015 matrices, what to verify
 - [InnerLayer (Resistive) Audit Checklist](resistive_layer_map.md) — rmatch->InnerLayer map, GGJ Wasow basis / Δ′, what to verify
 - [Galerkin Δ′ Assembly Map](galerkin_assembly_map.md) — gal.f<->GalerkinAssembly.jl; resonant sign chain verified; PASS
+- [Torque Scalar vs Profile](torque_scalar_vs_profile.md) — scalar toroidal_torque (ported, faithful) vs gpout_dw/dw_matrix psi-profile (unported); why PE vs fgar mismatch isn't necessarily a bug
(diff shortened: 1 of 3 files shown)


logan-nc:
Superseded: the resonance-pinning commit is now part of #414 (cherry-picked to `feature/kinetic-axis-validity`), per the request to combine the two. All evidence in this PR's body is reproduced there. Closing; the branch is left in place.

PR #424, opened 2026-08-22
PerturbedEquilibrium - DOCS - Delineate the boundary-response torque from the KineticForces NTV torque

logan-nc:
Small clarity fix prompted by issue #423's torque cross-validation: `PerturbedEquilibrium/Energies/toroidal_torque` was annotated simply as "net toroidal torque on the plasma", which invites direct comparison with `KineticForces/<method>/total_torque` — a distinct construction.

This PR renames nothing (clean-break renames deferred until the gpout_dw port lands, issue #423) but states the delineation in all three places a user would look: the `PerturbedEquilibriumState` docstring, the HDF5 `long_name`, and the computation site in `Response.jl`. The delineation:

- **Boundary-response torque** (this output): `−2n·Im⟨Φ_tot, Λ⁻¹Φ_tot⟩/4` — the anti-Hermitian boundary quadratic form. Equals the volume-integrated Euler–Lagrange kinetic torque for converged self-consistent solutions (δW surface-term identity); their numerical difference is a solution/grid diagnostic.
- **ψ-resolved EL torque profile** (`gpout_dw`, unported — issue #423): same model, volume-resolved.
- **NTV torque** (`KineticForces/`): independent energy-space bounce-average evaluation; agreement with the above tests whether the EL matrices faithfully discretize the drift-kinetic operator.

## Release note

- **Audience:** users
- **Numerical impact:** none _(harness @ 4a3c163a)_
- **Migration:** none

The `toroidal_torque` output in the perturbed-equilibrium results is now documented as the boundary-response torque, distinguishing it from the separately-computed NTV torque under `KineticForces/`. No output was renamed or moved and no value changed; only the docstring, the HDF5 `long_name` attribute, and a source comment.

## Regression report

```
Regression Report: diiid_n1
=================================================================================================
Ref 1: develop  @ 987e63f3b (2026-08-28)
       env: julia 1.11.7, x86_64-linux-gnu, manifest 7027ae20 (pinned), 112 threads/1 BLAS
Ref 2: local  @ local (2026-09-04)
       env: julia 1.11.7, x86_64-linux-gnu, manifest 7027ae20 (pinned), 112 threads/1 BLAS
-------------------------------------------------------------------------------------------------
Quantity                                      develop          local            Diff       Status
-------------------------------------------------------------------------------------------------
total energy Re(et[1])                        8.012318e-01     8.012318e-01     0.0e+00    OK
total energy Im(et[1])                        4.142529e-05     4.142529e-05     0.0e+00    OK
[...]

Issue #423, opened 2026-08-21
Port gpout_dw / gpout_dw_matrix (psi-resolved torque response profiles and matrices) as the grid-logic diagnostic


logan-nc:
## Summary (revised after fortran-physics-reviewer consultation)

**We need to port the ψ-resolved torque response machinery from `gpec/gpout.f` — `gpout_dw` (lines 2579–2650) and `gpout_dw_matrix` (2655–2891) — and use it as the standing grid-logic diagnostic.**

The Julia PE currently outputs only the **scalar boundary torque** `Energies/toroidal_torque = −2n·Im(⟨Φ_tot, Λ⁻¹Φ_tot⟩/4)`. A physics review against the Fortran confirms this scalar is a **faithful port** of gpout.f:1195–1201 (identical quadratic form, /4 normalization, total-flux argument, torque factor). But it is a *boundary quadratic form of the inductance matrix* — not the volume-resolved quantity:

- **`gpout_dw`**: ψ-resolved torque/energy density from the actual self-consistent EL solutions, `dW(ψ) = Σ conj(u1)·u2 / 2 / (2μ₀) · 2n·i` per step, splined; Re of the running ψ-integral = torque (N·m), Im = 2nδW.
- **`gpout_dw_matrix`**: the mpert×mpert response-matrix generalization per ψ step (dwks/dwk/gind/gres/gindp/gresp; normalized by ξ, Φ_boundary, Φ_x, and coordinate-independently).

Neither has a Julia equivalent (grep confirms). **The ψ-integrated `gpout_dw` torque is the true apples-to-apples comparator for the grid-free `KineticForces/fgar` NTV torque** — the comparison that in Fortran exposed ldp-grid inadequacy.

## Measured data motivating the port (DIII-D kinetic-calculated, suppression on)

| run | PE scalar torque [N·m] | fgar torque [N·m] | PE/KF |
|---|---|---|---|
| rot factor 1.0, auto grid (290) | 0.1831 | 0.7241 | 0.25 |
| rot factor 1.0, fixed 1024 | 0.0830 | 0.7281 | 0.11 |
| rot factor 0.2, auto grid (305) | 0.0946 | 0.1282 | 0.74 |
| rot factor 0.2, fixed 1024 | 0.0950 | 0.1286 | 0.74 |

Per the review, the scalar-vs-fgar gap is **not by itself a bug** — the two are different constructions — so these rows are motivation, not defect evidence. Two genuine flags remain:

[...]

PR #425, opened 2026-08-22
KF/PE - BUGFIX! - Self-consistent xi_s for the NTV kernel input; static scheduling for the kinetic kernel loop

logan-nc:
Stacked on #422 (`feature/kinetic-resonance-knots`). The overnight resolution of the PE-vs-NTV torque discrepancy (issue #423's measured motivation; RESULTS.md §34–§35).

## The bug

In kinetic runs, the Clebsch ξ^α fed to the NTV kernel (via `dbob_m`/`divx_m`) was **re-solved from the ideal A, B, C matrices** — inconsistent with the self-consistent kinetic tangential response exactly where resonant layers live. Fortran (`gpeq.f`, `kin_flag`) instead scales the **stored self-consistent ξ_s** by the singfac regularization factor and inverts nothing. This is the same physics flaw #407 fixes on the MatrixSplines branch (there: a silently broken Cholesky of the non-Hermitian kinetic A; here: a *valid* Cholesky of the *wrong* — ideal — A).

## The consequence (measured, DIII-D kinetic-calculated full chain)

The perturbative NTV evaluation was driven by a tangential displacement that ignores the kinetic layer response, producing a **phantom 4× torque** concentrated in a ~2e-3-wide resonant core at the pedestal ω_E crossing (fgar: 0.7241 N·m, of which +0.622 in ψ∈[0.90, 0.95]; the self-consistent EL solution collects 0.046 there — and the EL side was verified faithful at every level: boundary ≡ volume torque to 7 digits, FKG assembly and kinetic RHS sign-by-sign Fortran-conformant per fortran-physics-reviewer, matrix-spline resolution and integrator step size both exonerated by direct null experiments).

With the fix, the phantom collapses (+0.622 → +0.033) and the two independent torque calculations agree band-by-band (resonant layer: 0.0461 vs 0.0455). Domain-matched totals (ΔIm over [ψ_c, psilim] — fgar's exact domain; the sub-ψ_c Im(u1†u2) offset is non-dissipative initial-condition content, ±0.05 with opposite signs across cases):

| case (deterministic, post-:static) | PE (domain-matched) | fgar | agreement |
|---|---|---|---|
| rot factor 0.2, reg_spot 0.05 | 0.1456 | 0.1460 | **0.3%** |
| rot factor 1.0, reg_spot 0.05 | 0.1322 | 0.1655 | 20% |
| rot factor 1.0, reg_spot 0.01 | 0.1322 | 0.1397 | 5.4% |

Where the comparison is regularization-unambiguous (rot=0.2: resonances away from rationals), the two independent torque calculations now agree to **0.3%**. The rot=1.0 residual sits inside fgar's own reg_spot sensitivity band (its answer moves 18% between reg 0.05 and 0.01 — the acknowledged near-rational input ambiguity, present in Fortran too); PE is reg-independent by construction.

## Second fix in this PR: determinism

`CalculatedKineticMatrices.jl`'s per-surface kernel loop was the **only** threaded loop in the codebase using `Threads.@threads` without `:static` while indexing per-thread scratch by `threadid()` — dynamic-scheduling task migration corrupts kinetic matrix rows at rare-migration frequency. Observed live: three runs of one identical deck gave et[1] = 1.0053 / 1.1355 / 0.9895. With `:static`, identical decks reproduce **bit-identically** (verified), and et returns exactly to the pre-fix value — every historical calculated-kinetic result carries this latent race.

**Effects on outputs**: kinetic-run regularized quantities (`xi_clebsch_alpha`, `*_reg` fields) and every NTV torque driven by them change; ideal runs and `reg_spot = 0` are bit-unaffected. Kinetic harness cases will move — new baselines with this physics justification.

## Review items

- Kinetic-ness is currently detected via populated `kwmats` splines (sentinel = 5 knots); reviewers may prefer threading an explicit flag from `kinetic_factor > 0` — noted as the clean alternative.
- The domain-matched comparison rule (and the eventual `gpout_dw` ψ-profile port, issue #423) is where the "which torque is which" delineation (#424) completes.

### DIFF
--- a/src/KineticForces/CalculatedKineticMatrices.jl
+++ b/src/KineticForces/CalculatedKineticMatrices.jl
@@ -120,7 +120,7 @@ function compute_calculated_kinetic_matrices(
     thread_block_w = [zeros(ComplexF64, mpert, mpert, 6) for _ in 1:nthreads]
     thread_block_t = [zeros(ComplexF64, mpert, mpert, 6) for _ in 1:nthreads]
 
-    Threads.@threads for ipsi in 1:mpsi
+    Threads.@threads :static for ipsi in 1:mpsi
         tid = Threads.threadid()
         intr_t = thread_intrs[tid]
         full_w = thread_full_w[tid]
--- a/src/PerturbedEquilibrium/FieldReconstruction.jl
+++ b/src/PerturbedEquilibrium/FieldReconstruction.jl
@@ -365,6 +365,25 @@ function compute_clebsch_displacements(
         return clebsch_psi, clebsch_psi1, clebsch_alpha
     end
 
+    # Kinetic runs: mirror Fortran gpeq.f kin_flag — regularize the STORED self-consistent ξ_s
+    # by the same singfac factor and never re-solve from matrices. The ideal-matrix resolve is
+    # inconsistent with the kinetic solution's tangential response (the resonant layers live in
+    # exactly that dynamics), and the kinetic A is non-Hermitian and must not be re-inverted
+    # here. Kinetic-ness detected by populated kwmats splines (sentinel has 5 knots).
+    if length(ffit.kwmats[1].cache.x) > 8
+        for ipsi in 1:npsi
+            q = equil.profiles.q_spline(psi_grid[ipsi])
+            for ipert in 1:mpert
+                m = mlow + ipert - 1
+                singfac = m - nn * q
+                reg_factor = singfac^2 / (singfac^2 + reg_spot^2)
+                clebsch_psi1[ipsi, ipert] = xi_psi1_modes[ipsi, ipert] * reg_factor
+                clebsch_alpha[ipsi, ipert] = xi_s_modes[ipsi, ipert] * reg_factor / chi1
+            end
+        end
+        return clebsch_psi, clebsch_psi1, clebsch_alpha
+    end
+
     # Per-thread workspaces: matrix ops and spline hints are not safe to share across threads.
     # Size by maxthreadid() and index by threadid() under :static scheduling (GPEC convention).
     nt = Threads.maxthreadid()


logan-nc:
**Closing without merging — both halves are superseded, and one of my claims here needs correcting.**

**1. The `:static` fix is already on develop.** Commit `9fe8a60c9` ("KineticForces - IMPROVEMENT - :static kinetic-matrix threading + typed species + shared constants") added exactly this scheduling fix, with the same reasoning in its comment. It landed *after* this stack forked (our fork point `9491f893d` is dated 2026-08-16), so this branch never had it — which is why the nondeterminism was observable here.

**Correction to this PR's description**: it claimed "every historical calculated-kinetic run carries this latent race." That is wrong as written. The race is real and was real on *this stack*, but develop has been fixed since `9fe8a60c9`; only runs on branches forked before it (like this one) are affected. The observed scatter — et[1] = 1.0053 / 1.1355 / 0.9895 on identical decks — was a symptom of the stack being behind develop, not of a live defect in develop.

**2. The ξ_s self-consistency fix is owned by #407.** Same file, same function, same physics (kinetic runs must regularize the stored self-consistent ξ_s rather than re-solve it from the ideal matrices, mirroring Fortran `gpeq.f` `kin_flag`). #407 is based on develop, already reviewed, and additionally adds `[ForcingTerms]`/`[PerturbedEquilibrium]` to `Solovev_kinetic_calculated_example` plus tracked harness quantities — the coverage this branch lacked. No reason to ask anyone to reconcile two versions of one fix.

**Worth flagging for #407 and #414 reviewers**: #414 forces `reg_spot = 0` in self-consistent kinetic runs, and `compute_clebsch_displacements` early-returns when `reg_spot == 0` *before* the kinetic branch. So once #414 lands, the kinetic branch #407 fixes is unreachable on that path — still correct, still the right safety net if regularization is ever re-enabled, but no longer load-bearing. The physics evidence for turning regularization off is in #414 and in `handoff/issue376/RESULTS.md` §36–§37.

The measurement this PR produced still stands and is recorded in #414: with regularization off, the NTV torque and the Euler–Lagrange solution's own dissipation agree to **0.15%** (0.1324 vs 0.1322 N·m) on the DIII-D kinetic-calculated case.

Branch `bugfix/kinetic-clebsch-xi-s` is left in place for the record.

PR #426, opened 2026-08-22
PE/EQUIL - BUGFIX! - Withhold unvalidated kinetic tearing diagnostics and free the grid they constrained

logan-nc:
Stacked on #425. Two coupled changes, both driven by the same finding: **GPEC has no validated kinetic Δ′.**

## 1. Withhold the unvalidated kinetic tearing diagnostics

`Delta_prime`, `C_Delta_prime`, `resonant_current`, `resonant_area_weighted_field`, `island_width_sq`, `island_half_width` and `chirikov_parameter` all descend from one quantity: the outer-solution jump in ∂b^ψ/∂ψ across the rational surface, read as a tearing index through **ideal** outer-region asymptotic matching at an **ideal** rational surface. Kinetic terms displace the surfaces (`kinsing`) and change the layer response, so that reading has no validated basis. In kinetic runs these datasets are now withheld with a warning naming exactly what is missing and why; surface metadata (ψ, q, m, n, area) and the inner-layer penetrated field are kept, and **all ideal runs are bit-unaffected**.

Opt-in `[PerturbedEquilibrium] unvalidated_kinetic_tearing = true` restores them for research use, with a warning that they must not be published without independent verification.

(The Riccati `Delta_prime_matrix` path was already unreachable in kinetic runs — kinetic requires the forward integrator — so PE's singular coupling was the only leak.)

## 2. Free the grid that constraint was paying for

Rational-surface bracketing exists **solely** to keep the ideal Δ′ stencil's 3rd derivative consistent across the surface (`GridRefinement.jl`). When that Δ′ is withheld, the bracketing is skipped and the fine near-rational knots survive for the kinetic matrices instead; opting back into the diagnostics restores the bracketing, so the grid always matches the physics being delivered.

**Measured effect on the torque cross-check: none** (DIII-D rot=1.0, deterministic): PE 0.1322 / fgar 0.1655 before and after, grid 290 → 288 knots. The bracket's cleared zone is only ~2e-4 wide and already contained 17–18 knots, so this is a cleanup of a now-purposeless constraint, not a fix — reported as the null result it is.

## What the residual actually was

Chasing it produced the more useful result. The PE-vs-NTV gap on rot=1.0 is concentrated at the resonant layers — at ψ = 0.9933 fgar collects +0.0358 within ±0.002 while the EL solution collects +0.0026 — and it is a **formulation difference amplified by a numerical smoothing width**:

| | PE (EL dissipation) | fgar | full domain | layers excluded (±0.002) |
|---|---|---|---|---|
| reg_spot = 0.05 | 0.1322 | 0.1666 | 20% | 5.2% |
| reg_spot = 0.01 | 0.1322 (identical) | 0.1402 | **5.7%** | **0.8%** |

PE is exactly `reg_spot`-independent (it is the ODE's own dissipation); fgar's layer contribution collapses 0.0453 → 0.0138 as the regularization narrows. At a rational surface the resonant torque enters the outer-region description as a **surface term — the jump / resonant current**, i.e. precisely the quantity part 1 withholds as unvalidated, while fgar integrates a reg-broadened volume approximation of it. So the two agree wherever that term is absent (0.8%), and cannot agree on the full domain until a *validated* kinetic layer formulation supplies it (issue #423).

**Practical guidance this establishes**: compare PE and NTV torque either outside the resonant layers or at small `reg_spot`; treat any full-domain difference at large `reg_spot` as the smoothing width, not physics.

## Regression harness — and an honest coverage gap

`regress --cases diiid_n1,solovev_kinetic_calculated,solovev_kinetic_ntv --refs bugfix/kinetic-clebsch-xi-s,local`: **47 / 14 / 6 unchanged** — nothing moved anywhere.

That clean result is partly a coverage gap, and should be read as one: the Solovev kinetic decks run **fixed** `mpsi` (16 and 128), so they never enter the two-pass auto grid where the bracketing lives, and they track no Δ′-family quantity, so nothing went missing. The harness therefore confirms only that ideal runs are bit-identical (47/47) and that this PR is inert on fixed-grid kinetic decks. The bracketing change and the withheld datasets are exercised **only** by the DIII-D kinetic-calculated auto-grid runs reported above. This is the third time this configuration has been the sole coverage for a change — a DIII-D kinetic-calculated FFS-only harness case remains the outstanding test-infrastructure item (see #398's disclosure and #423).

# ⚠️ **REQUIRES THIRD-PARTY HUMAN REVIEW BEFORE MERGING — NON-NEGOTIABLE.**

### DIFF
--- a/src/Equilibrium/GridRefinement.jl
+++ b/src/Equilibrium/GridRefinement.jl
@@ -418,7 +418,7 @@ Build the refined pass-2 ψ grid from a formed pass-1 equilibrium: measured-curv
 density (`_knot_density`), equidistribution, a global minimum-spacing floor
 (`enforce_min_spacing`), and rational-surface bracketing (`bracket_mandatory_nodes`). `tau` is
 the target interpolation accuracy (`psi_accuracy`); `kin` optionally supplies kinetic profiles
-whose pedestal gradients attract knots; `mandatory` lists rational-surface ψ values to bracket; `pinned` lists ψ values inserted as plain knots without a cleared zone (kinetic-resonance surfaces);
+whose pedestal gradients attract knots; `mandatory` lists rational-surface ψ values to bracket (`bracket_mandatory = false` skips the bracketing and keeps the fine knots inside instead); `pinned` lists ψ values inserted as plain knots without a cleared zone (kinetic-resonance surfaces);
 `singfac_min` and `n_min` (smallest |n| in the run) set each surface's matching half-stencil
 `dpsi = singfac_min/(n_min·|q′|)`, and the bracket half-width is `bracket_coef·dpsi` (floored at
 `min_spacing`). Rational surfaces are bracketed, not pinned: a knot on the surface would make the
@@ -429,6 +429,7 @@ function refined_psi_grid(equil::PlasmaEquilibrium;
     kin::Union{Nothing,KineticProfileSplines}=nothing,
     mandatory::Vector{Float64}=Float64[],
     pinned::Vector{Float64}=Float64[],
+    bracket_mandatory::Bool=true,
     singfac_min::Float64=1e-4,
     n_min::Int=1,
     bracket_coef::Float64=BRACKET_COEF,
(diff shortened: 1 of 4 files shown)


logan-nc:
Closing and deleting: overstepped. The PE singular-coupling Δ′ is valid and must not be withheld — the untrusted-with-kinetics quantity is the **Riccati** Δ′, which is already unreachable in kinetic runs (kinetic requires the forward integrator, hard error at `EulerLagrange.jl:159`), so no guard was needed. The grid half of this PR is dropped with it; its measured result stands on the record as a null (removing the bracketing changed the PE-vs-NTV torque by nothing: 0.1322/0.1655 before and after).

PR #436, opened 2026-08-31
Vacuum - FEATURE - Add 3D conformal conducting wall capabilities

jhalpern30:
## Release note

- **Audience:** users
- **Numerical impact:** none on existing results _(harness @ 25a1096)_
- **Migration:** none

The 3D vacuum solver can now put a conducting wall around a non-axisymmetric boundary. With
`nzeta_in > 1`, `shape = "conformal"` builds the wall by displacing every plasma point a fixed
distance along its own normal. Previously any 3D-input case — stellarator or 3D tokamak alike —
could only be run with `shape = "nowall"`.

**Accuracy caveat for the new path.** A 3D walled result is only trustworthy while the plasma–wall
gap spans at least one grid cell of the *coarser* direction, which is toroidal on every stellarator
grid tested. Since `h_ζ ≈ 2πR₀/nzeta`, that is a real constraint: on a 24×24 grid with `R₀ = 1.7 m`
the gap must exceed ~0.45 m. The constructor warns below one cell and names the direction to refine.
"Numerical impact: none" above refers to existing 2D results, which are untouched — it is not a
claim that any particular 3D walled configuration is converged.

## Regression report

```
regress --cases solovev_n1,diiid_n1 --refs f36ecef8,25a1096
```

```
Case: solovev_n1 — Solovev analytical equilibrium, n=1, ideal stability

Regression Report: solovev_n1
=================================================================================
Ref 1: f36ecef8  @ f36ecef86 (2026-09-11)
       env: julia 1.12.6, arm64-apple-darwin24.0.0, manifest ece8f5d1 (pinned), 6 threads/6 BLAS
Ref 2: 25a1096  @ 25a1096fd (2026-09-15)
       env: julia 1.12.6, arm64-apple-darwin24.0.0, manifest ece8f5d1 (pinned), 6 threads/6 BLAS
   [per-quantity table omitted — every row OK]
=================================================================================
Summary: 22 unchanged

Case: diiid_n1 — DIII-D-like equilibrium, n=1, ideal + perturbed equilibrium

Regression Report: diiid_n1
Ref 1: f36ecef8  @ f36ecef86 (2026-09-11)
       env: julia 1.12.6, arm64-apple-darwin24.0.0, manifest ece8f5d1 (pinned), 6 threads/6 BLAS
Ref 2: 25a1096  @ 25a1096fd (2026-09-15)
       env: julia 1.12.6, arm64-apple-darwin24.0.0, manifest ece8f5d1 (pinned), 6 threads/6 BLAS
   [per-quantity table omitted — every row OK]
Summary: 47 unchanged
```

Ref 1 is the tip of develop, Ref 2 the head of this branch, so the comparison isolates this PR's own
changes from the develop commits merged in. Every tracked quantity identical (`Diff 0.0e+00`,
checksums identical) — expected, since the only `src/` changes here are in the 3D wall path.

### What to look at

- `WallGeometry3D` now takes `plasma_surf` and selects the shape with the same `if`/`elseif`
  chain `WallGeometry` uses in 2D: axisymmetric revolves the 2D contour, otherwise conformal,
  otherwise the same "shape not recognized" error the 2D reader gives. One sequential function,
  no helper indirection.
- The offset: `r_wall = r_plasma − d·n̂_plasma` with `d = a·r_minor` and
  `r_minor = ½(max R − min R)` — the same definition the 2D shapes use, which is what makes the
  two routes comparable on an axisymmetric boundary. On a non-axisymmetric boundary the extrema
  are taken over the whole torus, so `a` scales with half the global R-extent rather than a
  cross-section minor radius; this is documented on `WallShapeSettings.a`.
- The preconditions, both checked before the `nowall` early return: `plasma_surf` must be on the
  same `mtheta`/`nzeta` grid as the wall (the two surfaces are indexed together throughout), and
  `nfp == 1` (`expand_field_periods` tiles a non-axisymmetric boundary to the full torus and
  returns `nfp = 1`, so `nfp > 1` can only arrive from an axisymmetric boundary).
- The guards: fold (error), near-fold (warn), gap under one grid cell (warn), `equal_arc_wall`
  ignored (info). There is deliberately **no** centre-stack clamp — clamping a 3D surface would
  leave a cusp and corrupt the spline normals the near-field quadrature depends on, so it is
  rejected instead. There is also no separate machine-axis check: for a conformal offset the
  inboard toroidal radius of curvature *equals* the distance to the axis, so the fold guard fires
  exactly where the wall would reach R = 0 (measured: fold at `a = 3.9`, axis at `a = 3.857`).
  Both fold tests are local, so a global self-intersection between distant lobes is invisible to
  them; that limitation is stated in the docstring.

### Verification

| gate | what it tests | result |
|---|---|---|
| A | same Solovev plasma through both wall routes, vs a converged 2D solve | 4th-order convergence on both routes; their difference converges away faster still |
| B | W7-X wall geometry | offset error ≤ 6.7e-16 m, nfp-periodicity ≤ 4.2e-14 m, index alignment exact |
| C | W7-X `Wᵛ` vs wall distance, two grids | smooth and monotone to ~0.08 m; 1–3 % self-convergence there |
| D | Solovev `Wᵛ` vs wall distance, against a gap-independent 2D reference | both routes fail identically — the quadrature, not the wall construction |
| E | precise-QA boundary, three-grid Richardson ladder | geometry exact; the threshold reproduces on geometry that had no part in deriving it |

Unit tests 277 → 316. `docs/make.jl` builds clean.

The offset branch now has response-level coverage: a circular cross-section is checked against its
exact analytic wall (a circle of radius `r_minor(1 + a)` in every plane, matched to 9e-5 m), and
`compute_vacuum_response` is run on a non-axisymmetric conformal wall for finiteness, Hermiticity,
and the requirement that `Wᵛ` exceed the nowall value.

### Known limitation, now warned about

Accuracy is set by the plasma–wall gap **measured in grid cells of the coarser direction**
(toroidal, on every stellarator grid tested): at or above 1 cell the wall gap is no longer the
limiting error; between 0.3 and 1 cell the answer is one to two orders worse; below 0.3 cells it
is meaningless. The criterion is evaluated pointwise, so an under-resolved patch is not averaged
away by the rest of the surface, and the warning names the worst-resolved point. Note this says
when to stop blaming the wall, *not* that the answer is converged — what remains above one cell is
ordinary surface discretization and is geometry-dependent.

JMH - I am going to try to fix this in a future PR, but for now will just warn. This is not an inherently 3D issue. It also appears in 2D, but we can typically increase the resolution to be large enough that it isn't an issue.

image

image


src/Vacuum/DataTypes.jl
```diff
@@ -198,6 +198,9 @@ Struct containing input settings for vacuum wall geometry.
         + `"mod_dee"`: Modified Dee-shaped wall
         + `"filepath"`: Custom wall shape from the file you specify
 
+    A non-axisymmetric boundary (`nzeta_in > 1`) supports `"nowall"` and `"conformal"`; the others
+    are poloidal contours that get revolved and so need `nzeta_in == 1`.
+
   - `a::Float64`: Distance of wall from plasma in units of major radius (conformal), or shape parameter (others)
```
d-burg: NIT: Two fixes : (1) Says "in units of major radius", hasn't the code always used minor radius? (2) From claude: "`a` silently changes meaning in 3D: `0.5*(max R - min R)` is taken over the whole torus, so it's the minor radius when axisymmetric but half the global R-extent on a stellarator (~2x the minor radius for W7-X). So `a=0.2` calibrated on a tokamak gives about twice the intended gap. Worth stating explicitly."

src/Vacuum/DataTypes.jl
```diff
@@ -627,21 +632,36 @@ struct WallGeometry3D
...
 
   - `WallGeometry3D`: Struct containing wall surface coordinates and derivatives
+
+# Notes
+
+  - Axisymmetric boundaries (`nzeta_in == 1`) support nowall, conformal, elliptical, dee, mod_dee and
+    from_file; non-axisymmetric boundaries support nowall and conformal
+  - The conformal wall for a non-axisymmetric boundary displaces each plasma point along its own normal,
+    which keeps wall grid index `(i, j)` the closest wall point to plasma index `(i, j)` — the
+    correspondence the near-field patch of `compute_3D_kernel_matrices!` assumes
```
d-burg: SHOULD FIX (from claude): True for the conformal branch, not for the revolved one below it — there the wall sits on the geometric angle phi while PlasmaGeometry3D puts the plasma on zeta = phi - nu(theta), plus equal_arc_wall re-parameterizes poloidally. Since this is now the stated contract for the near-field patch, say which branch satisfies it.

src/Vacuum/DataTypes.jl
```diff
@@ -676,16 +696,35 @@ function WallGeometry3D(inputs::VacuumInput, wall_settings::WallShapeSettings)
...
 
-    inputs.nzeta_in > 1 && error("3D wall geometry not yet implemented for non-axisymmetric inputs")
-    inputs.nfp > 1 && error("3D wall geometry not yet implemented for nfp > 1")
-
-    # Plasma surface coordinates (2D)
-    surf_2D = PlasmaGeometry(inputs)
-    wall_2D = WallGeometry(inputs, surf_2D, wall_settings)
+    # expand_field_periods leaves an axisymmetric boundary alone, so nfp > 1 can only arrive here with
+    # per-period counts that the rest of the 3D solve reads as full-torus ones.
+    inputs.nfp > 1 && error("3D wall geometry requires a full-torus boundary. Call expand_field_periods(inputs) first, or supply the whole torus with nfp = 1.")
```
d-burg: NIT: The message says to call expand_field_periods first, but the caller already does and it looks like a no-op for nzeta_in == 1... so an axisymmetric boundary with nfp > 1 and a wall dies with advice that can't help? Also from claude: "the "per-period counts" comment is also wrong there: PlasmaGeometry3D extrudes over the full 2pi regardless of nfp. Suggest mirroring expand_field_periods' own condition:
`inputs.nfp > 1 && inputs.nzeta_in > 1`.

Out of scope but noticed here: with nowall the early return skips this guard, and then num_points = mtheta*nzeta*nfp no longer matches the mtheta*nzeta rows PlasmaGeometry3D built. Looks like a live bug on develop — worth an issue." Maybe follow up on that bug?

src/Vacuum/DataTypes.jl
```diff
@@ -676,16 +696,35 @@ function WallGeometry3D(inputs::VacuumInput, wall_settings::WallShapeSettings)
...
+    if inputs.nzeta_in == 1
+        # Axisymmetric boundary: build the 2D poloidal contour and revolve it toroidally
+        surf_2D = PlasmaGeometry(inputs)
+        wall_2D = WallGeometry(inputs, surf_2D, wall_settings)
+        for i in 1:mtheta, (j, ϕ) in enumerate(ϕ_grid)
+            r[i+mtheta*(j-1), :] .= [wall_2D.x[i] * cos(ϕ), wall_2D.x[i] * sin(ϕ), wall_2D.z[i]]
+        end
+    elseif wall_settings.shape == "conformal"
+        # Displace every plasma point outward along its own normal
+        wall_settings.equal_arc_wall && @warn "equal_arc_wall is ignored for non-axisymmetric (nzeta_in > 1) walls: it re-parameterizes a 2D contour and would break the plasma/wall index alignment the near-field patch relies on."
```
d-burg: NIT: equal_arc_wall defaults to true, so it looks like every stellarator run warns about a setting the user never chose. Warn only when explicitly set, or soften to @info

src/Vacuum/DataTypes.jl
```diff
@@ -676,16 +696,35 @@ function WallGeometry3D(inputs::VacuumInput, wall_settings::WallShapeSettings)
...
+
+        # Same logic as 2D
+        R_plasma = [hypot(plasma_surf.r[idx, 1], plasma_surf.r[idx, 2]) for idx in axes(plasma_surf.r, 1)]
+        offset_gap = wall_settings.a * 0.5 * (maximum(R_plasma) - minimum(R_plasma))
```
d-burg, line 747: NIT: max/min over the whole torus is what makes `a` mean something different on a 3D boundary, see the previous note on the `a` docstring

src/Vacuum/DataTypes.jl
```diff
@@ -676,16 +696,35 @@ function WallGeometry3D(inputs::VacuumInput, wall_settings::WallShapeSettings)
...
+        @info "Calculating conformal wall shape $((@sprintf "%.2e" offset_gap)) m from plasma surface."
+
+        # Plasma normal points into the plasma, so the displacement is along -normal
+        @inbounds for idx in axes(plasma_surf.r, 1)
```
d-burg: BLOCKING (from claude): This iterates axes(plasma_surf.r, 1) but writes into `r`, sized inputs.mtheta * inputs.nzeta. The fold loop below does the reverse. Nothing enforces that plasma_surf was built from these same inputs, and under @inbounds a mismatch is an out-of-bounds *write*, not an error. compute_vacuum_response passes a matching pair today, but this is now a two-surface constructor with a silent invariant. One line at the top:

    @assert size(plasma_surf.r, 1) == num_points "plasma_surf must be on the same mtheta/nzeta grid"

src/Vacuum/DataTypes.jl
```diff
@@ -705,6 +744,42 @@ function WallGeometry3D(inputs::VacuumInput, wall_settings::WallShapeSettings)
     normal_orient = normal[idx, 1] > 0 ? 1 : -1
     @views normal .*= normal_orient
 
+    # Same-index plasma–wall separation: the near-field patch is centred on this pair.
+    gap = minimum(hypot(r[idx, 1] - plasma_surf.r[idx, 1], r[idx, 2] - plasma_surf.r[idx, 2], r[idx, 3] - plasma_surf.r[idx, 3]) for idx in axes(r, 1))
```
d-burg: SHOULD FIX: This uses exactly the pairing the fold check two lines down refuses to trust ("equal_arc_wall re-parameterizes and breaks that pairing"), but unconditionally.

Measured this with claude on the axisymmetric route (R0=1.7, r_minor=0.3, a=0.3, nzeta=32): same-index gap 0.09032 vs true min surface-to-surface distance 0.08609, 4.9% high at nu = 0. C: "Worse with nu != 0, where the pair is at a different toroidal location. It errs high, so the sub-cell warning is missed exactly when it matters. Gate gap/cells on the branch where the pairing holds, or compute a real min separation."

src/Vacuum/DataTypes.jl
```diff
@@ -705,6 +744,42 @@ function WallGeometry3D(inputs::VacuumInput, wall_settings::WallShapeSettings)
...
+        @inbounds for idx in axes(normal, 1)
+            norm_wall = sqrt(normal[idx, 1]^2 + normal[idx, 2]^2 + normal[idx, 3]^2)
+            norm_plasma = sqrt(plasma_surf.normal[idx, 1]^2 + plasma_surf.normal[idx, 2]^2 + plasma_surf.normal[idx, 3]^2)
+            # Plasma n in, wall n out of vacuum → healthy offset dots to −1; store −cosine so +1 is aligned.
+            dot_wp = normal[idx, 1] * plasma_surf.normal[idx, 1] + normal[idx, 2] * plasma_surf.normal[idx, 2] + normal[idx, 3] * plasma_surf.normal[idx, 3]
+            min_align = min(min_align, -dot_wp / (norm_wall * norm_plasma))
+            min_area_ratio = min(min_area_ratio, norm_wall / norm_plasma)
+        end
+
+        min_align <= 0.0 && error(
```
d-burg, line 796: BLOCKING (from claude): The PR body lists "machine-axis crossing (error, reports the largest admissible `a`)" as a guard, but there's no R-check in the 3D path — grep for "admissible" only hits InnerLayer/GGJ/Ray.jl. 2D has both a clamp and a hard `any(x_wall .<= 0.0) && error(...)`. Restore it or drop it from the description.

I checked the cost and it's low: on a corrugated torus with min R_plasma = 1.4, a=2.0 builds at min R_wall = 0.65 and a=4.0 errors *here* — the fold guard fires right where the axis is reached, since the inboard toroidal curvature radius is the distance to the axis. So this is "fix the description or add the cheap check", not a correctness hole.

Real residual: this test is local, so global self-intersection between distant lobes (a bean cross-section offsetting into itself) has aligned normals and positive area element and is invisible to both checks. One docstring sentence would cover it.

src/Vacuum/DataTypes.jl
```diff
@@ -705,6 +744,42 @@ function WallGeometry3D(inputs::VacuumInput, wall_settings::WallShapeSettings)
...
+            "Conformal wall offset of $((@sprintf "%.2e" gap)) m self-intersects: it exceeds the local concave radius " *
+            "of curvature of the plasma surface somewhere, folding the offset surface. Reduce the wall distance a."
+        )
+        (min_align < 0.5 || min_area_ratio < 0.5) &&
```
d-burg, line 800: NIT: min_align < 0.5 is 60 degrees, this seems loose for a strongly shaped boundary? You might expect nuisance warnings on real stellarators well before anything is near folding, what do you think?

src/Vacuum/DataTypes.jl
```diff
@@ -705,6 +744,42 @@ function WallGeometry3D(inputs::VacuumInput, wall_settings::WallShapeSettings)
...
+            @warn "Conformal wall is close to folding: min(n̂_wall·n̂_plasma)=$((@sprintf "%.3f" min_align)), min area-element ratio=$((@sprintf "%.3f" min_area_ratio)). Reduce the wall distance a."
+    end
+
+    # The 3D singular quadrature resolves the double-layer kernel well only while the plasma-wall gap spans a grid cell or more
+    h_θ = max(sqrt(sum(abs2, dr_dθ) / num_points), sqrt(sum(abs2, plasma_surf.dr_dθ) / num_points)) * dθ
+    h_ζ = max(sqrt(sum(abs2, dr_dζ) / num_points), sqrt(sum(abs2, plasma_surf.dr_dζ) / num_points)) * dζ
+    cells = gap / max(h_θ, h_ζ)
+    if cells < 1.0
```
d-burg: SHOULD FIX (from claude): Two issues.

(1) Mixes a pointwise minimum (gap) with a global RMS (h_theta, h_zeta). Make it pointwise: minimum(gap_idx / max(h_theta_idx, h_zeta_idx)). Three lines, strictly sharper, catches the inboard under-resolution the RMS averages away.

(2) The criterion looks unsatisfiable at realistic resolution. Running the repo's own existing 3D conformal config (R0=1.7, r_minor=0.3, a=0.3, nzeta=32):

    Warning: Plasma-wall gap 9.03e-02 m spans only 0.27 cells of the coarser grid
             (h_theta=6.10e-02 m, h_zeta=3.40e-01 m)

The already-merged, CI-green test sits at 0.27 cells — inside the range this PR calls meaningless. Since h_zeta ~ 2*pi*R0/nzeta, one cell needs nzeta >~ 120.

Is *any* currently tested 3D walled config in the accurate regime? If not that belongs in the release note, not a footnote — "Numerical impact: none" is true of the harness but misleading about 3D walled results. And a warning that fires on nearly every run stops being read.

test/runtests_vacuum.jl
```diff
@@ -614,6 +618,111 @@
...
+            wall = GeneralizedPerturbedEquilibrium.Vacuum.WallGeometry3D(inputs, plasma, settings)
+
+            num_points = 24 * 24
+            @test wall.nowall == false
+            @test size(wall.r) == (num_points, 3)
+            @test all(isfinite, wall.r) && all(isfinite, wall.normal)
+
+            # Uniform offset along the plasma normal: every point moves the same distance, outward.
+            offsets = [norm(wall.r[i, :] - plasma.r[i, :]) for i in 1:num_points]
+            @test maximum(offsets) - minimum(offsets) < 1e-12
```
d-burg, line 654: SHOULD FIX (from claude): True by construction — scale = -offset_gap / ||n||, so offsets are identical whatever the normal direction or magnitude. Same for `all(aligns .< 0)` below, which restates the fold guard's own condition. Neither fails if the offset is wrong in a correlated way.

Non-circular alternative, ~5 lines: a circular cross-section torus through the 3D branch (nzeta_in > 1, epsilon = 0) has an exact wall — a circle of radius r_minor + d in every plane. Pins direction, magnitude and orientation at once.

test/runtests_vacuum.jl
```diff
@@ -614,6 +618,111 @@
...
+            @test all(hypot.(wall.r[:, 1], wall.r[:, 2]) .> 0)
```
d-burg: BLOCKING: Seems like it can never fail: hypot is non-negative by construction, and 0 only if x and y are both exactly 0.0. It can't detect axis crossing, which I assume is the intent? Claude suggests a signed test that does:

    @test all(wall.r[i,1]*plasma.r[i,1] + wall.r[i,2]*plasma.r[i,2] > 0 for i in 1:num_points)

d-burg:
Some final thoughts from my+claude's review:

One extra comment: 
BLOCKING: test coverage (couldn't attach inline): the `compute_vacuum_response 3D conformal wall` testset uses _make_3d_inputs, which is nzeta_in = 1. Therefore it runs the revolved branch, not the new one. And _make_3d_periodic_inputs is never passed to compute_vacuum_response anywhere.

So of the 30 new tests, none compute Wv on a non-axisymmetric wall, all new coverage is WallGeometry3D construction. With no 3D regression case either, the new numerical path has zero automated protection – it looks like gates A-E exist only as plots in the description. Do you agree? Claude suggested adding one compute_vacuum_response call on _make_3d_periodic_inputs with shape="conformal"? Finiteness and Hermiticity alone would catch a regression. What do you think?

Good work overall and the structure looks good: the branch chain mirrors 2D cleanly, normal alignment + area element ratio is a good pair of fold diagnostics, the offset preserves nfp-periodicity so the block reduction stays valid, and the accuracy limitation is reported.

Blocking right now: the missing end-to-end test, the unguarded @inbounds indexing, and the axis guard the description promises but the code doesn't have right now.

src/Vacuum/DataTypes.jl
jhalpern30: Correct, and it was purely a docstring overreach — the revolved branch and its patch centring
predate this PR, only the claim was new. `PlasmaGeometry3D` places the axisymmetric plasma at
ϕ = ζ − ν(θ) while the revolved wall sits at ϕ = ζ, so same-index is the same geometric angle only
when ν ≡ 0, and `equal_arc_wall` re-parameterizes θ on top. The bullet now says which branch
satisfies it and why the other doesn't.

src/Vacuum/DataTypes.jl
jhalpern30: Agreed. Added an explicit precondition at the top of the constructor:

```julia
(plasma_surf.mtheta, plasma_surf.nzeta) == (mtheta, nzeta) || error(
    "Plasma and wall must be built on the same grid: plasma_surf is ... but this wall is ..."
)
```

Comparing the two dimensions separately rather than the product, so a swapped `mtheta`/`nzeta`
is caught too. Used `error` rather than `@assert` — `@assert` is documented as removable under
optimization, and this guards memory safety, not a debug invariant. One guard covers both loops,
since the fold loop below fails in the opposite direction. `@inbounds` dropped from both loops as
well; this is O(N) setup next to an O(N³) LU, so it is unmeasurable. Test added for the mismatch.

src/Vacuum/DataTypes.jl
jhalpern30: You're right about the mismatch, and I've **fixed the description rather than adding the check** —
your own cost measurement is the reason. On the inboard midplane the toroidal principal radius of
curvature *equals* the distance to the machine axis, so for a conformal offset the area element
collapses exactly where the wall reaches R = 0. The fold guard doesn't fire "right where the axis
is reached" by coincidence; it's the same condition. Measured on the corrugated torus
(`min R_plasma = 1.35`, half-extent 0.35):

| `a` | gap [m] | min area ratio | min R_wall [m] |
|---|---|---|---|
| 3.0 | 1.050 | 0.661 | 0.300 |
| 3.5 | 1.225 | 0.164 | 0.125 |
| 3.86 | 1.350 | → 0 | → 0 |
| 3.9 | 1.365 | — | **fold error** |

The fold error fires at `a = 3.9`; the wall reaches the axis at `a = 3.857`. An R > 0 test on the
conformal branch would be unreachable code, and the revolved branch already inherits 2D's clamp and
hard error. So I dropped "reports the largest admissible `a`" from the PR body (it was never
implemented) rather than adding a check that cannot trip.

test/runtests_vacuum.jl
jhalpern30: Agreed, dead assertion. Replaced with your signed form, which tests the invariant actually at risk

test/runtests_vacuum.jl
jhalpern30: Agreed on both counts — `max(offsets) - min(offsets) < 1e-12` is true by construction of `scale`, and
`all(aligns .< 0)` restates the fold guard's own condition. Took your suggestion. I gave
`_make_3d_periodic_inputs` an `eps` knob so `eps = 0` is a circular cross-section, which has an exact
answer: `0.5(max R − min R)` is the minor radius exactly, so the wall is analytically a circle of
radius `r_minor(1 + a)` in every plane.

```
a = 0.2   analytic 0.36    max|err| 9.4e-5 m
a = 0.5   analytic 0.45    max|err| 8.9e-5 m
```

Residual is the 32 → 24 poloidal resampling of the boundary spline, not the offset. This pins
direction, magnitude and orientation at once — an inward offset would give 0.24, off by 0.12.

src/Vacuum/DataTypes.jl
jhalpern30: Both correct. "major radius" was simply wrong — 2D is `dx = a * r_minor` with
`r_minor = 0.5(x_max − x_min)`. Fixed, and the 3D behaviour is now stated explicitly in the
docstring, with a pointer from the `R_plasma` line.

I've **kept the definition as-is** rather than switching to a per-plane minor radius. It reduces
exactly to the axisymmetric result, so nothing about the 2D↔3D comparison changes; all it does is
give a new meaning on a 3D boundary. Since essentially every real 3D case will supply a wall
externally rather than generate a conformal one, what matters here is only that we *can* generate a
conformal wall at some controllable scaling — not that the scaling coincides with any particular
definition of minor radius. Documenting it is enough; redefining it would move gates B–E for no
practical gain.

src/Vacuum/DataTypes.jl
jhalpern30: You're right that the advice is unactionable, and the reason is sharper than we both had it:
`expand_field_periods` returns early for `nzeta_in ≤ 1` and sets `nfp = 1` otherwise, so `nfp > 1`
can reach this line **only** from an axisymmetric boundary — where calling it again is a genuine
no-op. The message now names the real problem (an axisymmetric surface has no field periodicity;
set `nfp = 1`) and reports the two point counts that disagree. The stale "per-period counts" comment
is gone.

On your out-of-scope find: I **moved the guard above the `nowall` early return**, which closes it
in-PR — cheaper than an issue. Test added that `nowall` no longer skips it.

src/Vacuum/DataTypes.jl
jhalpern30: Agreed — it defaults to `true`, so every stellarator run warned about a setting nobody chose.
Demoted to `@info`; detecting "explicitly set" would need a sentinel default and isn't worth it.
Test assertion updated to `(:info, ...)`.

src/Vacuum/DataTypes.jl
jhalpern30: I'd push back here, and I measured it. Under exact normal offset the parallel surface's normal is
*parallel to the original everywhere* until the caustic, where it flips sign — so `min_align` is
effectively binary, not a 60° cone. On the corrugated torus it is **exactly 1.0000** at every `a`
from 0.2 to 3.5, with the fold at 3.9:

| `a` | 0.2 | 0.5 | 1.0 | 1.3 | 2.0 | 3.0 | 3.5 | 3.9 |
| min align | 1.000 | 1.000 | 1.000 | 1.000 | 1.000 | 1.000 | 1.000 | fold |
| min area ratio | 1.102 | 1.223 | 1.339 | 1.357 | 1.237 | 0.661 | 0.164 | fold |

So the alignment threshold produces no nuisance warnings — the graded indicator is the **area
ratio**, which decays as ∏(1 − d/Rᵢ) and reaches 0 at the fold. If anything the concern runs the
other way: area ratio 0.5 is reached around `a ≈ 3.2` against a fold at 3.9, so the warning arrives
*late*, not early. I've left both thresholds alone on that basis, but happy to tighten the area
ratio to 0.75 if you'd rather have more margin. I don't have the W7-X or precise-QA boundaries in
the repo — if you want those numbers too, point me at the files and I'll add them to this table.

src/Vacuum/DataTypes.jl
jhalpern30: **(1) Pointwise, agreed.** Now `minimum(gap_idx / max(h_θ_idx, h_ζ_idx))` over the surface, and the
warning names the worst-resolved point rather than a surface average.

**(2) The unsatisfiable criterion.** You're right, and right that softening the warning is the wrong
answer. Refining to `nzeta ≳ 120` is unaffordable in a unit test, but the gap enters as a ratio, so
`a` gets there instead. The new end-to-end test sits at **1.11 cells** (`a = 2.5`) and **1.25 cells**
(`a = 3.0`) on a 24×24 grid, and asserts that neither draws a warning. I kept `mtheta, nzeta > 23`
deliberately: at 16×64 the gap criterion is satisfied at a much more natural `a = 0.8`, but it trips
the `PATCH_RAD` guard instead, which defeats the purpose.

So: yes, there is now a tested 3D walled configuration in the accurate regime — but note what it
costs. On a 24×24 grid, `h_ζ ≈ 2πR₀/nzeta ≈ 0.45 m`, so a *resolved* wall is necessarily a *distant*
one (0.875 m from a plasma of half-extent 0.35 m). You cannot have both a close wall and a resolved
gap at this resolution. That is the honest shape of the limitation, and I've rewritten the release
note to say so rather than footnoting it.

src/Vacuum/DataTypes.jl
jhalpern30: Agreed, and your 4.9 % follows directly from the item above. Rather than gate it, the gap is now
computed per branch in whichever way is exact there:

- **Conformal branch**: every point moves exactly `offset_gap` along its own normal, so the
  separation is known by construction — no search at all.
- **Revolved branch**: the wall is axisymmetric, so the true separation is a 2D minimum over the
  wall contour in each plasma point's own plane. O(N·mtheta), exact, and independent of the
  poloidal parameterization — so `equal_arc_wall` no longer perturbs the measurement.

jhalpern30:
> One extra comment: BLOCKING: test coverage (couldn't attach inline): the `compute_vacuum_response 3D conformal wall` testset uses _make_3d_inputs, which is nzeta_in = 1. Therefore it runs the revolved branch, not the new one. And _make_3d_periodic_inputs is never passed to compute_vacuum_response anywhere.
> 
> So of the 30 new tests, none compute Wv on a non-axisymmetric wall, all new coverage is WallGeometry3D construction. With no 3D regression case either, the new numerical path has zero automated protection – it looks like gates A-E exist only as plots in the description. Do you agree? Claude suggested adding one compute_vacuum_response call on _make_3d_periodic_inputs with shape="conformal"? Finiteness and Hermiticity alone would catch a regression. What do you think?

All done, thanks for the review. Note that I am purposefully leaving out a regression harness case. There's no hookup from DCON to a non-axisymmetric run yet, so there is no end-to-end 3D pipeline for the harness to pin. I'd am waiting to add the case when there's a real configuration to track. However, I've added the additional conformal wall coverage to the unit tests

jhalpern30:
@d-burg can you please grace me with an official approval if this is to your liking

d-burg (approved):

PR #437, opened 2026-09-01
Vacuum - FEATURE! - Add the 3D surface-current matrix and cut the solve with stellarator symmetry and conjugate pairing

jhalpern30:
## Release note

- **Audience:** users
- **Numerical impact:** 3D vacuum results shift ~2.2e-3 relative (interpolation-stencil change; 2D untouched) _(harness @ 2fcdba0db)_ — `solovev_n1`, 22 quantities, unchanged
- **Migration:** 3D vacuum results are no longer comparable to those from before this change; regenerate any saved 3D `gpec.h5` baselines or reference values. 2D is unaffected.

The 3D vacuum solve now produces the surface-current matrix `I_v`, the same inductance diagnostic
the 2D path already provided. It is also several times cheaper: folding the mode-family phase into
the kernel, exploiting stellarator symmetry, and serving conjugate toroidal residue classes from a
single factorization together cut a representative `nfp = 5` solve **~4x**, and halve (or quarter)
the operator memory — which is what makes production 3D grids fit in memory at all.

## Scope of this PR

**#436 is merged** and this PR now targets `develop` directly; the merge conflicts it created have
been resolved on this branch. A separate commit adding an HDF5 conducting-wall reader was **split
out** to `feature/vacuum-3d-wall-from-file` so this PR matches what is reviewed here.

**#438 has been merged into this branch** (`2d89bbdb`) and closed. This PR is now the single
home for all three pieces of work:

| commit | what |
|---|---|
| `ef9f2036` | 3D surface-current matrix `I_v`; RHS projected onto the Fourier basis before the solve; mode-family phase folded into the kernel, so only the reduced `[nb·M x nb·M]` operator is stored, one residue class at a time |
| `a61d7dde` | **stellarator symmetry** — the class operator becomes real, and splits into two half-size real blocks for a self-conjugate class |
| `185acd08` | **conjugate pairing** — classes `k` and `−k` share one assembly and one LU |
| `8dfb08ae` | real arithmetic for self-conjugate classes (~2.9x at `nfp = 2`) |
| `551408d6` | comment recording where the singular-quadrature constants come from |

Net diff against `develop`: **5 files, +722 / −242** (`Vacuum.jl`, `Kernel3D.jl`, new
`Symmetry3D.jl`, `DataTypes.jl`, `runtests_vacuum.jl`).

**Review package (diagrams, derivations, benchmarks):**
https://claude.ai/code/artifact/17fac702-393d-4993-9b49-e290234d3ecc

## Regression report

Re-run at the branch head after the commit-subject reword and the `develop` merge. Deliberately
minimal: **no harness case reaches the 3D code this PR changes** — every case is 2D (`nzeta == 1`,
`nfp == 1`) — so a wider sweep confirms nothing a single case does not.

| refs | case | quantities | changed |
|---|---|---|---|
| `cbe890c0c` → `2fcdba0db` | `solovev_n1` | 22 | none |

An earlier full sweep at `ed72b9f55 → b89b46269` covered `diiid_n1` (47), `solovev_n1` (22) and
`solovev_multi_n` (15) — 84 quantities, all unchanged. Those SHAs left the branch in the reword,
but the rewrite changed commit messages only (verified byte-identical trees), so that result still
describes this code.

`develop` is pinned to a commit rather than the branch name: the harness's cached `develop`
reference was a month stale at one point and attributed three KineticForces bugfixes
(`2afc6c3e`, `53ae8081`, `b69e40ce`) that had landed in `develop` to this branch.

Run at `2fcdba0db`, `test/runtests_vacuum.jl` at **378 / 378**.

Zero movement confirms the 2D path is untouched; it does **not** cover the 3D result shift
described below, which no harness case reaches. **There is still no 3D or `nfp > 1` regression
case** — `examples/Solovev_ideal_example_3D` exists and has no case TOML, so nothing would catch a
future 3D regression. See *Worth pushing on*.

## The one thing that moves results

`Kernel3D.jl:127-128` changed the 5x5 Lagrange stencil selection from `trunc(Int, u − 2)` to
`round` before the offset. This is **not** an accuracy fix — against a 72x36 reference the two
agree to 0.06%, the winner flips with resolution (`round` better at 24x12 and 32x16, `trunc`
better at 48x24), and both are exact through degree 4.

It is a *symmetry* fix. `trunc` is not equivariant under the interpolation patch's π-rotation, so
790 of 800 polar nodes got a stencil that was not the mirror of its partner's, and the operator
could not inherit the boundary's symmetry (asymmetry 1.9e-3 nowall / 1.0e-2 with a close wall,
versus 5.2e-13 after). It is a prerequisite for the stellarator-symmetry work and buys nothing on
its own, so it ships here.

Cost: 3D results move once. On `examples/Solovev_ideal_example_3D`, `et[1]` goes
`7.942458514881e-01` → `7.959714927451e-01` (2.2e-3 relative) — comparable to that example's own
discretization error.

### Stellarator symmetry

When both surfaces are stellarator symmetric, the class operator is transformed by a
`StellSymBasis` (new `Symmetry3D.jl`) that makes it **real**, and for a self-conjugate class
splits it into two blocks of roughly half the size. Halves the operator memory and the kernel
work; cuts the factorization ~2.6-2.8x. A boundary that fails the symmetry test (1e-10 relative)
falls through to the untransformed solve automatically.

### Conjugate pairing

The field-period blocks `D_d`, `S_d` are real for *any* boundary — that is already why the code
picks `Float64` when `nfp == 1`. So `D̂₋ₖ = conj(D̂ₖ)` and `Ŝ₋ₖ = conj(Ŝₖ)`, and with
`F := conj(E')` for the partner class's Fourier rows:

```
wv[k'] = c·E' (conj(A) \ conj(B) E'ᴴ)|p  =  conj( c·F (A \ B Fᴴ)|p )        A = D̂ₖ, B = Ŝₖ
```

The partner therefore needs no new code path: run the existing per-class body with `conj(E')` as
its mode basis against the representative's factorization, then conjugate the output block. That
expression is invariant under a simultaneous unitary change of basis, which is why the same rule
works verbatim on the stellarator-symmetric path.

`I_v` is accumulated pre-conjugation and the loop already applies a trailing `conj!`, so for a
partner the two cancel and that call is skipped. **That one-line asymmetry is the likeliest place
for a sign error** — `Vacuum.jl:504`, the `partner || conj!(Iv_block)` line — and is what the A/B
matrix below exists to catch.

## Measured

`nfp = 5`, 64x32 per period, conformal wall, 6 threads, back-to-back in one process:

| modes | K | neither | symmetry only | pairing only | both |
|---|---|---|---|---|---|
| `n = [2,3]` | 2 | 3.189 s | 1.393 s | 1.777 s | **0.769 s (4.15x)** |
| `n = 1:4` | 4 | 6.523 s | 2.825 s | 3.337 s | **1.570 s (4.15x)** |
| `n = 0:4` | 5 | 8.179 s | 3.158 s | 5.093 s | **2.036 s (4.02x)** |

The two mechanisms are multiplicative and do not interfere. Symmetry alone: 2.5x at `nfp = 1`
64x64, 3.8x at 96x96, 7.6x for `nfp = 5, k = 0`. Operator memory 2x, or 4x for a self-conjugate
class with `nfp > 1`; pairing is memory-neutral (peak is still one class's operator).

## Verification

| check | result |
|---|---|
| pairing on/off — 64 configs (`nfp` 1-5, wall/nowall, symmetry on/off, `compute_Iv` on/off) | ≤1.3e-14 |
| `nfp = 1`, `nfp = 2`, single-class mode family (no pair possible) | bitwise identical |
| asymmetric boundary (falls through the stellarator path, still pairs) | 1.1e-16 |
| symmetry disabled vs pre-symmetry code, 7 A/B cases | bit-identical |
| symmetry on vs off | ≤1.4e-12 |
| `examples/Solovev_ideal_example_3D` vs base branch, 135 numeric datasets | bitwise identical |
| `test/runtests_vacuum.jl` | **378 / 378** (re-run at `b89b4626`) |

Pairing agreement is roundoff rather than bitwise on purpose: the unpaired path builds
`cis(-2π(nfp−k)d/nfp)` while the paired path conjugates `cis(-2πkd/nfp)`, and those differ in the
last bits.

## Worth pushing on

- **Zero benefit for anything currently runnable end-to-end.** `VacuumInput(equil, …)`
  (`DataTypes.jl:64`) has no `nfp` argument and `PerturbedEquilibrium` never sets one, so every
  full GPEC run is `nfp = 1` — exactly one class, self-conjugate, nothing to pair. Every example
  and both regression cases are in that regime. The gain is entirely prospective until `nfp` is
  plumbed through. Every `nfp > 1` claim above rests on unit tests.
- **Pairing does not asymptote to 2x.** Full class coverage always leaves 1 (odd `nfp`) or 2
  (even `nfp`) unpairable self-conjugate classes, so the realistic band is 1.3-1.6x; 1.79x needs
  a mode set that avoids `k = 0` and `k = nfp/2`.
- **The 3D Solovev example gains nothing from the symmetry work.** Its equilibrium-extracted
  boundary is stellarator-symmetric only to 1.9e-3, so detection correctly rejects it and it falls
  through. Loosening the tolerance would trade an uncontrolled ~1e-3 error for the speedup; I did
  not. Making flux-surface extraction symmetry-preserving is separate work in `Equilibrium`.
- **Neither reduction has an off switch any more.** The `use_symmetry` / `use_conjugate_pairing`
  kwargs were removed in the cleanup commits: both reductions are now unconditional, gated only by
  the runtime stellarator-symmetry test (`Symmetry3D.jl:93`) and by whether a conjugate class is
  present. That is defensible — they are exact algebra, not tunables — but it does mean there is no
  escape hatch if a boundary is ever found where detection misfires.
- **No 3D regression coverage.** `examples/Solovev_ideal_example_3D` has no harness case, so the
  2.2e-3 shift below is recorded only in prose and nothing would catch a future 3D regression.

## Before this leaves draft

- [x] Re-run the regression harness against the branch head and replace the stale table above
- [ ] Reword the commit subjects — several use `VACUUM`/`WIP`/`IMPROVEMENT`, which
      `ci/conventions/check_subject.py` rejects, and two say the work is unreviewed
- [ ] Add a `solovev_3d` regression case so the 3D path is tracked
- [ ] Assign a human reviewer

> **Reminder, non-negotiable:** this must not be merged into `develop` without approval from a
> third-party human reviewer.


jhalpern30:
Stacked on top of this: #438 (Vacuum - PERF - Serve conjugate toroidal residue classes from one factorization), which targets this branch rather than `develop` and should merge into it first.

jhalpern30:
@claude review

jhalpern30:
@logan-nc this one is rather beefy but I think it can be more or less skimmed by you since none of this is truly production code yet without 3D DCON coupling. The code is rather complex but I've gone through it with a solid first pass to make things readable. The gist of this PR is 1) adding the 3D inductance matrix required for PerturbedEquilibrium and 2) some fancy math to speedup and reduce memory requirements in 3D vacuum. Both are in the 3D VACUUM paper

Issue #240, opened 2026-05-19
PeriodicBC endpoint validator triggered by machine-epsilon mismatches; use targeted endpoint closure instead of blanket check=false


matt-pharr:
## Background

A FastInterpolations version bump introduced a stricter `PeriodicBC(endpoint=:inclusive)` validator: it now errors if `y[1] != y[end]` on the periodic axis. GPEC's `cubic_interp(...; bc=PeriodicBC())` calls in `Equilibrium`, `ForceFreeStates`, `Vacuum`, and `Analysis` started failing because the periodic-axis data is off by ~2e-15 (machine epsilon) from the upstream computation, not because the data is genuinely non-periodic.

PR #196 currently works around this by setting `PeriodicBC(; check=false)` at all 28 call sites across `src/Equilibrium/*`, `src/ForceFreeStates/{Bal,Mercier}.jl`, `src/Vacuum/DataTypes.jl`, and `src/Analysis/PerturbedEquilibriumModes.jl`. **This is a workaround, not a fix** — it silences the validator globally, including for future callers.

## What the proper fix looks like

The codebase already uses the right pattern in a couple of places — explicitly close the endpoint, then keep the strict validator:

- `src/Equilibrium/DirectEquilibriumByInversion.jl:118-120`
  ```julia
  # (The computation above may have broken periodicity due to subtracting rz_in_ys values)
  @views r2[:, end] .= r2[:, 1]
  @views deta[:, end] .= deta[:, 1]
  ```
- `src/Equilibrium/DirectEquilibrium.jl:571`
  ```julia
  ff_fs_nodes[end, :] .= ff_fs_nodes[1, :]  # enforce periodic endpoint
  ```

Every other site that hit the validator should adopt the same pattern: identify where the data is constructed, write `data[..., end] .= data[..., 1]` (or the equivalent slice for the periodic axis), then revert `PeriodicBC(; check=false)` → `PeriodicBC()`. This preserves the safety net that catches genuinely non-periodic input.

## Tasks

- [ ] Inventory the 28 sites currently with `check=false` and identify their data-construction sources
- [ ] Add explicit endpoint closure at each source (matching the existing pattern)
- [ ] Revert each `PeriodicBC(; check=false)` back to `PeriodicBC()`
- [ ] Confirm `runtests_equil.jl` (and the rest of the suite) still pass

## Why this is not blocking for PR #196

[...]

PR #439, opened 2026-09-01
ForceFreeStates - BUGFIX! - Snap the periodic theta endpoint before integrating GGJ geometry

adrianaghiozzi:
## Release note

- **Audience:** developers
- **Numerical impact:** none _(harness @ 987e63f3b, diiid_n1: 47/47 quantities bit-identical, develop vs. local)_
- **Migration:** none

`resist_geometry` (ResistEval.jl) and `resist_eval` (Resist.jl) build the GGJ
theta-integrand independently at each theta node, including both periodic
endpoints, then integrate with `cubic_interp(...; bc=PeriodicBC())`. The two
endpoint evaluations can differ by machine epsilon, which a stricter
`PeriodicBC(endpoint=:inclusive)` validator now rejects outright, aborting
force-free-state prep on some equilibria (mine hit it via a DIII-D case
driven from TokToxStability).

## What

Snap `ff[end, :] .= ff[1, :]` right before each `cubic_interp` call, in both
`ForceFreeStates/Surfaces/ResistEval.jl` and the sibling
`ForceFreeStates/Surfaces/Resist.jl` (same GGJ-integrand pattern, same latent
bug -- not yet observed in the wild but a spot check shows it will trip the
same way on the right input).

This mirrors the fix already in `Equilibrium.jl`'s GS-residual integrator:

    # Snap the repeated endpoint exactly equal to the start
    fs_matrix[end, :] .= fs_matrix[1, :]
    itp = cubic_interp(equil.rzphi_ys, Series(fs_matrix); bc=PeriodicBC())

## Context

- Closes the remaining gap from #240, whose "proper fix" pattern (explicit
  endpoint closure at the data-construction site, not `check=false`) was
  applied everywhere in PR #243 except `ForceFreeStates/Surfaces/`.
- Unrelated to the still-open KineticForces parallel-velocity-spline
  periodicity issue (#341, #390) -- that's a different spline (`tspl`/`vpar`
  in `Torque.jl`/`BounceAveraging.jl`), not the GGJ geometry integrand here.

## Regression report

```
regress --cases diiid_n1 --refs develop,local

Regression Report: diiid_n1
Ref 1: develop  @ 987e63f3b (2026-08-28)
Ref 2: local    @ local     (2026-09-01)
-------------------------------------------------------------------------------------------------
Summary: 47 unchanged
```

Also ran the existing `test/runtests_resist_eval.jl` suite: 63/63 passed.

## Notes for reviewers

No behavior change expected or observed -- this only prevents a false-positive
validator abort on machine-epsilon endpoint drift, matching the fix pattern
already established for #240.

### DIFF
--- a/src/ForceFreeStates/Surfaces/Resist.jl
+++ b/src/ForceFreeStates/Surfaces/Resist.jl
@@ -94,9 +94,7 @@ function resist_eval(sing::SingType, equil::Equilibrium.PlasmaEquilibrium,
         ff[itheta, 6] = dpsisq / bsq
         @views ff[itheta, :] .*= jac / v1
     end
-    # Snap the repeated endpoint exactly equal to the start (see Equilibrium.jl's
-    # GS-residual integrator and issue #240 -- independent spline evaluations at
-    # theta=0/1 can differ by machine epsilon and trip PeriodicBC's check)
+    # Snap the repeated endpoint exactly equal to the start
     @views ff[end, :] .= ff[1, :]
     itp = cubic_interp(equil.rzphi_ys, Series(ff); bc=PeriodicBC())
     avg = FastInterpolations.integrate(itp)
--- a/src/ForceFreeStates/Surfaces/Resist.jl
+++ b/src/ForceFreeStates/Surfaces/Resist.jl
@@ -94,6 +94,10 @@ function resist_eval(sing::SingType, equil::Equilibrium.PlasmaEquilibrium,
         ff[itheta, 6] = dpsisq / bsq
         @views ff[itheta, :] .*= jac / v1
     end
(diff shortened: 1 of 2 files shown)


matt-pharr (changes requested):
@adrianaghiozzi Looks reasonable, but the comments are not really necessary, are they? The git blame will still point to this pr with reasoning. At most if you really think this needs comments, one curt line suffices

adrianaghiozzi:
@matt-pharr Verbose Claude comments strike again lol I just shortened them in [cefa010](https://github.com/OpenFUSIONToolkit/GPEC/pull/439/commits/cefa01003f59a0cc34e20b77681428b2efe54788)

Issue #461, opened 2026-09-13
KineticForces - FEATURE - Rotation prediction: a momentum-balance module fed by tabulated torque surfaces


logan-nc:
## Why

Several questions we want the code to answer are the same computation seen from different sides:

- **What steady-state rotation do we predict if these 3D fields are on, and what if they are switched off?** Today the answer stops at "the NTV torque at the given rotation profile".
- **What is the true neoclassical offset rotation profile** for a given perturbation, rather than the rough `~2·ω_*T` estimate?
- **How much error field can a correction coil cancel before its own non-resonant torque costs the rotation** that keeps the penetration threshold up (the NTV-limits analysis in `ErrorFields`; its current form assumes torque ∝ rotation far from the offset and a positive torque).
- **Profile-consistent critical resonant field** (`Tearing.CriticalResonantField`, #415, #371, #416): the layer's rotation should come from a momentum balance that includes the 3D-field torques, not only from the unperturbed profile.
- Soon: the **resonant J×B torque** from the resistive perturbed equilibrium as a second torque source (#401, #370, #333 track its sign and Hermiticity problems).

All of these are a momentum balance with torque sources that depend on the rotation they act on. The torque sources are expensive (a kinetic torque evaluation is minutes); the balances are cheap. If a transport or balance solver calls the kinetic solve at every iteration or time step the whole thing is unusable, so the framework has to separate the two from the start.

## Principle: torque surfaces, evaluated once, consumed by any solver

Every torque source publishes the same object:

```
TorqueSurface
  psi          :: Vector       # kinetic ψ_N grid the profile is resampled onto
  rotation     :: Vector       # rotation-shift grid Δω (rad/s), adaptive, contains 0
  torque       :: Matrix       # cumulative torque T(ψ, Δω) [N·m], per unit perturbation amplitude²
  dtorque_dpsi :: Matrix       # torque density
  source, method, perturbation label, amplitude normalization, sign convention
```

built **once per perturbation** by scanning the rotation with an adaptive sampler (the NTV torque has sharp features: the superbanana-plateau resonance near ω_E ≈ 0, bounce-harmonic resonances, the offset crossing; a fixed grid misses them or wastes evaluations). Solvers interpolate the surface and never call the physics kernel. Sources add: NTV now, resonant EM later, each its own surface with the same contract. Amplitude scaling is quadratic, so a surface per unit current serves any current.

[...]

PR #455, opened 2026-09-12
ErrorFields - FEATURE - Add NTV-limited error-field correction analysis

logan-nc:
Tenth and last PR of the OMFIT → Julia error-field tolerance migration, stacked on #453: how much error field a correction coil can cancel before the NTV torque of its own field costs the rotation that holds the penetration threshold up — the paper's "NTV limits of EFC" figure, without a new KineticForces API.

**What it does**

- `src/ErrorFields/NTVLimits.jl`: `EFCCoupling` (per kilo-ampere-turn of an array: overlap `C_c`, resonant fraction, NTV torque of the whole field and of the field with the dominant mode projected out), `residual_spectrum`, `correction_current` (linear `(δ_EF − s δ_t)/C_c`, or the smaller root of `a I² − C_c I + (δ_EF − s δ_t) = 0`, `a = s δ_t T_residual / T_0`, when the residual torque lowers the threshold in proportion to the budget it spends), `max_correctable_overlap` (`with_ntv`, where the roots vanish; `torque_only`, where the whole field's torque alone exhausts the budget), `efc_current_curve`.
- `efc_couplings(ffs, coil_sets, rc, dom, kf_ctrl, kinetic_profiles)` (driver level): the torque is a quadratic form of the applied spectrum, so each coupling is one plasma-response evaluation of the unit-current spectrum, injected as forcing modes with nothing written, followed by the kinetic torque — two evaluations per array instead of a current scan. `[ErrorFields.NTV] efc_coils = [...]` runs it after the risk stage and writes `ErrorFields/NTV/`; torque budget, threshold and safety factor stay analysis choices (`read_efc_couplings`, `Analysis.ErrorFields.plot_efc_ntv_limits`).
- The DIII-D example gains a `[KineticForces]` section and treats its C-coil as the correction array; `analyze_example.jl` plots the current-vs-overlap curves with the NTV limit.

**Verification**

- `test/runtests_ntv_limits.jl` (21): projection removes the mode exactly; linear current; the NTV current satisfies the quadratic and exceeds the linear one; safety factor; zero residual torque recovers the linear current; `NaN` past the limit; the closed-form limits; curve shape.
- Solovev run test: the tilted hoop as correction array — its `delta_per_kat` equals the table's nominal overlap over its ampere-turns, residual field orthogonal to the mode, finite torques, HDF5 round trip with the metadata contract, plot, and the kinetic-context guard.

cc @matt-pharr

## Release note

- **Audience:** users
[...]

PR #460, opened 2026-09-13
ForceFreeStates - BUGFIX! - Use the fixed start for integrations that begin away from the axis

logan-nc:
Found while benchmarking the original OMFIT project's error-field case against its Fortran GPEC run. That case starts the integration well inside the plasma rather than at the axis, and with both #457 and #458 applied its innermost resonant-coupling row was still 1.34× Fortran's and the total energies 3 % apart, insensitive to every resolution and tolerance knob in either code.

**Cause.** `initialize_el_at_axis!` applies the Frobenius axis start [Glasser 2016 Eq. 51] wherever the integration begins. That start is a power series about the magnetic axis; at an interior surface it is neither the axis-regular solution nor Fortran DCON's start, which is always U₁ = 0, U₂ = I (zero displacement at the first surface). The DIII-D ideal example, which matches Fortran to four digits at `psilow = 1e-4`, shows the drift directly:

| psilow | innermost coupling row, Julia/Fortran norm | lowest W_t, Julia vs Fortran |
| --- | --- | --- |
| 1e-4 | 1.000 | 1e-4 |
| 0.01 | 1.010 | 0.03 % |
| 0.03 | 1.042 | 0.1 % |
| 0.10 | 1.243 | 2.3 % |
| 0.30 | 2.140 | 7 % |
| 0.30 with the fixed start (`frobenius_psi_max = 0`) | 0.995 | 0.4 % |

On that case the fixed start takes the coupling rows to 1.0000 correlation with Fortran and every per-coil overlap to within 1 % of OMFIT.

**Change.** One control, `frobenius_psi_max` (default 0.01), bounds where the Frobenius start is used. Above it the fixed start is used and a warning says so; `0` selects the fixed start everywhere, silently. It replaces the undocumented `fixed_axis` flag (removed; no example, test or case file set it). Every example in the repo starts at or below 0.01, so nothing tracked moves; the `!` is for the interface and for interior-start runs, whose results change.

**Verification.** `test/runtests_eulerlagrange.jl` gains a Solovev testset: axis start keeps a nonzero Frobenius U₁ with U₂ = I; a start at ψ_N = 0.3 switches to U₁ = 0, U₂ = I and logs the warning; raising `frobenius_psi_max` keeps the Frobenius start; `frobenius_psi_max = 0` selects the fixed start silently even at the axis. ODE Tests 99/99.

[...]

PR #458, opened 2026-09-12
Equilibrium - BUGFIX! - Build the root-area weight with the inverse Fourier transform

logan-nc:
# ⚠️ NEEDS CAREFUL REVIEW: this changes the Fourier convention used inside one routine

**Nothing in `Utilities/FourierTransforms.jl` changes.** The forward transform `ft(data) = basis·data/N` with `basis = exp(−i(mθ − nν))` and the inverse `inverse(ft, modes) = adjoint(basis)·modes` are untouched, and their round-trip test still passes (20/20). What changes is that `compute_sqrtamat` stops rolling its own backward step and calls that inverse.

**History of the routine, so the reviewer can judge whether it was ever right:**

| commit | date | backward step in `compute_sqrtamat` | library inverse at that time |
| --- | --- | --- | --- |
| `02b0f0b7` (first version) | 2026-06-11 | `(1/N) Σ_m c_m exp(−imθ_j)` via `cslth`/`snlth` | `exp(+imθ)` (forward was `exp(−imθ)/N`, Fortran `iscdftf`) |
| `ece3e442` (complex overhaul) | 2026-06-30 | `(basis · c)/N` = the same `exp(−imθ)/N` | `conj(basis)·modes` = `exp(+imθ)` |
| `1c955043` (basis transposed) | 2026-06-30 | `(transpose(basis) · c)/N`, mechanically adjusted | `adjoint(basis)·modes` = `exp(+imθ)` |
| this PR | | `inverse(ft, c)` | unchanged |

At every step the routine's backward kernel was the **forward** kernel `exp(−imθ)` with a 1/N, while the library's inverse was `exp(+imθ)` without one. The two refactors carried the hand-rolled line across faithfully; the mismatch dates from the first version. The routine was never tested directly: the develop `runtests_coordinate_invariant.jl` uses "a generic invertible complex operator standing in for the √weight", and the `scripts/test_power_norm_invariance.jl` the old docstring cited is not in the tree.

**Why the early benchmarks could not see it.** Reflecting m → −m is unitary and the 1/N is a scalar, so every norm or Parseval-type identity holds for the old operator up to a constant. The difference only shows in the structure. On the DIII-D example surface (ψ_N 0.99, mtheta 256, 35 modes; script in the review package):

| check | old operator | this PR |
| --- | --- | --- |
| `S_old[:, m] == S_new[:, −m] / N` over the 25 mirrored modes | max relative residual **0.0** | |
[...]

PR #462, opened 2026-09-13
ErrorFields - FEATURE! - Tabulate the correction-coil NTV torque against rotation and solve a torque balance

logan-nc:
Follow-up to #455 on the same branch line, and the first concrete piece of #461. The NTV-limited correction model in #455 evaluates the torque once, at the nominal rotation, and lets the threshold fall linearly with the torque spent. That assumes rotation far from the neoclassical offset, where the torque is roughly proportional to rotation, and a braking sign. As a plasma brakes it approaches the offset, where the torque falls to zero and reverses, and it can meet the superbanana-plateau resonance on the way; the linear model cannot see either.

**What it does**

- **Torque against rotation, tabulated once.** For each correction array the two plasma responses (whole field, residual field) are computed once as before; the kinetic torque is then evaluated on a grid of rigid E×B rotation shifts `Δω` with the diamagnetic frequencies held fixed (`Equilibrium.shift_exb_rotation`). The grid spans `±span_factor × max(|ω_E|, offset_factor·|ω_*T|)` at the innermost kinetic surface (offset factor 2 as the rough Park, Boozer and Menard 2009 estimate) so it reaches past the offset in both senses, and refines itself where the torque changes fastest and across every sign change (`Utilities.adaptive_sample`, new, reusable). The cumulative torque profile `T(ψ, Δω)` is stored with the totals: the torque surface #461 proposes as the contract between torque sources and any balance or transport solver, so nothing downstream ever calls the kinetic kernel per iteration.
- **Torque balance.** `Δω·T_0/ω_ref = T(Δω)·I²` solved on the branch continuous from the unshifted state; losing the root is the braking bifurcation. `ω_ref` is the density- and volume-weighted ion toroidal rotation (stored; overridable in the analysis). The sign relating torque to rotation is a named constant checked at the balance point.
- **Threshold exponent.** `δ_thresh·(1 + Δω/ω_ref)^α`, `α` an analysis kwarg (default 1). The correction current is a bracketed root solve; the correctable limits are found numerically. A constant braking torque with `α = 1` reproduces #455's closed form exactly, kept as `model = :linear` (the automatic choice for a run without a scan).
- Controls in `[ErrorFields.NTV]` (scan on/off, initial points, cap, tolerance, span and offset factors) decide what is computed; `T_0`, `α`, the safety factor and the model stay analysis choices. HDF5: `ErrorFields/NTV/RotationScan/<array>/` plus `omega_reference` and `omega_offset_estimate`; the reader accepts files without the scan. The plot gains a torque-against-rotation panel with the offsets found.

**Verification**

- `runtests_ntv_limits.jl`: a linear-in-rotation table (braking at the nominal rotation, vanishing at the offset) checks the balance residual, that the rotation never crosses the offset, the exponent, the bifurcation `NaN`, and that the constant table reproduces the closed-form quadratic and its limits; HDF5 round trip with and without a scan.
- `runtests_utilities.jl`: the sampler resolves a Lorentzian narrower than its initial spacing, also when it rides on a background ten times its height (the case the first review figure showed it missing; a scale-free trigger relative to the median interval error and linear end stencils were added for it), brackets a zero crossing, stops on a smooth function, honours the cap, and is deterministic; the rotation shift moves only ω_E.
- DIII-D error-field example with the scan on (C-coil as the correction array): 9 scan points spanning ±100 krad/s; the residual torque is positive at the nominal rotation and crosses zero at a shift of about −89 krad/s (the whole field's at about −81 krad/s), so the offset lies inside the span; the nominal-rotation torques are bit-identical to #455 (harness); the balance and the linear budget agree while the braking is small and part ways as the rotation approaches the offset, where the linear budget drives the threshold to zero near 50 kAt and the balance saturates (review package). `runtests_error_fields.jl` runs the Solovev stage with a three-point scan (101 tests, 9 min instead of 33 with the default scan).
- Regression harness: `diiid_n1` 52 unchanged; new case `diiid_error_field` (the error-field example, the only case exercising `[ErrorFields.NTV]`) 5 unchanged and 3 scan quantities new against `feature/errorfields-ntv-limits`.

Review package (DIII-D and synthetic data only): https://claude.ai/code/artifact/77f402ba-a2cb-4521-a6a9-6f6b7e33f09a (private until shared); design context for the torque-surface contract: #461.

cc @matt-pharr

## Release note

- **Audience:** users
- **Numerical impact:** the NTV-limited correction analysis now uses a torque balance on a torque-versus-rotation table when the run tabulates one (`rotation_scan = true`, the new default); the nominal-rotation torques are unchanged, and the `:linear` model now refuses a root at which the residual torque has exhausted the budget (as OMFIT's script did), which caps its correctable-overlap limit at `C_c √(T_0/T_res)` _(harness @ 98013a66: diiid_error_field 5 unchanged + 3 new scan quantities; diiid_n1 52 unchanged)_
- **Migration:** none; set `rotation_scan = false` to keep the single evaluation

`[ErrorFields.NTV]` tabulates each correction array's NTV torque against a rigid E×B rotation shift past the neoclassical offset on an adaptive grid, stores the torque surfaces under `ErrorFields/NTV/RotationScan/`, and the analysis solves a zero-dimensional torque balance with a threshold exponent instead of the linear torque budget. `Utilities.adaptive_sample` and `Equilibrium.shift_exb_rotation` are new.

## Regression report

```
At 98013a66 (head), refs feature/errorfields-ntv-limits @ 55ef4ba5 vs feature/errorfields-ntv-rotation-scan @ 98013a66:
  diiid_error_field: 5 unchanged (EFC overlap per kAt, whole-field and residual NTV torque per kAt², dominant singular values,
                     |delta_nominal|), 3 new (ω_ref, rotation-scan shifts, residual torque against rotation); runtime 333 s → 713 s
  diiid_n1:          52 unchanged; runtime 271 s → 295 s
```

## Notes for reviewers

- Stacked on #455 (base branch `feature/errorfields-ntv-limits`).
- Interpolation is FastInterpolations cubic splines throughout (torque lookup, zero crossings, the balance's slope check, the resampled torque profiles), consistent with the code base; the harness report below was rerun at the head commit after that change.
- Cost: the plasma responses are unchanged (two per array); the kinetic torque is now evaluated `2 × n_scan` times per array instead of twice, with `n_scan` between the initial grid and the cap. In the DIII-D example, 18 kinetic evaluations (9 scan points × 2 spectra) take 410 s on 16 threads against two before; the harness runtime of the case goes from 333 s to 686 s. The example deck caps the scan at 9 points; the default cap is 21.
- The sign convention between the kinetic torque and ω_E is encoded once (`TORQUE_ROTATION_SIGN`) and verified on the example: the tabulated torque rises through zero with the shift, so a restoring balance needs `TORQUE_ROTATION_SIGN = −1` (a positive kernel torque brakes the file's co-rotation); `rotation_shift` checks the slope at every crossing it uses and warns if a run disagrees.

https://claude.ai/code/session_01BzKDmtKYDotJWrcxWf1oJu


logan-nc:
Scope note: two commits that stored the torque *density* surface for transport use (874d4b97, 710c5737) were pushed and then removed by force-push; the branch is back at d3812cf9, the state the tests and harness report in the body describe. That capability belongs to the momentum-balance work in #461, and this PR stays the simple zero-dimensional model.

logan-nc:
Review package for this PR (republished at a new link; DIII-D-like and synthetic data only): https://claude.ai/code/artifact/77f402ba-a2cb-4521-a6a9-6f6b7e33f09a

logan-nc:
## Stack summary — OMFIT → Julia error-field tolerance migration (updated 2026-09-17)

**Nothing in this set merges without a third-party human review**, the three small bugfix PRs included.

Full page with the diagram, per-PR table, review order and links to every review package (a private claude.ai artifact until the author shares it; DIII-D-like and synthetic data only): https://claude.ai/code/artifact/408ddd36-e52c-42d8-8f5c-52f680f41da9

| Group | PRs | Base |
| --- | --- | --- |
| Bugfixes the stack's numbers depend on — **review first** | #458 root-area weight (careful review: Fourier convention inside one routine), #457 helicity from the current sign, #460 interior-start initialization | `develop` |
| Prerequisites | #446 resonant-coupling SVD API, #447 per-coil-set forcing modes | `develop` |
| ErrorFields stack, each based on the previous branch | #448 sensitivities → #449 tolerance TOML → #450 sampling → #451 Monte Carlo → #452 risk + scan → #453 plots + phasing → #455 NTV limits → #462 NTV torque against rotation | #448 on `develop` (carries merges of #446/#447); rebase the chain after the bugfixes and #446/#447 land |

**Benchmark, qualitatively.** The original OMFIT project's case, rebuilt from that project's own run inputs, is reproduced once #458, #457 and #460 are in: dominant-mode singular values agree to 1e-4, singular-coupling rows match Fortran GPEC at 1.0000 correlation and within 0.5 % in norm, and every coil set's dominant-mode overlap agrees within 1 % once both codes use the same converged toroidal coil grid. The DIII-D-like example agrees with Fortran to 1e-4 throughout. Any comparison shown on these PRs uses the DIII-D-like examples only.

**Design issue for what comes after the stack:** #461 (momentum-balance module fed by tabulated torque surfaces; stationary solvers only).

logan-nc:
Window change merged up the stack (2026-09-17): the dominant-mode SVD and every entry point built on it now default to the core window ψ_N ≤ 0.9 (`PerturbedEquilibrium.CORE_PSI_HIGH`, from #446 at f43dd785), instead of the full window. In-memory and `gpec.h5` analyses therefore agree by default, and other windows remain an explicit analysis choice. The DIII-D-like error-field example's analysis uses the core window alone (the earlier edge-only contrast is gone). Tests on this branch pass after the merge; harness at the top of the stack (#462): against develop, 47 unchanged plus the stack's new quantities; against the previous top (98013a66), the window moves exactly what it should and nothing else — `diiid_n1`: dominant singular values now over the 3 core surfaces instead of 5, the forcing overlap, |delta_nominal| and the shift/tilt sensitivity norms by 6–11 %; `diiid_error_field`: the C-coil coupling per kAt by 11 %, the residual torque (whose projection uses the mode) by 56 %, the whole-field torque, ω_ref and the scan grid unchanged.

